PR #7, opened 2019-01-05
Improvements for Segment Iterator

Rahix:
This series of commits contains a few improvements to the segment iterator.  I split the changes into small commits so it is easier to understand.  If some changes don't suit you, feel free to exclude these commits.

1. 9547e43cb96b51bd144313d4f8606efa86b757a7: A few types that should implement `Debug` didn't.
2. 710973600f27e0380e6e2b088af03c4284944bad: In my opinion it is unidiomatic Rust to include the index in the iterator by default.  As visible in your code, most of the time you just throw it away anyway ... If you do need it, you can easily add the index back using `path.segments().enumerate()`.
3. 354c392f42788f87f9643d0184a874a292931e7d: Just a small change, `slice::get()` behaves the same way as the previous implementation.
4. ad506cab9c193012524d794a809603484713d256: As we are iterating over a slice, we can use the slice-iterator instead of doing it on our own.
5. 8127083d3b8f239ff2f8497dd6725526e4b5306d: To quote the commit message:
  > Logically `start` and `last` should always have values.  This commit reflects this in code.  If no value for first could be found (because the first element is not a `PathEl::Moveto`), a `panic!` with a proper message will occur.
6. ea1a94af277ba07e4b529c15cb37e668ff2d05f1: I am not quite sure about this one.  I tried reducing code duplication, but some people might argue that the new version is harder to understand.  Your choice if you want to include it ...

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -82,9 +82,7 @@ impl BezPath {
     }
 
     /// Iterate over the path segments.
-    ///
-    /// The iterator returns the index within the path and the segment.
-    pub fn segments<'a>(&'a self) -> impl Iterator<Item = (usize, PathSeg)> + 'a {
+    pub fn segments<'a>(&'a self) -> impl Iterator<Item = PathSeg> + 'a {
         BezPathSegs {
             c: &self.0,
             ix: 0,
@@ -140,23 +138,23 @@ impl BezPath {
     /// Note: this is one of the methods that could be implemented on an iterator
     /// of `PathEl`, to save allocation.
     pub fn arclen(&self, accuracy: f64) -> f64 {
-        self.segments().map(|(_, seg)| seg.arclen(accuracy)).sum()
+        self.segments().map(|seg| seg.arclen(accuracy)).sum()
     }
 
     /// Compute the total signed area.
     pub fn area(&self) -> f64 {
-        self.segments().map(|(_, seg)| seg.signed_area()).sum()
+        self.segments().map(|seg| seg.signed_area()).sum()
     }
 
     /// Find the nearest point.
     ///
     /// Panics if path is empty or invalid.
     ///
-    /// Returns the index of the element, the parameter within that segment, and
+    /// Returns the index of the segment, the parameter within that segment, and
     /// the square of the distance to the point.
     pub fn nearest(&self, p: Vec2, accuracy: f64) -> (usize, f64, f64) {
         let mut best = None;
-        for (ix, seg) in self.segments() {
+        for (ix, seg) in self.segments().enumerate() {
             let (t, r) = seg.nearest(p, accuracy);
             if best.map(|(_, _, r_best)| r < r_best).unwrap_or(true) {
                 best = Some((ix, t, r));
@@ -169,7 +167,7 @@ impl BezPath {
     ///
     /// TODO: make sure all the signs are consistent.
     pub fn winding(&self, p: Vec2) -> i32 {
-        self.segments().map(|(_, seg)| seg.winding(p)).sum()
+        self.segments().map(|seg| seg.winding(p)).sum()
     }
 }
 
@@ -212,9 +210,9 @@ struct BezPathSegs<'a> {
 }
 
 impl<'a> Iterator for BezPathSegs<'a> {
-    type Item = (usize, PathSeg);
+    type Item = PathSeg;
 
-    fn next(&mut self) -> Option<(usize, PathSeg)> {
+    fn next(&mut self) -> Option<PathSeg> {
         while self.ix < self.c.len() {
             let ix = self.ix;
             let el = self.c[ix];
@@ -227,24 +225,24 @@ impl<'a> Iterator for BezPathSegs<'a> {
                 PathEl::Lineto(p) => {
                     let seg = PathSeg::Line(Line::new(self.last.unwrap(), p));
                     self.last = Some(p);
-                    return Some((ix, seg));
+                    return Some(seg);
                 }
                 PathEl::Quadto(p1, p2) => {
                     let seg = PathSeg::Quad(QuadBez::new(self.last.unwrap(), p1, p2));
                     self.last = Some(p2);
-                    return Some((ix, seg));
+                    return Some(seg);
                 }
                 PathEl::Curveto(p1, p2, p3) => {
                     let seg = PathSeg::Cubic(CubicBez::new(self.last.unwrap(), p1, p2, p3));
                     self.last = Some(p3);
-                    return Some((ix, seg));
+                    return Some(seg);
                 }
                 PathEl::Closepath => {
                     let last = self.last.take();
                     let start = self.start.take();
                     if last != start {
                         let seg = PathSeg::Line(Line::new(last.unwrap(), start.unwrap()));
-                        return Some((ix, seg));
+                        return Some(seg);
                     }
                 }
             }
@@ -440,7 +438,7 @@ impl Shape for BezPath {
 
     fn bounding_box(&self) -> Rect {
         let mut bbox: Option<Rect> = None;
-        for (_, seg) in self.segments() {
+        for seg in self.segments() {
             let seg_bb = seg.bounding_box();
             if let Some(bb) = bbox {
                 bbox = Some(bb.union(seg_bb));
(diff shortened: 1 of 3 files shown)


raphlinus:
Thanks for the contribution! I'll dig in in detail when I get a chance, unfortunately not likely today.

One quick comment re 7109736: this is by design, and reflects the fact that the index of a path *element* is not going to be the same as the position in a sequence of path *segments.* This part of the API is basically for hit testing parts of a curve. See for example https://github.com/raphlinus/spline-research/blob/master/bezpath.js#L67 , which uses a slightly different approach - it adds a "mark" command along with moveto/curveto, and then returns the mark.

I'm thinking of alternate approaches. One that comes to mind is that the client maintains a Vec of its own "mark" and then *segment* indices, and in general does bsearch to find the mark of a segment index. But I think element indices have their own appeal, mostly they can be used for random indexing of slices. In any case, I wanted to explain why I made that choice.

Thanks for the interest and the contribution!

Rahix:
> the index of a path element is not going to be the same as the position in a sequence of path segments

Thanks for the feedback! Ouch, looks like I completely missed this ...

[This line](https://github.com/linebender/kurbo/pull/7/commits/710973600f27e0380e6e2b088af03c4284944bad#diff-9ffa5553a20eeac4d38ef50d6b11ee37R159) is actually broken then ...

Ok, I can see that we need some way to map segments to elements.  The other way around is trivial ([`BezPath::get_seg()`](https://docs.rs/kurbo/0.1.2/kurbo/struct.BezPath.html#method.get_seg)) but this way would not be O(1) in either time or space.  Luckily the only case where we need this mapping (from my understanding) is for hit-testing, which runs in a loop.  This is why your current implementation can efficiently keep track of the index.

I still feel like the segment iterator should *not* be the place to do this though.  From my downstream perspective I don't see many cases where I need it and even in the library code 4/5 uses just throw the index away.  The one case where it is needed ([`BezPath::nearest()`](https://docs.rs/kurbo/0.1.2/kurbo/struct.BezPath.html#method.nearest)) seems special enough to me to use a different approach.

In code:

```rust
// Adapted `nearest` implementation
for ix in 0..self.0.len() {
    if let Some(seg) = self.get_seg(ix) {
        let (t, r) = seg.nearest(p, accuracy);
        if best.map(|(_, _, r_best)| r < r_best).unwrap_or(true) {
            best = Some((ix, t, r));
        }
    }
}
```

Speed-wise this should be exactly the same but it allows keeping the segment iterator "clean".  As long as this is the only place where we need access to the element index, I would be ok with this implementation, but if there are more, a better solution is needed.  One thing that might work is storing the index in the `PathSeg` directly?  Or maybe a separate iterator type which keeps track of the element index?

I am unsure, I can only see my user perspective right now, if there is more that I have missed so far, please correct me!

raphlinus:
Rahix: thanks for responding. I think you're right that `nearest` is the special case here.

I thought about it some more, trying to see it from the point of view of solving the UI problem, rather than just following what I did in Spiro and the latest spline tool. I think the mis-correspondence between segment and element index is really only a problem with multiple subpaths. So one approach is to require the client to generate a path object for each subpath. This has significant value aside from just the iterator index issue, as it would be possible to do separate bounding-box logic for each subpath. Then, we could document that for a "simple" path consisting of a moveto followed by lineto/quadto/cubicto, and optionally terminated by a closepath, then the segment index is guaranteed to be one plus the element index. I think this would solve the problem fairly neatly and simplify the interface.

I'll see if I can review the rest of the patch in detail today (weekends tend to be busy for me with non-code things).

src/bezpath.rs
```diff
@@ -84,12 +84,17 @@ impl BezPath {
     /// Iterate over the path segments.
     ///
     /// The iterator returns the index within the path and the segment.
```
raphlinus: This comment is no longer relevant and should be removed.

src/bezpath.rs
```diff
@@ -156,10 +161,12 @@ impl BezPath {
     /// the square of the distance to the point.
     pub fn nearest(&self, p: Vec2, accuracy: f64) -> (usize, f64, f64) {
         let mut best = None;
-        for (ix, seg) in self.segments() {
-            let (t, r) = seg.nearest(p, accuracy);
-            if best.map(|(_, _, r_best)| r < r_best).unwrap_or(true) {
-                best = Some((ix, t, r));
+        for ix in 0..self.0.len() {
```
raphlinus: I think I prefer the previous state (before afcb951) if the comment is fixed to indicate that it returns the segment index.

src/bezpath.rs
```diff
@@ -205,51 +212,39 @@ impl Mul<BezPath> for Affine {
...
+                    break;
                 }
-            }
+            };
+
+            self.last = last;
+            return Some(ret);
         }
-        None
+        return None;
```
raphlinus: Why the added `return`? I personally find the former case more idiomatic Rust (this feels like a C-ism to me).

src/bezpath.rs
```diff
@@ -205,51 +212,39 @@ impl Mul<BezPath> for Affine {
...
                 PathEl::Closepath => {
-                    let last = self.last.take();
-                    let start = self.start.take();
-                    if last != start {
-                        let seg = PathSeg::Line(Line::new(last.unwrap(), start.unwrap()));
-                        return Some((ix, seg));
+                    if self.last != self.start {
+                        return Some(PathSeg::Line(Line::new(self.last, self.start)));
                     }
+                    break;
```
raphlinus: I think this is wrong. In the case of two successive triangles of the form "moveto(p1) lineto(p2) lineto(p3) moveto(p1) closepath", I think this makes it return `None` after the first triangle.

Of course, to really settle this question there should be a test :) The lack of tests is just laziness, as I was pounding out the code.

src/bezpath.rs
```diff
@@ -205,51 +212,39 @@ impl Mul<BezPath> for Affine {
...
-                    return Some((ix, seg));
+                    (PathSeg::Cubic(CubicBez::new(self.last, p1, p2, p3)), p3)
                 }
                 PathEl::Closepath => {
-                    let last = self.last.take();
-                    let start = self.start.take();
-                    if last != start {
-                        let seg = PathSeg::Line(Line::new(last.unwrap(), start.unwrap()));
-                        return Some((ix, seg));
+                    if self.last != self.start {
```
raphlinus: One note: this changes the semantics of which paths are considered valid. Now, if you do (say) a lineto after a closepath, it continues it from the last point specified (not the start point). This is different than [the SVG spec](https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand), which starts the next path at the same start. I think setting self.last to self.start would be better, as at least it would match SVG semantics.

src/bezpath.rs
Rahix: Looks like you are right ... Well, I wrote a really dirty test to check for regressions but it looks like this is something I missed ... Will fix!

src/bezpath.rs
Rahix: I guess this is preference.  My personal "rule" is:  If a function is written in an imperative style, use explicit returns and if it is functional, use the expression syntax.  In this case, the `for` loop design makes this function imperative.

The other reason is that for the first one I *have to* use an explicit return because break with value is not yet supported for anything except `loop`s.  To keep the code homogeneous I feel like both returns should use the same syntax ...

raphlinus: I understand your reasoning here, but the consensus of the Rust community is to not use an explicit return. There's a [Stack Overflow thread](https://stackoverflow.com/questions/27961879/why-is-using-return-as-the-last-statement-in-a-function-considered-bad-style) on the exact topic, and it's formalized as a clippy lint: [needless return](https://rust-lang.github.io/rust-clippy/master/index.html#needless_return)

So, not to be too strict, but I would like this changed. Consistency of style is important, and where there's a legitimate difference of opinion I think it makes sense to go with clear community consensus.

Other than that, I'm happy to merge. I'm satisfied with the current state of segment enumeration and nearest (although I'll probably want to clarify the documentation to make it absolutely clear that it won't, for example, elide zero-length lines).

Thanks!

Rahix: Ok, I changed it.

PR #8, opened 2019-01-07
Rework ownership of Shape

raphlinus:
Allow either owned shape or reference to be passed to `impl Shape`.
Allow taking ownership of Vec from BezPath. Add `impl Shape` for for
slice of PathEl.

There are some open questions, mostly how much to expose iterators
on PathSeg.

Closes linebender/piet#3


raphlinus:
I added some clarification of indices for the iterators, which is a bit of a followup to #7 , not strictly related to ownership but probably not deserving a separate PR. @Rahix might want to take a look.

Issue #4, opened 2019-01-02
Figure out sign conventions


raphlinus:
One of the current points of confusion is whether the coordinate system is y-up or y-down. The former is more conventional in math, the latter completely standard in graphics these days (though this was less true, PostScript was y-up, as are fonts to this day).

Related issues are conventions around whether a positive area shape is clockwise or withershins, and whether a clockwise circle has positive or negative signed curvature. The winding number should match area - it should be positive if the point is inside a positive area shape. Either convention will do, but it should be consistent.

Another thing that falls in the same category is the convention of `Vec2::cross`. The right hand rule says right ^ up = 1, but is up (0, 1) or (0, -1)? For this, I think we pick one and document it.

I do think kurbo is interesting in both math and graphics contexts. For methods like `Vec2::from_angle` and `Affine::rotate` it's hard to just pick a convention, so I'm inclined to split them into `_y_up` and `_y_down` variants, so that the user makes an explicit choice. For the signed area stuff, I'm not sure there's a strong convention; e^{iπt} is anticlockwise in a y-up coordinate space as t increases. Do people know of other conventions we can follow? Something along the lines of https://github.com/JuliaLang/julia/issues/8750 which surveys rounding conventions of many languages might be a model.

Tagging @behdad @jrus @pomax @nical @Connicpu @kvark as they might have horses in this race.

Pomax:
Call me crazy, but if the idea is, basically, "the best library" then it feels like there should be a way to bootstrap the coordinate system so that after calling the appropriate function, the library "just works" using whatever convention it should be using. And then whether that takes the form of a static call, or per-call option, or even literally a duplicated library with the two different y conventions should probably be informed by "what performs best".

raphlinus:
@Pomax I'm open to that, but want to make sure it doesn't come at a significant cost of compile time, usability, doc clarity, etc.

Sketching this out, here are a couple ways this might look:

One is to have a zero-size type for Y direction, then instead of `Affine` and `Vec2` you have `Affine<YUp>` and `Vec2<YDown>`. You might have `From` conversions that flip the sign of y as well, so you can interop. However, I fear that this might be really bad for ergonomics - I think the user would get annoyed frequently having to constantly specify this explicitly.

Another possibility is to do this at the module level: instead of `use kurbo::{Affine, Vec2}` as now, write `use kurbo::y_up::{Affine, Vec2}`. Internally that might be implemented as above.

[...]

PR #9, opened 2019-01-09
Add way to get coefficients from Affine

raphlinus:
Add `as_coeffs` method to Affine so we can get the coefficients out
(necessary for conversion to Direct2D types, etc).

Also tweak documentation to explain the convention for rotation
(partially addressing #4) and the augmented matrix representation of
Affine, particularly that it's backwards from PostScript.

This doesn't address all of #4, it doesn't change any behavior.


raphlinus:
Ping @Connicpu for sanity-check regarding matrices. In particular, does the mint conversion (ColumnMatrix2x3) still make sense?

PR #14, opened 2019-05-07
Minor tweaks to bezier paths

raphlinus:
Allow multiplication by affine to take a BezPath by reference.

Add "A", "S", "H", "V" commands to SVG path parsing. The SVG arc logic is adapted from lyon_geom. Thus, Nicolas is credited in the AUTHORS file (and there's a bit of bureaucracy around attribution).

Note that the arc flattening is very crude. I'll probably update this PR with a more refined version.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,7 @@
+# This is the list of kurbo authors for copyright purposes.
+#
+# This does not necessarily list everyone who has contributed code, since in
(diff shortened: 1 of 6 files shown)


raphlinus:
@nical Tagging you to get your sign-off that I've handled attribution of code adapted from lyon_geom properly.

PR #17, opened 2019-06-15
Add 'euclid' feature for interop with euclid crate

cmyr:
This also adds a bunch of conversion implementations, as well
as implementing `Shape` for euclid::Rect.

It also makes a few call sites that take `Vec2` more flexible, and impls `From<(f32, f32)> for Vec2`, which is maybe questionable but seems low-stakes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,18 @@ arrayvec = "0.4.10"
 version = "0.5.1"
 optional = true
 
+[dependencies.euclid-crate]
+package = "euclid"
+version = "0.19.9"
+optional = true
+
+[dependencies.num-traits]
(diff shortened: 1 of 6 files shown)


cmyr:
No need to merge this immediately, was just working on druid interop and ended up down here.

RazrFalcon:
Can this be optional? The whole reason I've switched to `kurbo` from `lyon`, is because `euclid` is too heavy https://github.com/servo/euclid/issues/345.

cmyr:
@RazrFalcon thanks for the link, happy to learn from others' experiences. Will look into just adding our own `Point` and `Size` types, and put interop stuff behind a feature flag.

PR #18, opened 2019-06-16
Add rounded rectangle shape

msiglreith:
Rectangle shape with rounded corners (all corners same radius).

TODO:
- [x] Winding (not sure what to do here)
- [x] More accurate area estimate (handle invalid states radius > width/2?)
- [x] Path generation


src/rounded_rect.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::{PathEl, Rect, Shape, Vec2};
+use std::f64::consts::PI;
+
+/// A rectangle with rounded corners.
+#[derive(Clone, Copy, Default, Debug)]
+pub struct RoundedRect {
+    /// Coordinates of the rectangle.
+    pub rect: Rect,
+    /// Radius of all four corners.
+    pub radius: f64,
```
raphlinus: An interesting question, should there be a single radius value or one for each corner? I think I would lean toward the latter as it's more general and I can easily imagine such things used in UI.

src/rounded_rect.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    /// The center point of the rectangle.
+    #[inline]
+    pub fn center(&self) -> Vec2 {
+        self.rect.center()
+    }
+
+    /// Take absolute value of width, height and radius.
+    ///
+    /// The resulting rect has the same extents as the original, but is
```
raphlinus: What happens when the radius has different sign as the area? I'm wondering whether this case should be forbidden by construction.

src/rounded_rect.rs
```diff
@@ -0,0 +1,138 @@
...
+            + 2.0 * PI * self.radius.abs()
+    }
+
+    #[inline]
+    fn winding(&self, pt: Vec2) -> i32 {
+        unimplemented!() // TODO
+    }
+
+    #[inline]
+    fn bounding_box(&self) -> Rect {
```
raphlinus: Note that this will require `abs` if we allow negative area rectangles, and concern for out-of-range radius.

src/rounded_rect.rs
msiglreith: Good point, I currently focused on one radius for all corners as this is natively supported by d2d (actually they use ellipses but not sure how useful that is). Nonetheless, the backend could decide internally which path to use for rendering.

msiglreith:
Rebased on master and implemented the missing pieces. Please have a look again (:
I moved the arc part into a separate module as I use it for implementing the path generation for the rounded rectangle.

src/rounded_rect.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Note: nothing forbids negative height.
+    #[inline]
+    pub fn height(&self) -> f64 {
+        self.rect.height()
+    }
+
+    /// Absolute radius of the rounded corners.
+    ///
+    /// Clamped to half the size of the rectangle.
+    pub fn radius(&self) -> f64 {
```
raphlinus: I think it would be better do this on construction.

src/rounded_rect.rs
```diff
@@ -0,0 +1,302 @@
...
+            rect: Rect::from_points(p0, p1),
+            radius,
+        }
+    }
+
+    /// A new rectangle from origin and size.
+    ///
+    /// The result will have non-negative width and height.
+    #[inline]
+    pub fn from_origin_size(origin: Point, size: Vec2, radius: f64) -> RoundedRect {
```
raphlinus: The method in `Rect` that this is copied from is wrong, the `size` argument can have negative components :/

src/rounded_rect.rs
```diff
@@ -0,0 +1,302 @@
...
+        assert!((rect.area() - rounded_rect.area()).abs() < epsilon);
+
+        // Extremum: half-size radius corner -> circle
+        let circle = Circle::new((0.0, 0.0), 50.0);
+        let rounded_rect = RoundedRect::new(0.0, 0.0, 100.0, 100.0, 50.0);
+        assert!((circle.area() - rounded_rect.area()).abs() < epsilon);
+    }
+
+    #[test]
+    fn winding() {
```
raphlinus: I think if this was tested on a rect not centered at the origin, it would reveal more problems.

src/rounded_rect.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+}
+
+// This is clockwise in a y-down coordinate system for positive area.
+impl Iterator for RoundedRectPathIter {
+    type Item = PathEl;
+
+    fn next(&mut self) -> Option<PathEl> {
+        if self.idx > 4 {
+            return None;
```
raphlinus: Does this miss the `ClosePath` at the end? I'm finding it a bit hard to follow the logic.

src/rounded_rect.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    #[inline]
+    fn winding(&self, pt: Point) -> i32 {
+        let radius = self.radius();
+        let inside_half_width = (self.width() / 2.0 - self.radius()).max(0.0);
+        let inside_half_height = (self.height() / 2.0 - self.radius()).max(0.0);
+
+        // project point out of the inner rectangle (positive quadrant)
+        let px = (pt.x.abs() - inside_half_width).max(0.0);
```
raphlinus: I think you need to be subtracting the center from `pt` first; then I believe the logic is probably correct, but haven't validated it carefully.

src/rounded_rect.rs
msiglreith: Yes, also verified it visually. Added a few more inline comments for better explanation.

msiglreith:
Thanks for the great review!
Fixed the issues and improved documentation a bit.

src/rounded_rect.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Coordinates of the rectangle.
+    rect: Rect,
+    /// Radius of all four corners.
+    radius: f64,
+}
+
+impl RoundedRect {
+    /// A new rectangle from minimum and maximum coordinates.
+    #[inline]
+    pub fn new(x0: f64, y0: f64, x1: f64, y1: f64, radius: f64) -> RoundedRect {
```
cmyr: Two thoughts:

- I would add a new constructor that takes a `Rect` and a `radius`.
- In `RoundedRect::new`, I would just call `Rect::new`, and then I would call the new constructor mentioned above.

src/arc.rs
```diff
@@ -0,0 +1,113 @@
+use crate::{PathEl, Point, Vec2};
+use std::f64::consts::{FRAC_PI_2, PI};
+
+/// A single arc segment.
+#[derive(Clone, Copy, Debug)]
+pub struct Arc {
```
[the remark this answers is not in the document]
msiglreith: This has been made public recently and also exported as part of the svg module.

src/rounded_rect.rs
msiglreith: Thanks, done!

msiglreith:
ping, any open tasks left for me (:?

PR #20, opened 2019-06-18
Add barebones Size and Point types

cmyr:
These implementations are intentionally minimal; I would like to
avoid adding public API that we might want to break until we have
a clearer idea of how these types are to be used.

Currently, I think there are two options: either we treat these as
convenience types that are exposed so that clients that want them
can use them, and we make minimal changes to public API, _or_ we
take a larger look at where Vec2 is used and attempt to use these
types more extensively where it makes sense.

In any case, this gives us just enough that I can replace the use of 
`Geometry` and `(f32, f32)` in druid, which I am motivated to do.


src/point.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct Point {
+    /// The x coordinate.
+    pub x: f64,
+    /// The y coordinate.
+    pub y: f64,
+}
+
+impl Point {
+    /// The point (0, 0).
+    pub const ZERO: Point = Point::new(0., 0.);
```
raphlinus: A case can be made that this should be called ORIGIN instead of ZERO.

cmyr: Here and with `Size` names were in deference to existing APIs. (I'd wanted to call the `(0, 0)` size 'empty', but landed on zero after looking around.)

PR #23, opened 2019-06-19
Be systematic in vector types

raphlinus:
Use `Point` and `Size` where appropriate instead of `Vec2`.

Also rename the bezpath ops to conform to Rust naming conventions (`move_to` rather than `moveto`, etc).

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -8,7 +8,7 @@ use crate::common::{solve_cubic, solve_quadratic};
 use crate::MAX_EXTREMA;
 use crate::{
     Affine, CubicBez, Line, ParamCurve, ParamCurveArclen, ParamCurveArea, ParamCurveExtrema,
-    ParamCurveNearest, QuadBez, Rect, Shape, Vec2,
+    ParamCurveNearest, Point, QuadBez, Rect, Shape,
 };
 
 /// A path that can Bézier segments up to cubic, possibly with multiple subpaths.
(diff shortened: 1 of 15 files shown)


src/cubicbez.rs
```diff
@@ -31,7 +31,7 @@ struct ToQuads {
 impl CubicBez {
     /// Create a new cubic Bézier segment.
     #[inline]
-    pub fn new<V: Into<Vec2>>(p0: V, p1: V, p2: V, p3: V) -> CubicBez {
+    pub fn new<P: Into<Point>>(p0: P, p1: P, p2: P, p3: P) -> CubicBez {
```
cmyr: this has the limitation of requiring all of these arguments to be of the same type. Assuming we like `Into<Point>`, there's a case that we might want to have each argument `impl Into<Point>` independently, so you could do things like `CubicBez::new(point_a, point_b, (1.0, 32.0), (0., 101.))`?

src/bezpath.rs
```diff
@@ -316,7 +316,7 @@ impl ParamCurveArea for PathSeg {
 }
 
 impl ParamCurveNearest for PathSeg {
-    fn nearest(&self, p: Vec2, accuracy: f64) -> (f64, f64) {
+    fn nearest(&self, p: Point, accuracy: f64) -> (f64, f64) {
```
cmyr: Is this the best return type here?

PR #25, opened 2019-06-20
Add some helper functions

cmyr:
These are a few small helpers I've wanted while doing stuff in druid. Will likely have more to add in the future but no reason not to PR these now.

### DIFF
--- a/src/point.rs
+++ b/src/point.rs
@@ -35,14 +35,20 @@ impl Point {
 
     /// Linearly interpolate between two points.
     #[inline]
-    pub fn lerp(&self, other: Point, t: f64) -> Point {
+    pub fn lerp(self, other: Point, t: f64) -> Point {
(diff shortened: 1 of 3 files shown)


src/point.rs
```diff
@@ -35,15 +35,21 @@ impl Point {
...
     /// Determine the midpoint of two points.
     #[inline]
-    pub fn midpoint(&self, other: Point) -> Point {
+    pub fn midpoint(self, other: Point) -> Point {
         Point::new(0.5 * (self.x + other.x), 0.5 * (self.y + other.y))
     }
+
+    /// Euclidean distance.
+    #[inline]
+    pub fn distance(self, other: Point) -> f64 {
```
raphlinus: This is probably better written as simply `(self - other).hypot()`. Note that the `powf(2.)` is not the most efficient way to square, `powi(2)` is probably better (although I wouldn't be surprised if llvm were smart enough to optimize this anyway).

PR #32, opened 2019-06-21
Add Rect::contains

cmyr:


src/rect.rs
```diff
@@ -104,6 +104,12 @@ impl Rect {
         Point::new(0.5 * (self.x0 + self.x1), 0.5 * (self.y0 + self.y1))
     }
 
+    /// Returns `true` if `point` lies within `self`.
+    #[inline]
+    pub fn contains(&self, point: Point) -> bool {
+        point.x >= self.x0 && point.x <= self.x1 && point.y >= self.y0 && point.y <= self.y1
```
raphlinus: I prefer `< x1`, that way a tiling of the plane has one rect containing the point. That's the way winding it computed as well. But this is super-subtle; a case can be made a `Rect` contains the points on its boundary as well.

PR #33, opened 2019-06-24
Add Circle::contains

cmyr:


cmyr:
This makes me wonder if we should have some general concept of checking shapes & points for overlap. That is: my motivation for these functions is writing gui code; in this particular case it's about checking whether the mouse is over some circular region of the screen. This has different needs than `winding` (for instance, I think 'on the line' should probably count as a hit for my purposes) but is definitely similar in implementation.

For runebender, we're going to be doing tons of these sorts of fiddly little heuristic computations: does this mouse click close the current path, or just add a new point _quite close_ to the first point? etc. etc.

It may not make sense to have all of these in kurbo; I'm not sure.

raphlinus:
Basically, if it's a geometric primitive on shapes (as opposed to interfacing with platform graphics APIs) it belongs in kurbo. I'm very interested in a catalog of what's needed. I think many of the tests other than "insideness" are covered by [`nearest`](https://docs.rs/kurbo/0.2.0/kurbo/struct.BezPath.html#method.nearest), which for perhaps "inside baseball" reasons is implemented on `BezPath` but not `Shape`; I'm open to adding it as a convenience though.

cmyr:
I'm going to close this for now, pending some thinking.

PR #36, opened 2019-07-10
Document tolerance parameter in Shape

### LINKED ISSUE #35: Document `tolerance` argument
`Shape::to_bez_path` and `Shape::into_bez_path` both take a `tolerance` argument, but nowhere is it mentioned how the user might select an appropriate value, or what a reasonable choice might be in a given scenario.

raphlinus:
Closes #35

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -13,11 +13,29 @@ pub trait Shape: Sized {
     /// likely to be more efficient; in the general case, this
     /// allocates.
     ///
-    /// TODO: When GAT's land, the type of this can be changed to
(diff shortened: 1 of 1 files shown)


raphlinus:
> I might even consider exposing a `DEFAULT_TOLERANCE` const somewhere?

I'm reluctant to do that because it depends on the use case, and I have delusions of kurbo as being useful in scientific and mathematical applications. Also, it's perhaps something that people should tweak - in resource constrained environments maybe it should be bigger, and a default would discourage that.

PR #44, opened 2019-07-19
Add PathSeg::reverse & PathSeg::to_cubic

cmyr:


src/bezpath.rs
```diff
@@ -336,6 +336,25 @@ impl ParamCurveExtrema for PathSeg {
...
+            PathSeg::Line(Line { p0, p1 }) => PathSeg::Line(Line::new(*p1, *p0)),
+            PathSeg::Cubic(c) => PathSeg::Cubic(CubicBez::new(c.p3, c.p2, c.p1, c.p0)),
+            PathSeg::Quad(q) => PathSeg::Quad(QuadBez::new(q.p2, q.p1, q.p0)),
+        }
+    }
+
+    /// Convert this segment to a cubic bezier.
+    pub fn to_cubic(&self) -> CubicBez {
+        match *self {
+            PathSeg::Line(Line { p0, p1 }) => CubicBez::new(p0, p0, p1, p1),
```
raphlinus: This is something of an interesting question. I would expect "raise" to make the control points evenly spaced, i.e. t= 0, 1/3, 2/3, 1. But this is also a valid line, and I think a case can be made that downstream should be able to handle it. So I think I'm good with this.

src/bezpath.rs
```diff
@@ -336,6 +336,25 @@ impl ParamCurveExtrema for PathSeg {
 }
 
 impl PathSeg {
+    /// Returns a new `PathSeg` describing the same path as `self`, but with
+    /// the points reversed.
+    pub fn reverse(&self) -> PathSeg {
+        match self {
+            PathSeg::Line(Line { p0, p1 }) => PathSeg::Line(Line::new(*p1, *p0)),
+            PathSeg::Cubic(c) => PathSeg::Cubic(CubicBez::new(c.p3, c.p2, c.p1, c.p0)),
```
raphlinus: As the most incredibly minor style point, I'd put the match arms in Line, Quad, Cubic order (degree 1, 2, 3).

PR #45, opened 2019-07-19
Add Vec2::normalize

cmyr:
I'm not sure about either the function name or the docstring here.

### DIFF
--- a/src/vec2.rs
+++ b/src/vec2.rs
@@ -92,6 +92,15 @@ impl Vec2 {
     pub fn lerp(&self, other: Vec2, t: f64) -> Vec2 {
         *self + t * (other - *self)
     }
+
+    /// Returns a vector of magnitude 1.0 with the same angle as `self`; i.e.
+    /// a unit/direction vector.
+    ///
+    /// This produces `NaN` values when the magnitutde is `0`.
+    #[inline]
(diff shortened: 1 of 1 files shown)


src/vec2.rs
```diff
@@ -92,6 +92,13 @@ impl Vec2 {
     pub fn lerp(&self, other: Vec2, t: f64) -> Vec2 {
         *self + t * (other - *self)
     }
+
+    /// Returns a vector of magnitude 1.0 with the same angle as `self`; i.e.
+    /// a unit/direction vector.
```
raphlinus: I'd suggest saying that it will panic for zero-magnitude vectors. Other than that I think it's fine.

http://mathworld.wolfram.com/NormalizedVector.html is guidance for the terminology, and I think it checks out.

cmyr: hmm, I hadn't considered that. Do we have a policy of handling these sorts of situations? I think panicking is better than a pair of `NaN`s, but I might prefer returning (0., 0.) untouched? (Are there any other magic vectors with zero magnitude?)

raphlinus: Duh, this was a brainfart. We get inf in that case. I think that behavior is fine, but suggest it would be useful to document it. And I think I need to update #42, I make the claim it panics in `inverse` but that's clearly wrong.

PR #46, opened 2019-07-22
Version bump, deps bump, and docs cleanup

cmyr:
Anything else we'd like in before a release?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kurbo"
-version = "0.4.0"
+version = "0.5.0"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 license = "MIT/Apache-2.0"
 edition = "2018"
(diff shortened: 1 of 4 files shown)


raphlinus:
I don't think this needs to be a semver bump. That said, if we *are* doing one, this might be a good time to address #4. I think this'll take me a couple hours of solid work, and if I just do that, we can get it out of the way.

The patch itself looks fine to me.

cmyr:
Okay, I though you mentioned some behaviour change that would require a major bump.

re: #4, what approach have you decided on?

raphlinus:
The potential behavior change was for #4. I believe relative to 0.4 we currently have only new methods.

The plan for #4 is to codify the plan in [this comment](https://github.com/linebender/kurbo/issues/4#issuecomment-451250594). That basically involves writing tests to verify the behavior, and flipping signs when it doesn't match. One it'll almost certainly catch is the area of `Circle` with negative radius.

PR #47, opened 2019-07-22
Rationalize sign conventions

### LINKED ISSUE #4: Figure out sign conventions

raphlinus:
Document sign conventions in Shape. Add tests to confirm this is
indeed the behavior. Fix broken cases. Some doc touchup.

Fixes #4

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -393,12 +393,12 @@ impl PathSeg {
             if p.y < start.y || p.y >= end.y {
                 return 0;
             }
-            1
+            -1
         } else if end.y < start.y {
             if p.y < end.y || p.y >= start.y {
                 return 0;
             }
-            -1
+            1
         } else {
             return 0;
         };
@@ -490,8 +490,6 @@ impl Shape for BezPath {
     }
 
     /// Signed area.
-    ///
-    /// TODO: figure out sign convention, see #4.
     fn area(&self) -> f64 {
         self.elements().area()
     }
@@ -501,8 +499,6 @@ impl Shape for BezPath {
     }
 
     /// Winding number of point.
-    ///
-    /// TODO: figure out sign convention, see #4.
     fn winding(&self, pt: Point) -> i32 {
         self.elements().winding(pt)
     }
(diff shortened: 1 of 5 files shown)


behdad:
Anticlockwise -> counterclockwise? I never heard the former. Might be just me.

cmyr:
... I'm trying to come up with some joke about RTL clocks but it's falling flat 😒

raphlinus:
Just be thankful I didn't say "widdershins." But I can change it to "counter" if there's a consensus it's clearer.

behdad:
> ... I'm trying to come up with some joke about RTL clocks but it's falling flat

It's a matter of time...

jrus:
I like “anticlockwise” better than “counterclockwise”. I believe the former is more common in British than American English. But both are commonly used all over the place.

But also +1 for switching to “[sunwise](https://en.wikipedia.org/wiki/Sunwise)” and “[widdershins](https://en.wikipedia.org/wiki/Widdershins)” :P

cmyr:
It hurts me that the companion to 'dextrogyre' isn't 'sinistegyre'.

jrus:
http://argos.vu/wp-content/uploads/2016/05/1976_Laffoley_The-Levogyre.pdf

PR #48, opened 2019-07-23
Derive PartialEq for more primitive types

cmyr:
I at least want this for `Rect`, added some others while I was in there, not sure if there's anything tricky to consider here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kurbo"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 license = "MIT/Apache-2.0"
 edition = "2018"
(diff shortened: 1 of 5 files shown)


raphlinus:
One reason to be wary of `PartialEq` here is that floating point rounding error often defeats actual equality. That said, comparisons such as testing zero are entirely valid, so I'll approve this if you have at all a good case (ie if it's not clear that doing epsilon testing is better).

cmyr:
I can back this off to just be for `Rect`, but maybe it's bad. My use case is for things like `same`, where for my purposes any change is inequality.

raphlinus:
Ahhh, `same` is a pretty good motivator. However. One of the issues with `PartialEq` and floats is that it's never true for NaN values. The general solution is to `Arc` complex structures.

Another approach to pursue is implementing `same` (carefully, using bit equality) for a whole bunch of likely types, in the data module. I'd honestly lean towards that rather than this PR.

PR #49, opened 2019-07-25
Better fmt::Display impls for Point & Size

cmyr:
These will now respect formatter arguments for percision, width,
printing the sign, etcetera.

I added this because I wanted to be able to print lower-precision points, and that didn't work with the default format syntax. Now it does.

I _think_ this is a major version bump, since it would theoretically change the behaviour of existing code? In which case I propose holding it back until we next do that for other reasons.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kurbo"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 license = "MIT/Apache-2.0"
 edition = "2018"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "kurbo"
-version = "0.5.0"
+version = "0.6.0"
```
raphlinus: Looks good, but I don't see why this is a semver bump...

cmyr: yea I was unsure about this, see my (revised) comment above.

raphlinus: I guess this is something of a judgment call; it changes the behavior, but I'm pretty sure the contract for both Display and Debug is "human-readable" (the docs say "user-facing" which seems more ambiguous). My personal take is that code shouldn't be depending on the exact formatting, but otoh it also isn't going to hurt to be conservative here.

cmyr: I am happy here to defer to the practical reality that this will break nobody's code, and do a minor bump. This would also let us re-publish and see if the docs get fixed.

PR #16, opened 2019-06-10
Make Arc public.

RazrFalcon:
Arc to Cubic conversion is done via closure now. So it can be used on
any container.

Fixed Arc conversion.

PS: I'm switching from lyon to kurbo, since it's more lightweight.

PR #52, opened 2019-08-12
Fix incorrect SVG arc parsing

### LINKED ISSUE #51: Arc::from_svg_arc and Arc::to_cubic_beziers are broken again
It was fixed in https://github.com/linebender/kurbo/pull/16, but after updating from 0.2.3 to 0.5.1 it broke down again.

SVG path: `M 100 100 h 25 a 25 25 0 1 0 -25 25 z`

raphlinus:
The arc part of the fix in #16 got lost when the arc logic was refactored
in #18. This patch reinstates it and adds a basic test.

Fixes #51

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -18,6 +18,7 @@ impl Arc {
     ///
     /// The generated elemets can be append to an existing bezier path.
     pub fn append_iter(&self, tolerance: f64) -> ArcAppendIter {
+        let sign = self.sweep_angle.signum();
         let scaled_err = self.radii.x.max(self.radii.y) / tolerance;
         // Number of subdivisions per circle based on error tolerance.
         // Note: this may slightly underestimate the error for quadrants.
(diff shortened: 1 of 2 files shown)


RazrFalcon:
Can confirm that this patch fixes the issue.

cmyr:
The original issue brings up a good point about `PartialEq`. Do we want this for more types?

raphlinus:
I'm always reluctant to implement `PartialEq` for float-based types because it's so often not what you want, between roundoff and NaN concerns. But if people can make a case, it is one of the basic Rust traits.

RazrFalcon:
@raphlinus I'm actually using a custom traits for this. I call them [FuzzyEq](https://github.com/RazrFalcon/svgtypes/blob/14e116ec8af5a2c26c283b12d415f02d0e838452/src/traits.rs#L89) and [FuzzyZero](https://github.com/RazrFalcon/svgtypes/blob/master/src/traits.rs#L117). And they are [implemented](https://github.com/RazrFalcon/svgtypes/blob/14e116ec8af5a2c26c283b12d415f02d0e838452/src/number.rs#L13..L39) using float-cmp crate.

cmyr:
Yea, something like that may well make sense at some point.

PR #55, opened 2019-08-25
Serialize -> deserialize roundtrip, quadratic Bezier support in parser

jatentaki:
This PR
1. Enables a full serialize -> deserialize roundtrip by providing `BezPath::from_path_segments` and filling in some missing pieces.
2. Adds support for quadratic Bezier curves in the parser (it was silently converting them to `Line`s)
3. Causes unknown d-string commands to return `SvgParseError::UnknownCommand(offending_char))`. Adding this enum variant is the only (I think) breaking change in this PR. I think it is worth it but could be swapped out for `SvgParseError::Wrong`.
4. Features tests which could be also reworked to become a serialize/deserialize example
5. Derives `PartialEq` on `CubicBez`, `QuadBez`, `Line` and `PathSeg` which makes testing easier.

### DIFF
--- a/src/svg.rs
+++ b/src/svg.rs
@@ -3,7 +3,7 @@
 use std::f64::consts::PI;
 use std::io::Write;
 
-use crate::{Arc, BezPath, PathEl, Point, Vec2};
+use crate::{Arc, BezPath, PathEl, PathSeg, Point, Vec2, ParamCurve};
 
 // Note: the SVG arc logic is heavily adapted from https://github.com/nical/lyon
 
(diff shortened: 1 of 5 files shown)


src/svg.rs
```diff
@@ -373,4 +416,130 @@ mod tests {
...
+        let mut rng = rand::thread_rng();
+
+        for _ in 0..N_TESTS {
+            let vec = gen_random_path_sequence(&mut rng);
+            let ser = BezPath::from_path_segments(vec.iter().cloned()).to_svg();
+            let deser = BezPath::from_svg(&ser).expect("failed deserialization");
+
+            let deser_vec = deser.segments().collect::<Vec<PathSeg>>();
+
+            assert_eq!(vec, deser_vec);
```
raphlinus: Generally it's not recommended to rely on floating point equality as there can be roundoff errors. I'm wondering though whether the guarantees of the underlying operations are strong enough though.

Otherwise we'd need to implement some kind of approximate equality, which is not conceptually difficult but is pretty tedious.

jatentaki:
Honestly, I slapped `PartialEq` onto the structs just because it's fast and convenient and was surprised myself that the serialize -> deserialize procedure results in no numeric errors, but I didn't complain. If you don't mind, I would defer building a more robust comparison solution to this once either this test starts failing due to some further changes to the parser or it comes up naturally as a part of some other feature (right now I am working on curve fitting functionality and I may end up implementing this stuff anyway). [approx](https://docs.rs/approx/0.3.2/approx/) may come in handy when doing so.

My philosophy is that tests are not set in stone and we can change them later. In fact, I am more concerned about the other test cases for serialization, since they don't actually check if the resulting string is correct but rather check for equality with a very particular result. This means they enforce a particular pattern of whitespace and decimal representation - if the serializer produced `M 10.0 10.0` instead of `M 10 10`, the tests will fail. Because of that I was thinking if it maybe makes sense to omit those entirely, in favor of the serialize->deserialize roundtrip test (floating equality aside). Of course, this is not my crate so feel free to apply your own philosophy :)

jatentaki:
As I expected, over the course of work on fitting curves I implemented support for the `approx` crate. Please have a look at [this branch](https://github.com/jatentaki/kurbo/tree/approx_eq) of my fork and if you like it, I can merge it into this PR.

jatentaki:
@raphlinus sure, that's a good excuse :) should I do anything more before this PR can be merged, or are we just waiting for you to have the time?

raphlinus:
Sounds good, thanks! Didn't merge til now because I was away at a workshop.

PR #58, opened 2019-09-01
bugfix + a regression test

### LINKED ISSUE #57: QuadBez::nearest may return `NaN`
When working on curve fitting with quadratic Bezier curves, I ran into an issue which can be briefly described as the following test failing:

```rust
#[test]
fn quadbez_nearest_large_numbers() {
    let quad = QuadBez {
        p0: (0.0, 0.0).into(),
        p1: (32.5, 20.0).into(),
        p2: (65.0, 40.0).into(),
    };

   let p: Point = (90., 30.).into();

   let (t, d) = quad.nearest(p, 1e-6);

   assert!(!t.is_nan());
   assert!(!d.is_nan());
}
```

I will try to understand and maybe fix this issue myself, but I would expect the author of the implementation to have an easier time doing so :)

jatentaki:
Fixes #57. Three notes:
1. Testing if the coefficients are equal to `0.` prevents `NaN` from appearing in case of lower order polynomials but does not prevent significant precision issues if the coefficient is very close to 0.
2. Returning 0 as the solution when all coefficients are 0 is somewhat arbitrary, but I expect it to cause more good than harm in most cases.
3. I made the `solve_poly_3` method `pub(crate)`. In general, I don't see the reason to expose rootfinding procedures from a curve crate - in other words, I would also make `solve_cubic` and `solve_quadratic` `pub(crate)` and hide the `common` module from the public API altogether.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -2,6 +2,29 @@
 
 use arrayvec::ArrayVec;
 
+/// Find real roots of a polynomial up to order of 3.
+/// Returns values of x for which c0 + c1 x + c2 x² + c3 x³ = 0, allowing for the coefficients
+/// to be equal to 0.
+pub(crate) fn solve_poly_3(c0: f64, c1: f64, c2: f64, c3: f64) -> ArrayVec<[f64; 3]> {
+    if c3 != 0. {
+        return solve_cubic(c0, c1, c2, c3);
+    };
+
+    let mut roots = ArrayVec::new();
+
+    if c2 != 0. {
+        for root in solve_quadratic(c0, c1, c2) {
+            roots.push(root);
+        }
+    } else if c1 != 0. {
+        roots.push(-c0 / c1);
+    } else if c0 == 0. {
+        roots.push(0.);
+    };
+
+    return roots;
+}
+
 /// Find real roots of cubic equation.
 ///
 /// Assumes c3 is nonzero.
--- a/src/quadbez.rs
+++ b/src/quadbez.rs
@@ -4,7 +4,7 @@ use std::ops::{Mul, Range};
 
 use arrayvec::ArrayVec;
 
-use crate::common::solve_cubic;
+use crate::common::solve_poly_3;
 use crate::MAX_EXTREMA;
 use crate::{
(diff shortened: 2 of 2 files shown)


src/common.rs
```diff
@@ -2,6 +2,29 @@
 
 use arrayvec::ArrayVec;
 
+/// Find real roots of a polynomial up to order of 3.
+/// Returns values of x for which c0 + c1 x + c2 x² + c3 x³ = 0, allowing for the coefficients
+/// to be equal to 0.
+pub(crate) fn solve_poly_3(c0: f64, c1: f64, c2: f64, c3: f64) -> ArrayVec<[f64; 3]> {
```
raphlinus: Generally this looks good to me, but I'm still not sure it's as numerically robust as possible. Coefficients of exactly 0 are handled well here, but extremely tiny coefficients still are not. There's a TODO in `solve_quadratic` for this.

src/common.rs
```diff
@@ -2,6 +2,29 @@
...
+    if c3 != 0. {
+        return solve_cubic(c0, c1, c2, c3);
+    };
+
+    let mut roots = ArrayVec::new();
+
+    if c2 != 0. {
```
raphlinus: *Most* of this logic is redundant with checks already in `solve_quadratic`, but I see a special case to handle all coefficients 0. This is interesting, because all values of x satisfy the equation in this case. I'm wondering whether it would be better to move this logic to solve_quadratic.

In any case, returning a single "root" in the all-zero coefficient case should probably be added to the documentation.

raphlinus:
I made the quadratic solver numerically robust, and also return `[0.0]` in the degenerate case, in #59. We can try to do similar magic for the cubic solver, or revise this PR lightly, handling the c3=0 case only for now, which should fix the immediate problem.

cmyr:
What's the status of this?

raphlinus:
I believe the content of this PR is subsumed by #59 and #60, so this can be closed.

cmyr:
@jatentaki feel free to reopen if we've overlooked anything!

PR #59, opened 2019-09-03
Make quadratic solver numerically robust

raphlinus:
Part of the work decribed in review of #58

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -45,24 +45,54 @@ pub fn solve_cubic(c0: f64, c1: f64, c2: f64, c3: f64) -> ArrayVec<[f64; 3]> {
 /// Find real roots of quadratic equation.
 ///
 /// Returns values of x for which c0 + c1 x + c2 x² = 0.
+///
+/// This function tries to be quite numerically robust. If the equation
+/// is nearly linear, it will return the root ignoring the quadratic term;
+/// the other root might be out of representable range. In the degenerate
+/// case where all coefficients are zero, so that all values of x satisfy
+/// the equation, a single `0.0` is returned.
 pub fn solve_quadratic(c0: f64, c1: f64, c2: f64) -> ArrayVec<[f64; 2]> {
     let mut result = ArrayVec::new();
-    // TODO: make this more robust to very small values.
-    if c2 == 0.0 {
-        if c1 == 0.0 {
-            return result;
+    let sc0 = c0 * c2.recip();
+    let sc1 = c1 * c2.recip();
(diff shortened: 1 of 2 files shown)


raphlinus:
Thanks, merging, but more work is remaining to actually fix #57; I was hoping to hear feedback from @jatentaki (author of #58) but if not, I'll have a crack at it which will attempt to take care of zero and very small cubic coefficients but does not (as in this patch for quadratics) be bulletproof against all inputs.

src/common.rs
```diff
@@ -45,24 +45,54 @@ pub fn solve_cubic(c0: f64, c1: f64, c2: f64, c3: f64) -> ArrayVec<[f64; 3]> {
...
+    let sc1 = c1 * c2.recip();
+    if !sc0.is_finite() || !sc1.is_finite() {
+        // c2 is zero or very small, treat as linear eqn
+        let root = -c0 / c1;
+        if root.is_finite() {
+            result.push(root);
+        } else if c0 == 0.0 && c1 == 0.0 {
+            // Degenerate case
+            result.push(0.0);
         }
```
jatentaki: If both `if` and `elif` fail, we return an empty result set. Can this condition occur only if we have `c1 == c2 == 0` and `c0 != 0.` (or, in other words, when we ask for roots of a non-zero constant)? I am asking instead of checking because it's late and I am tired. It's just that whenever I see this kind of `if` statements, it makes me wonder if the author did check what happens when all conditions fail.

src/common.rs
```diff
@@ -45,24 +45,54 @@ pub fn solve_cubic(c0: f64, c1: f64, c2: f64, c3: f64) -> ArrayVec<[f64; 3]> {
...
+        } else if arg == 0.0 {
+            result.push(-0.5 * sc1);
+            return result;
+        }
+        // See https://math.stackexchange.com/questions/866331
+        -0.5 * (sc1 + arg.sqrt().copysign(sc1))
+    };
+    let root2 = sc0 / root1;
+    if root2.is_finite() {
+        // Sort just to be friendly and make results deterministic.
```
jatentaki: I like that idea, it could be reflected in the docstring

src/common.rs
raphlinus: Yes. This is basically `c == 0` which has no solutions when `c != 0` (when written this way, it's kinda obvious, isn't it?)

I did a modest amount of testing on a local branch, but ideally there would be a test suite that would cover all the edge cases. Unfortunately for cubics it's a lot harder, so my approach will probably be even more ad-hoc.

In any case, thanks for raising the isssue!

PR #61, opened 2019-09-18
Add inverse affine transformation

raphlinus:


raphlinus:
This isn't urgent but will come in handy - when applying a transformation globally to a font, when a component has transform T, then you want A * T * Ainv.

src/affine.rs
```diff
@@ -64,6 +64,24 @@ impl Affine {
...
         self.0
     }
+
+    /// Compute the determinant of this transform.
+    pub fn determinant(self) -> f64 {
+        self.0[0] * self.0[3] - self.0[1] * self.0[2]
+    }
+
+    /// Compute the inverse transform.
+    pub fn inverse(self) -> Affine {
```
cmyr: Maybe we should document that this will be all `NaN` if the determinant is 0?

raphlinus:
It's complicated. The case where the determinant is exactly zero is kinda easy, this is where you'd return a `None`. The more complicated case is when the determinant is very very small, which can happen due to roundoff. Then it's possible for the inverse of the determinant to be finite, but trigger overflow when multiplied by the coefficients.

My sense is that it's best to provide reasonable primitives and assume the caller knows what they're doing. This is exactly why I've exposed the determinant as part of the API - the caller can decide what's a "reasonable" value and make a decision based on that. For example, in font usage, a value less than, say 1e-3 is a pretty good sign that you don't want to actually do the inverse, but if you're doing projections for, say, VR, you might want to recover the inverse transform except when the plane is (almost) exactly on-edge.

So I think this API is defensible, and we can add other stuff (like `try_inv`) as a demonstrable need arises.

PR #64, opened 2019-10-16
Add Insets type

### LINKED ISSUE #63: Add EdgeInsets type
This would be a type that could be used to describe the distance between the edges of two rectangles.

```rust
struct EdgeInsets {
    top: f64,
    right: f64,
    bottom: f64,
    left: f64,
}
```

operators: `rect +- insets => rect`, and possibly `rect - rect => insets`.

from/into: `From<f64>` is uniform insets; `From<(f64, f64)>` is uniform top/bottom and left/right. `From<(f64, f64, f64, f64)>` is as expected.

cmyr:
This is a minimal implementation, that tries to conform as much as
possible with the existing Rect implmentation and API.

closes #63

-----

### Things to note:

- nonnegative: I'm not sure about the behaviour in the case of negative size; this PR takes the position that we will always normalize (`Rect::abs`) a rectangle before applying insets; I'm not sure if that makes sense or not.
- `Add` not `Sub`: I've implemented `Add` for `Rect` and `Insets`, but not `Sub`; my thinking is that if you want `Sub` you can just create negative insets and add them. Thinking this through, though, it might be nice if we had `Insets::invert`, that will return the inverse insets?


src/insets.rs
```diff
@@ -0,0 +1,194 @@
...
+            x0: d,
+            y0: d,
+            x1: d,
+            y1: d,
+        }
+    }
+
+    /// New insets with uniform values along each axis.
+    #[inline]
+    pub const fn uniform_xy(x: f64, y: f64) -> Insets {
```
raphlinus: "centered" might be a more vivid name for this.

src/insets.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// use kurbo::Insets;
+    ///
+    /// let insets = Insets::uniform_xy(3., 8.);
+    /// assert_eq!(insets.x_value(), 6.);
+    ///
+    /// let insets = Insets::new(5., 0., -12., 0.,);
+    /// assert_eq!(insets.x_value(), -7.);
+    /// ```
+    #[inline]
+    pub fn x_value(self) -> f64 {
```
raphlinus: I like the fact that `(r + inset).width() = r.width() + inset.x_value()` is an invariant (modulo FP roundoff of course).

src/insets.rs
```diff
@@ -0,0 +1,194 @@
...
+/// the center of the rect to that edge. For instance, with an inset of `2.0` on
+/// the x-axis, a rectange at `(0.0, 0.0)` with that inset added will be at
+/// `(-2.0, 0.0)`
+///
+/// Put alternatively, a positive inset represents increased distance from center,
+/// and a negative inset represents decreased distance from center.
+///
+/// # Examples
+///
+/// Positive insets produce larger rectangles:
```
raphlinus: I would say "when added" just to be clear here, especially as I recommend implementing Sub also.

src/insets.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// let rect = Rect::from_origin_size((0., 0.,), (10., 10.,));
+/// let insets = Insets::uniform_xy(3., 0.,);
+///
+/// let inset_rect = rect + insets;
+/// assert_eq!(inset_rect.width(), 16.0, "10.0 + 3.0 * 2");
+/// assert_eq!(inset_rect.x0, -3.0);
+/// ```
+///
+/// Negative insets produce smaller rectangles:
```
raphlinus: Same.

src/insets.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// let rect = Rect::from_origin_size((0., 0.,), (10., 10.,));
+/// let insets = Insets::uniform_xy(-3., 0.,);
+///
+/// let inset_rect = rect + insets;
+/// assert_eq!(inset_rect.width(), 4.0, "10.0 - 3.0 * 2");
+/// assert_eq!(inset_rect.x0, 3.0);
+/// ```
+///
+/// Insets ignore negative width & height when computing new rectangles.
```
raphlinus: I think "ignore" is a little unclear here. I might say. "Insets can result in a negative width or height when the original rectangle is smaller than the inset value."

src/insets.rs
cmyr: struggling with this. "centered" doesn't quite feel right, because the insets themselves aren't centered; it's more a description of their effects. I also want to include `xy` because it's different from CSS, where padding specified with two values is `yx`.

cmyr:
Good notes, I've added those impls and clarified the docs somewhat.

PR #69, opened 2019-10-29
Add Rect::min_x/y & max_x/y

cmyr:
Was just doing some drawing code in runebender and wanted this; in the general case you could use `x0/x1/y0/y1` directly, but these methods are more robust when dealing with negative width or height, and I find them more readable when writing UI code.


cmyr:
yea there are definitely ways to do this currently, this is 100% about ergonomics.

PR #71, opened 2019-10-29
Update arrayvec and rand in Cargo.toml

cmyr:


cmyr:
this would require us to increase our minimum supported rust version. It's unclear to me if that is worth it.

raphlinus:
I don't have any serious problem with bumping msrv at this point.

PR #73, opened 2019-11-05
Add `Affine::transform_rect_bbox`

raphlinus:
Compute the enclosing bounding box of a Rect after affine transformation. This came up during the discussion of https://github.com/xi-editor/druid/pull/243

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -2,7 +2,7 @@
 
 use std::ops::{Mul, MulAssign};
 
-use crate::{Point, Vec2};
+use crate::{Point, Rect, Vec2};
 
 /// A 2D affine transform.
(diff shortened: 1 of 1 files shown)


src/affine.rs
```diff
@@ -91,6 +91,21 @@ impl Affine {
             inv_det * (self.0[1] * self.0[4] - self.0[0] * self.0[5]),
         ])
     }
+
+    /// Compute the bounding box of a transformed rectangle.
+    ///
+    /// Returns the minimal `Rect` that encloses given `Rect` after affine transformation.
```
cmyr: maybe _the_ given `Rect`? I'm actually not sure if this is a typo or a style choice.

PR #74, opened 2019-11-08
Add Vec2::round/floor/ceil

cmyr:


src/vec2.rs
```diff
@@ -105,6 +105,47 @@ impl Vec2 {
     pub fn normalize(self) -> Vec2 {
         self / self.hypot()
     }
+
+    /// Returns a new `Vec2` with each of `x` and `y` rounded to the nearest integer.
+    /// Half-way cases are arounded away from `0.0`.
```
raphlinus: "rounded". But I'm not sure we should nail this down in the docs. I'd prefer round-to-even, it's just not easily available in Rust right now. (Of course, we can always change the docs when we change the impl)

PR #82, opened 2020-01-17
Drive-by refactoring

Michael-F-Bryan:
These are a couple things I've picked up while reading through the `kurbo` source code.

I've tried to put each change into its own commit so they can be easily reverted if not desired.

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -6,10 +6,16 @@ use std::f64::consts::{FRAC_PI_2, PI};
 /// A single arc segment.
 #[derive(Clone, Copy, Debug)]
 pub struct Arc {
+    /// The arc's centre point.
     pub center: Point,
+    /// The arc's radii, where the vector's x-component is the radius in the
+    /// positive x direction after applying `x_rotation`.
     pub radii: Vec2,
+    /// The start angle in radians.
     pub start_angle: f64,
+    /// The angle between the start and end of the arc, in radians.
     pub sweep_angle: f64,
+    /// How much the arc is rotated, in radians.
     pub x_rotation: f64,
 }
 
@@ -60,6 +66,7 @@ impl Arc {
 }
 
 #[doc(hidden)]
+#[derive(Debug)]
 pub struct ArcAppendIter {
     idx: usize,
 
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -66,7 +66,6 @@ impl Arc {
 }
 
 #[doc(hidden)]
-#[derive(Debug)]
 pub struct ArcAppendIter {
     idx: usize,
 
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -20,7 +20,8 @@ pub struct BezPath(Vec<PathEl>);
 /// A valid path has `Moveto` at the beginning of each subpath.
 #[derive(Clone, Copy, Debug)]
 pub enum PathEl {
-    /// Move directly to the point without drawing anything.
+    /// Move directly to the point without drawing anything, starting a new
+    /// subpath.
     MoveTo(Point),
     /// Draw a line from the current location to the point.
     LineTo(Point),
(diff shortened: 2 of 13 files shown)


Michael-F-Bryan:
@raphlinus would you be able to have a look at this and tell me what you think?

The main area I'm concerned about is docs. I added the `#[deny(missing_docs)]` link, but that meant I needed to fill in the missing docs. In particular, my interpretation of arc radii (both SVG and `arc.rs`) may not be 100% correct.

raphlinus:
Generally this looks good to me. The one thing I might push back on is the Debug derives on iterator types. I feel those are internal details of the functions that produce the iterators, so very unlikely to be useful. I'm also (perhaps overly sensitive) to compile time, as I hope this to be a basic, foundational crate, but admit to not having measured the impact of Debug derives.

src/arc.rs
```diff
@@ -6,10 +6,16 @@ use std::f64::consts::{FRAC_PI_2, PI};
 /// A single arc segment.
 #[derive(Clone, Copy, Debug)]
 pub struct Arc {
+    /// The arc's centre point.
```
raphlinus: I believe this is right, but it's no longer fresh in my mind, and the logic was adapted from a number of sources (including lyon).

src/bezpath.rs
```diff
@@ -20,18 +20,26 @@ pub struct BezPath(Vec<PathEl>);
 /// A valid path has `Moveto` at the beginning of each subpath.
 #[derive(Clone, Copy, Debug)]
 pub enum PathEl {
+    /// Move directly to the point without drawing anything.
```
raphlinus: Maybe add ", starting a new subpath."

src/svg.rs
```diff
@@ -10,11 +12,18 @@ use crate::{Arc, BezPath, ParamCurve, PathEl, PathSeg, Point, Vec2};
...
+    /// The arc's end point.
     pub to: Point,
+    /// The arc's radii, where the vector's x-component is the radius in the
+    /// positive x direction after applying `x_rotation`.
     pub radii: Vec2,
+    /// How much the arc is rotated, in radians.
     pub x_rotation: f64,
+    /// Does this arc sweep through more than π radians?
     pub large_arc: bool,
+    /// Determines if the arc should begin moving at positive angles.:wq
```
raphlinus: Vi user? :)

src/svg.rs
```diff
@@ -47,141 +56,156 @@ impl BezPath {
...
+                )?,
+                PathEl::ClosePath => write!(writer, "Z")?,
             }
         }
-        String::from_utf8(result).unwrap()
+
+        Ok(())
     }
 
+    /// Try to parse a bezier path from an SVG path element.
```
raphlinus: I'm wondering if we should say that we don't guarantee that all SVG paths are handled correctly (this was discussed in #78), in the interest of full disclosure. It's intended for cases (such as vector icons) where the user controls the source of paths, and is not intended as a replacement for a general, robust SVG parser.

src/svg.rs
Michael-F-Bryan: Guilty.

Michael-F-Bryan:
@raphlinus I believe I've addressed all the points you raised.

I was tossing up whether to add a `FromStr` impl for `BezPath` which just defers to `BezPath::from_svg()`. Eventually I decided against it because that implies there's only one way to encode a `BezPath` as text (SVG), but from what I can tell a `BezPath` is just a list of draw instructions and has nothing to do with SVG... Does that sound correct?

cmyr:
@Michael-F-Bryan your last point is correct, there are many different possible ways a bezier might be represented in text outside of the SVG case, so I think `FromStr` would be overstepping.

Michael-F-Bryan:
@raphlinus is there anything else I need to do for this PR?

PR #89, opened 2020-03-03
Implement Add/Sub traits for Size

cmyr:
we lost this when we moved away from using Vec2 for everythin, but I've wanted it a few times.


cmyr:
I'm willing to consider that this does not actually make sense, in which case I would propose `Add<Insets> for Size`.

PR #90, opened 2020-03-10
Make Line::new take two impl Intos

cmyr:
This lets you do, `Line::new(Point::ORIGIN, (10., 42.))`, which
would previously fail because they are different types.


cmyr:
there are two other little fixup prs sitting in the queue if you're having trouble concentrating ;)

PR #91, opened 2020-03-16
Fixup RoundedRect signatures & docs

cmyr:
This was motivated by noticing that RoundedRect::from_origin_size
took a Vec2 instead of a Size. Now they match the equivalent methods
on Rect.

Included some doc tweaks while I was in there.

--- I thought this was PR'd and merged already but clearly I'd forgotten 😒


cmyr:
going to merge this, it is uncontroversial

PR #93, opened 2020-03-16
Add ceil/floor/trunc/expand to various shapes.

xStrom:
This PR adds:
* `ceil` and `floor` methods to `Point` and `Size`
* `kurbo::common::FloatExt` trait which adds the `expand` method to `f32` and `f64`
* `expand` and `trunc` methods to `Rect`

All of these methods will make it easier to quantize druid layout to integers.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,6 +4,45 @@
 
 use arrayvec::ArrayVec;
 
+/// Adds convenience methods to `f32` and `f64`.
+pub trait FloatExt<T> {
+    /// Rounds to the nearest integer away from zero,
+    /// unless the provided value is already an integer.
+    ///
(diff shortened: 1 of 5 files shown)


xStrom:
Pushed some more changes:
* Unified `ceil`, `floor`, `trunc`, `expand` implementations on `Point`, `Vec2`, `Size`, `Rect`
* Rewrote the docs for these methods so that they use the same language
* Added examples/tests to each of the methods

cmyr:
@xStrom if you want a release let me know :)

xStrom:
@cmyr I'll probably do some work using these soon. So having a new release out by the end of the week would be useful. If no other changes are planned for merge then could of course do the release faster. Thanks!

cmyr:
will try and clear out the queue and can then do one whenever.

PR #94, opened 2020-03-17
Add Size::max/min_side

cmyr:


src/size.rs
```diff
@@ -25,6 +25,32 @@ impl Size {
...
+    /// Returns the min of `width` and `height`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use kurbo::Size;
+    /// let size = Size::new(-10.5, 42.0);
+    /// assert_eq!(size.min_size() -10.5);
+    /// ```
+    pub fn min_size(self) -> f64  {
```
raphlinus: Probably should be `min_side` to be consistent?

cmyr: rups

PR #97, opened 2020-03-27
Adds optional serde support.

### LINKED ISSUE #54: serde support?
Is there any specific consideration that this crate does not have conditional serde serialization support? by the way great work.

jneem:
All public types are serializable and deserializable except for
SvgParseError.

Closes #54.


jneem:
I'm happy to use git for now, but thanks for asking!

PR #98, opened 2020-04-07
Be more permissive in segments_of_slice.

jneem:
My use-case is this: I have a druid program that allows for some free-hand drawing. I want to find a bounding box for the part of the curve that was drawn since the last frame. It would be very convenient if I could just slice the current curve and call `bounding_box` on the slice, but it panics if the part of the curve since the last frame doesn't start with a `MoveTo`.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -118,7 +118,10 @@ impl BezPath {
     fn segments_of_slice<'a>(slice: &'a [PathEl]) -> BezPathSegs<'a> {
         let first = match slice.get(0) {
             Some(PathEl::MoveTo(ref p)) => *p,
-            Some(_) => panic!("First element has to be a PathEl::Moveto!"),
+            Some(PathEl::LineTo(ref p)) => *p,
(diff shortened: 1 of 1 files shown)


src/bezpath.rs
```diff
@@ -118,7 +118,10 @@ impl BezPath {
     fn segments_of_slice<'a>(slice: &'a [PathEl]) -> BezPathSegs<'a> {
         let first = match slice.get(0) {
             Some(PathEl::MoveTo(ref p)) => *p,
-            Some(_) => panic!("First element has to be a PathEl::Moveto!"),
+            Some(PathEl::LineTo(ref p)) => *p,
```
raphlinus, line 121: Incidentally, these `ref`s are no longer required, I think that changed with NLL.

jneem: I guess it's the "match ergonomics" business. I didn't really follow that, tbh...

raphlinus:
Thanks!

PR #100, opened 2020-04-10
Fix pathological behaviour when computing the arc length of a degenerate bezier

### LINKED ISSUE #99: CubicBezier.arclen exhibits pathological behaviour when passed a bezier curve which is just a point
When passed a bezier curve in which all control points are identical then the est_gauss9_error method will always return NaN and the function will recursively subdivide the curve 16 times. This leads to 2^16 = 65536 evaluations of the `rec` function which is very very slow (comparatively).

The method should check for degenerate bezier curves before trying to calculate the length of them.

HalfVoxel:
Before
```
test bench_cubic_arclen_1e_4            ... bench:         281 ns/iter (+/- 28)
test bench_cubic_arclen_1e_5            ... bench:         281 ns/iter (+/- 20)
test bench_cubic_arclen_1e_6            ... bench:         608 ns/iter (+/- 26)
test bench_cubic_arclen_degenerate      ... bench:   5,914,628 ns/iter (+/- 243,332)
```

After
```
test bench_cubic_arclen_1e_4            ... bench:         277 ns/iter (+/- 8)
test bench_cubic_arclen_1e_5            ... bench:         280 ns/iter (+/- 38)
test bench_cubic_arclen_1e_6            ... bench:         600 ns/iter (+/- 41)
test bench_cubic_arclen_degenerate      ... bench:           6 ns/iter (+/- 0)
```

Fixes #99.

### DIFF
--- a/src/cubicbez.rs
+++ b/src/cubicbez.rs
@@ -157,7 +157,19 @@ impl ParamCurveArclen for CubicBez {
                 rec(&c0, accuracy * 0.5, depth + 1) + rec(&c1, accuracy * 0.5, depth + 1)
             }
         }
-        rec(self, accuracy, 0)
+
+        // Check if the bezier curve is degenerate, or almost degenerate
+        // A degenerate curve where all points are identical will cause infinite recursion in the rec function (well, until MAX_DEPTH at least) in all branches.
+        // This test will in addition be true if the bezier curve is just a simple line (i.e. p0=p1 and p2=p3).
+        // The constant 0.5 has not been mathematically proven to be small enough, but from empirical tests
+        // a value of about 0.87 should be enough. Thus 0.5 is a conservative value.
+        // See https://github.com/linebender/kurbo/pull/100 for more info.
+        if (self.p1 - self.p0).hypot2() + (self.p2 - self.p3).hypot2() <= 0.5 * accuracy * accuracy
+        {
+            (self.p0 - self.p3).hypot()
+        } else {
+            rec(self, accuracy, 0)
+        }
     }
 }
 
--- a/benches/cubic_arclen.rs
+++ b/benches/cubic_arclen.rs
@@ -25,6 +25,12 @@ fn bench_cubic_arclen_1e_6(b: &mut Bencher) {
 }
 
 #[bench]
+fn bench_cubic_arclen_degenerate(b: &mut Bencher) {
+    let c = CubicBez::new((0.0, 0.0), (0.0, 0.0), (0.0, 0.0), (0.0, 0.0));
+    b.iter(|| test::black_box(c).arclen(1e-6))
+}
+
+#[bench]
 fn bench_cubic_arclen_1e_7(b: &mut Bencher) {
     let c = CubicBez::new((0.0, 0.0), (1.0 / 3.0, 0.0), (2.0 / 3.0, 1.0), (1.0, 1.0));
     b.iter(|| test::black_box(c).arclen(1e-7))


src/cubicbez.rs
```diff
@@ -157,7 +157,15 @@ impl ParamCurveArclen for CubicBez {
                 rec(&c0, accuracy * 0.5, depth + 1) + rec(&c1, accuracy * 0.5, depth + 1)
             }
         }
-        rec(self, accuracy, 0)
+
+        // Check if the bezier curve is degenerate, or almost degenerate
+        // A degenerate curve where all points are identical will cause infinite recursion in the rec function (well, until MAX_DEPTH at least) in all branches.
+        const SMALL : f64 = 0.000000001;
```
raphlinus, line 163: This seems like a good general approach, but I think it would be better if this were keyed off the accuracy; it's certainly possible for this to give a wrong answer if everything is scaled down by (say) 1e-9.

HalfVoxel: Yeah, that's probably good.
I have pushed a commit which does this.

src/cubicbez.rs
```diff
@@ -157,7 +157,15 @@ impl ParamCurveArclen for CubicBez {
                 rec(&c0, accuracy * 0.5, depth + 1) + rec(&c1, accuracy * 0.5, depth + 1)
             }
         }
-        rec(self, accuracy, 0)
+
+        // Check if the bezier curve is degenerate, or almost degenerate
+        // A degenerate curve where all points are identical will cause infinite recursion in the rec function (well, until MAX_DEPTH at least) in all branches.
+        let small = accuracy * 0.00000001;
```
raphlinus, line 163: Pretty sure this will still lead to incorrect results in some cases, mostly because you're comparing a squared distance with a distance.

Thinking about this, I propose:

`(self.p1 - self.p0).hypot2() + (self.p2 - self.p3).hypot2() <= CONST * accuracy * accuracy`

This has the side effect of short-cutting the case where p0 = p1 and p2 = p3. The only question is, what is CONST? I think 1.0 is adequate, but it would be better to do mathematical analysis.

HalfVoxel: Hmmm.. I think 0.5 will work.
The upper bound for the length of the curve is

```|p1-p0| + |p2 - p1| + |p3 - p2|```

Given fixed values of p0 and p3 (assume wlog that they lie on a straight horizontal line) and your constraint

```|p1 - p0|^2 + |p2 - p3|^2 <= C^2```

Then the largest value of `|p2 - p1|` happens (I postulate) when `p1` is `C/sqrt(2)` units to the left of `p0` and `p2` is `C/sqrt(2)` units to the right of `p2`.

Or in other words `|p2 - p1| <= |p3 - p0| + C*(2*1.0/sqrt(2)) = |p3 - p0| + C*2*sqrt(2)`.

Thus the error

`||p3 - p0| - (|p1-p0| + |p2 - p1| + |p3 - p2|)| <= C*4/sqrt(2)`

Therefore I think we must set `C*4/sqrt(2) = precision` and thus `C^2 = precision*precision*1/8`.

Comment (raphlinus) [reply src/cubicbez.rs:163]: Ok, I've been properly nerd-sniped, and wanted to test this empirically. Here's my [code](https://gist.github.com/raphlinus/37a0ed567a9dcc92f16bbecca802ab4a).

From random testing, CONST = 0.8714 seems to do the trick, but it's hard to know if that's a hard bound. Interestingly, the worst case behavior seems to be a "tear drop" shape where p0 and p3 nearly coincide.

```html
<!DOCTYPE html>
<html>
<body>

<svg height="400" width="450">
  <path d="M165.61722290078575 43.216428638303974C5.65289484019762 101.67969576236118 68.23986815921117 182.9261339463449 165.61722290078575 43.216428638303974" stroke="blue" stroke-width="2" fill="none" />
</svg>

</body>
</html>
```

So it's certainly the case that 0.5 is a conservative bound, as you suggest :)

HalfVoxel: Haha :p. That is an interesting worst case, I wouldn't have expected that.
And my 1/8 is very very conservative in that case :p.

I don't think it matters that much which constant is chosen as long as it is good enough. I'd propose 0.5 in that case since it seems to be well under the empirical bound and it's a nice number. However a comment should explain the reasoning.

raphlinus: That sounds like a good plan, I am happy with 0.5. Feel free to make the change and I will merge it with thanks.

HalfVoxel:
Updated the code.

HalfVoxel:
Pushed btw.

PR #102, opened 2020-04-10
Add ParamCurveDistanceEval trait and implement for CubicBez

HalfVoxel:
This adds a trait that allows one to evaluate a curve based on an arc length distance from the start point. This is very useful when you want to get a number of evenly spaced points on a curve.

Currently I have only implemented this for the cubic bezier.

The implementation is based on the arclen method. Some helper methods should possibly be moved out to avoid code duplication.

### DIFF
--- a/src/cubicbez.rs
+++ b/src/cubicbez.rs
@@ -8,8 +8,8 @@ use arrayvec::ArrayVec;
 use crate::common::solve_quadratic;
 use crate::common::GAUSS_LEGENDRE_COEFFS_9;
 use crate::{
-    Affine, ParamCurve, ParamCurveArclen, ParamCurveArea, ParamCurveCurvature, ParamCurveDeriv,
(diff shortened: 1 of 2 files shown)


HalfVoxel:
I just implemented it as a core trait as that seemed to be what you had done for all the other functionality, but it can be implemented on top of ParamCurveArclen. A default implementation using a binary search over the arc length would not be very efficient, but it would work for all types implementing that trait.

raphlinus:
I'm looking at the backlog of PRs and issues. From a quick look, this seems to be the same as the `inv_arclen` method. Am I missing something?

HalfVoxel:
Might be that inv_arclen was added after I made this PR? It was three years ago, after all.

HalfVoxel:
Feel free to close

raphlinus:
Thanks!

PR #104, opened 2020-04-19
Helper to construct Rect from center and size.

derekdreery:
Just a simple helper for when you want a rectangle centred at a given point. I even use the USA spelling of center! :P

### DIFF
--- a/src/rect.rs
+++ b/src/rect.rs
@@ -9,6 +9,7 @@ use crate::{Insets, PathEl, Point, RoundedRect, Shape, Size, Vec2};
 /// A rectangle.
 #[derive(Clone, Copy, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(test, derive(PartialEq))]
 pub struct Rect {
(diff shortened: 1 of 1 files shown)


src/rect.rs
```diff
@@ -9,6 +9,7 @@ use crate::{Insets, PathEl, Point, RoundedRect, Shape, Size, Vec2};
 /// A rectangle.
 #[derive(Clone, Copy, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(test, derive(PartialEq))]
```
raphlinus, line 12: I've been resisting adding `PartialEq` to the basic types because of the usual caveats about floating point comparison, but agree it makes sense in your test. I'm a bit uneasy about making it conditional, though - I think it would be more user-friendly to just derive it. I'll start a discussion on the Zulip.

Otherwise, looks great!

raphlinus: After some discussion on Zulip, we've decided to turn on PartialEq. Feel free to update this PR to take off the `cfg_attr` for `Rect`, and we can have a followup PR to turn it on for the other basic types.

PR #105, opened 2020-04-23
Add BezPath flattening

raphlinus:
This implements the fancy flattening algorithm that I've been blogging about. It also simplifies the cubic to quadratic conversion and cleans up iterator conversions around `BezPath`.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -117,7 +117,58 @@ impl BezPath {
 
     /// Flatten the path, invoking the callback repeatedly.
     ///
-    /// The callback is only invoked with lines.
+    /// Flattening is the action of approximating a curve with a succession of line segments.
+    ///
+    /// <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 120 30" height="30mm" width="120mm">
+    ///   <path d="M26.7 24.94l.82-11.15M44.46 5.1L33.8 7.34" fill="none" stroke="#55d400" stroke-width=".5"/>
+    ///   <path d="M26.7 24.94c.97-11.13 7.17-17.6 17.76-19.84M75.27 24.94l1.13-5.5 2.67-5.48 4-4.42L88 6.7l5.02-1.6" fill="none" stroke="#000"/>
+    ///   <path d="M77.57 19.37a1.1 1.1 0 0 1-1.08 1.08 1.1 1.1 0 0 1-1.1-1.08 1.1 1.1 0 0 1 1.08-1.1 1.1 1.1 0 0 1 1.1 1.1" color="#000" fill="none" stroke="#030303" stroke-linecap="round" stroke-opacity=".5"/>
+    ///   <path d="M77.57 19.37a1.1 1.1 0 0 1-1.08 1.08 1.1 1.1 0 0 1-1.1-1.08 1.1 1.1 0 0 1 1.08-1.1 1.1 1.1 0 0 1 1.1 1.1" color="#000" fill="#fff"/>
+    ///   <path d="M80.22 13.93a1.1 1.1 0 0 1-1.1 1.1 1.1 1.1 0 0 1-1.08-1.1 1.1 1.1 0 0 1 1.1-1.08 1.1 1.1 0 0 1 1.08 1.08" color="#000" fill="none" stroke="#030303" stroke-linecap="round" stroke-opacity=".5"/>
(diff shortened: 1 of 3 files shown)


raphlinus:
Ideally this would have more automated testing, but in the meantime I tried a few examples by hand and also integrated it into piet-metal, where it gave significantly better results than the hacky attempt at flattening in that codebase.

It would be possible to do this as an iterator instead of a callback, but the code would be quite a bit more arduous; you'd need to track a lot of state, including an enum for whether you were in a line, quadratic, or cubic.

See https://github.com/nical/lyon/pull/548 and https://github.com/nical/lyon/pull/552 for comparison. There are strong similarities, but I took advantage of a number of opportunities to make the math slightly more efficient, and also did a couple of things considerably more fancy: the quadratics are subdivided fractionally (their endpoints are not included), and there is more careful approximation of cusps.

@nical should feel free to adapt this code or the ideas in it; I borrowed some stuff from his Lyon PR's, but ended up putting my own spin on it, as is my wont.

src/bezpath.rs
```diff
@@ -114,6 +115,27 @@ impl BezPath {
...
+    /// but with some refinements. For one, there is a more careful approximation
+    /// at cusps. For two, the algorithm is extended to work with cubic Béziers
+    /// as well, by first subdividing into quadratics and then computing the
+    /// subdivision of each quadratic. However, as a clever trick, these quadratics
+    /// are subdivided fractionally, and their endpoints are not included.
+    ///
+    /// TODO: write a paper explaining this in more detail.
+    ///
+    /// Note: the [`flatten`](fn.flatten.html) function provides the same
+    /// functionality but works with slices and other [`PathEl`] iterators.
```
cmyr, line 132: missing a link for `PathEl`?

src/bezpath.rs
```diff
@@ -114,6 +115,27 @@ impl BezPath {
         BezPath::segments_of_slice(&self.0)
     }
 
+    /// Flatten the path, invoking the callback repeatedly.
```
cmyr, line 118: This should try and explain what 'flatten' means. Something like,

"Flatten the path, converting it into a `BezPath` that includes only straight lines. The `callback` argument will be called in order with each element of the generated `BezPath`.", or something.

That is: I think this documents the nitty stuff, but not the high-level stuff or how the API should be used.

src/bezpath.rs
```diff
@@ -214,6 +236,110 @@ impl<'a> IntoIterator for &'a BezPath {
...
+/// Flatten the path, invoking the callback repeatedly.
+///
+/// See [`BezPath::flatten`](struct.BezPath.html#method.flatten) for more discussion.
+/// This signature is a bit more general, allowing flattening of `&[PathEl]` slices
+/// and other iterators yielding `PathEl`.
+pub fn flatten(
+    path: impl IntoIterator<Item = PathEl>,
+    tolerance: f64,
+    mut callback: impl FnMut(PathEl),
+) {
```
cmyr, line 260: looking at this, my first thought is that it would be more natural to express this as an iterator, which would remove the need for the callback and simplify the cases where the caller just wants to collect the elements into a vec or a new `BezPath`; and the iterator version should still have all the lazy-evaluation advantages that you might get from the callback?

It's possible I'm misunderstanding the goals, here, but this API is a bit unconventional, and I'm wondering if that's necessary?

raphlinus: I agree that an iterator would be more natural and more convenient for the caller. But having to write this algorithm in continuation-passing style would be a pain. I'll note that Lyon also provides a [for_each_flattened](https://docs.rs/lyon_geom/0.15.2/lyon_geom/cubic_bezier/struct.CubicBezierSegment.html#method.for_each_flattened) method with a similar callback signature, though it also provides the iterator.

nical: As @raphlinus said, lyon has both iterator and callback APIs for path flattening. In practice, while iterators look a bit nicer/rustier, the callback APIs tend to be a faster and easier to implement especially when there is nesting (like cubics into quadratics into line segments). I end up using the callback APIs whenever I can because of the slight performance advantage.
The main reason I keep the iterator version as well is that it can compose with other iterators in places where the callback wouldn't (for example an iterator that flattens an entire path). It's not entirely clear to me whether these few things where I use the flattening iterators are things that I really need, though.

cmyr: okay, thanks @nical.

I do think this API is going to be confusing to someone who isn't expecting it, but maybe that's a vanishingly small number of people? As in: if you're looking for this, you already have a solid idea of what's going on.

PR #110, opened 2020-05-09
Handle exponential notation in svg numbers

raphlinus:
It's valid for numbers in svg to be in exponential notation (eg 1e-3).
Fixing this lets us parse more files in the MPVG input set.


wraybowling:
Whoa. Holy crap. That’s a thing???

PR #111, opened 2020-05-10
impl Shape for Arc

derekdreery:
Defers the implementation to the approximation Bezier curves path.

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -1,7 +1,10 @@
 //! A circle arc.
 
 use crate::{PathEl, Point, Rect, Shape, Vec2};
-use std::f64::consts::{FRAC_PI_2, PI};
+use std::{
+    f64::consts::{FRAC_PI_2, PI},
+    iter,
+};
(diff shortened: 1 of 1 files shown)


src/arc.rs
```diff
@@ -109,15 +109,78 @@ impl Iterator for ArcAppendIter {
...
     Vec2::new(
         pt.x * angle.cos() - pt.y * angle.sin(),
         pt.x * angle.sin() + pt.y * angle.cos(),
     )
 }
+
+impl Shape for Arc {
+    type BezPathIter = ArcIter;
+
+    fn to_bez_path(&self, tolerance: f64) -> Self::BezPathIter {
```
raphlinus, line 131: This looks fine to me but you may be able to express the idea more succinctly with something like `std::iter::once(MoveTo(...)).chain(self.append_iter(tolerance))`

derekdreery:
I've updated to use tools from `std::iter`.

PR #114, opened 2020-05-13
Add circle segments and ellipses.

derekdreery:
I needed these shapes for some graph drawing I'm doing using kurbo. I present them here in case you would like to include them. They will need a little bit of touching up if you want them in the library, I think.


src/ellipse.rs
```diff
@@ -0,0 +1,132 @@
...
+    #[inline]
+    fn sub(self, v: Vec2) -> Ellipse {
+        Ellipse {
+            center: self.center - v,
+            radii: self.radii,
+            x_rotation: self.x_rotation,
+        }
+    }
+}
+
```
cmyr, line 60: I wonder if it also makes sense to do `Mul<Affine>`?

src/ellipse.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    #[inline]
+    fn bounding_box(&self) -> Rect {
+        self.clone().into_bez_path(0.1).elements().bounding_box()
+    }
+}
+
+/*
+#[cfg(test)]
```
cmyr, line 99: any reason not to include this?

src/ellipse.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use std::f64::consts::PI;
+use std::{
+    iter,
+    ops::{Add, Sub},
+};
+
+use crate::{Arc, ArcAppendIter, PathEl, Point, Rect, Shape, Vec2};
+
+/// A ellipse.
```
cmyr, line 11:
```suggestion
/// An ellipse.
```

src/ellipse.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// The center.
+    pub center: Point,
+    /// The radii.
+    pub radii: Vec2,
+    /// How much to rotate the major/minor axes.
+    pub x_rotation: f64,
+}
+
+impl Ellipse {
+    /// A new ellipse from center and radii.
```
cmyr, line 24: and x_rotation?

It would be nice if we could include, in the docs, a link to some resource (wikipedia, mathematica) that explains how this representation works (or if we just do it ourselves)?

src/circle.rs
```diff
@@ -147,6 +150,132 @@ impl Iterator for CirclePathIter {
     }
 }
 
+/// A segment of a circle.
+///
+/// If `inner_radius > 0`, then the shape will be a doughnut segment.
```
cmyr, line 155: What if the inner radius is greater than the outer radius?

src/circle.rs
```diff
@@ -147,6 +150,132 @@ impl Iterator for CirclePathIter {
...
+    /// The outer radius.
+    pub outer_radius: f64,
+    /// The inner radius.
+    pub inner_radius: f64,
+    /// The angle to start drawing the segment (in radians).
+    pub start_angle: f64,
+    /// The arc length of the segment (in radians).
+    pub sweep_angle: f64,
+}
+
```
cmyr, line 168: It would be nice to have a `new` method here for constructing a segment. I could also imagine a convenience method on `Circle` that creates a `CircleSegment` given an inner radius, start angle, and sweep angle?

src/circle.rs
```diff
@@ -1,9 +1,12 @@
 //! Implementation of circle shape.
 
 use std::f64::consts::{FRAC_PI_2, PI};
-use std::ops::{Add, Sub};
+use std::{
```
cmyr, line 4: if we're moving to this notation I would also combine the other `std` import, above.

raphlinus:
Some very rough thoughts, I've only skimmed, not done a careful review.

It definitely makes sense to have this functionality, and it's likely we'll use it.

I agree that Mul<Affine> is important. The specific representation chosen here is not particularly convenient for that operation. Unfortunately there are a lot of representations for elliptical arc segments (the SVG spec suffers from this) and it's not easy to choose a best one. I'm ok with just picking one as long as the external interface is reasonable.

src/ellipse.rs
derekdreery: It's just a cut and past from circle atm. It should either be updated to use ellipses (preferable) ore be deleted before this PR is accepted.

derekdreery:
> The specific representation chosen here

If we are representation-agnostic in the interface, then it's non-breaking to swap out the interface later. :)

src/circle.rs
derekdreery: I don't mind at all which to use - should I use the separate-line way to be more consistent?

src/circle.rs
derekdreery: Good point - I've made it so that the shape is not dependent on the radii being in the correct order.

src/circle.rs
derekdreery: I've added the methods, are they implemented how you would like?

derekdreery:
Ok this PR should be ready for review now. The `Ellipse` has been rewritten to use an `Affine` representation. I've (hopefully) addressed the previous review comments for the path segment.

derekdreery:
I'm keen that someone else who approached the code could work out what's going on from knowledge of geometry, so if there is anything unclear, please tell me so I can add more docs.

src/rect.rs
```diff
@@ -550,7 +550,7 @@ impl fmt::Debug for Rect {
         } else {
             write!(
                 f,
-                "Rect {{ x0 {:?}, y0: {:?}, x1: {:?}, y1: {:?} }}",
```
derekdreery, line 553: This is kind of unrelated to the PR, I just spotted it as I was working.

src/ellipse.rs
```diff
@@ -0,0 +1,261 @@
...
+    fn from(circle: Circle) -> Self {
+        Ellipse::new(circle.center, Vec2::splat(circle.radius), 0.0)
+    }
+}
+
+impl Shape for Ellipse {
+    type BezPathIter = iter::Chain<iter::Once<PathEl>, ArcAppendIter>;
+
+    fn to_bez_path(&self, tolerance: f64) -> Self::BezPathIter {
+        // TODO is there a better way of doing this?
```
derekdreery, line 153: Would it be better to convert a unit circle to a bezier and then apply the affine map?

raphlinus: I'm happy with the code the way it is, mostly because it puts the error tolerance logic in a single place, namely `Arc::append_iter`. Sampling from the unit circle might allow some micro-optimizations, but I don't think it's worth it.

Another subtle point here: currently the arc logic is not as optimized as it might be in producing a minimal number of bezier segments as the ellipse goes more eccentric (it's based on the max of the major/minor axes). Applying affine to circle would lock us into not doing that, and if Arc ever is improved, again it would be best to have a single point.

src/ellipse.rs
```diff
@@ -0,0 +1,261 @@
...
+        // circle.
+        let inv = self.inner.inverse();
+        if (inv * pt).to_vec2().hypot2() < 1.0 {
+            1
+        } else {
+            0
+        }
+    }
+
+    // Compute a tight bounding box of the ellipse.
```
raphlinus, line 196: This warms my heart :)

src/ellipse.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    /// Set the amount (in radians) that the ellipse should be rotated by (clockwise).
+    pub fn with_x_rotation(self, new_x_rotation: f64) -> Ellipse {
+        let scale = self.inner.svd().0;
+        let translation = self.inner.get_translation();
+        Ellipse::private_new(translation, scale.x, scale.y, new_x_rotation)
+    }
+}
+
+// We need to implement this ourselves as multiple affine maps represent the same ellipse. If `==`
```
raphlinus, line 101: It seems to me this will generate `false` in almost all nontrivial cases, because of floating point rounding error. We have been agonizing for a while about "fuzzy equality" but haven't adopted a solution. I'm ok with this as it is but am wondering whether it might be simpler to just use == on the affine.

src/affine.rs
```diff
@@ -114,6 +114,72 @@ impl Affine {
...
+    /// any) circle about its center always results in the same circle. This is the reason that an
+    /// ellipse mapped using an affine map is always an ellipse.
+    ///
+    /// Will return NaNs if the matrix (or equivalently the linear map) is singular.
+    ///
+    /// First part of the return tuple is the scaling, second part is the angle of rotation (in
+    /// radians)
+    #[inline(always)]
+    pub(crate) fn svd(self) -> (Vec2, f64) {
+        // I've put inline(always) here because we want to give the compiler the opportunity to
```
raphlinus, line 142: This should just be `#[inline]`; the compiler applies heuristics for exactly the sort of thing you mention, so `always` should be reserved for cases where there is empirical evidence the compiler is missing an optimization and it's impacting performance.

src/affine.rs
```diff
@@ -114,6 +114,72 @@ impl Affine {
...
+    /// affine map on the unit circle, we don't calculate V^T, since a rotation of the unit (or
+    /// any) circle about its center always results in the same circle. This is the reason that an
+    /// ellipse mapped using an affine map is always an ellipse.
+    ///
+    /// Will return NaNs if the matrix (or equivalently the linear map) is singular.
+    ///
+    /// First part of the return tuple is the scaling, second part is the angle of rotation (in
+    /// radians)
+    #[inline(always)]
+    pub(crate) fn svd(self) -> (Vec2, f64) {
```
cmyr, line 141: If this were public API I would suggest using a struct with named fields instead of a tuple as the result type, but this is fine for internal code. 👌

src/circle.rs
```diff
@@ -24,6 +27,17 @@ impl Circle {
             radius,
         }
     }
+
+    /// Create a circle segment by cutting out parts of this circle.
```
cmyr, line 31: Might as well link to `CircleSegment`:

```suggestion
    /// Create a [`CircleSegment`] by cutting out parts of this circle.
    ///
    /// [`CircleSegment`]: struct.CircleSegment.html
```

src/ellipse.rs
```diff
@@ -0,0 +1,261 @@
...
+    #[inline]
+    pub fn new(center: impl Into<Point>, radii: impl Into<Vec2>, x_rotation: f64) -> Ellipse {
+        let Point { x: cx, y: cy } = center.into();
+        let Vec2 { x: rx, y: ry } = radii.into();
+        Ellipse::private_new(Vec2 { x: cx, y: cy }, rx, ry, x_rotation)
+    }
+
+    /// This gives us an internal method without any type conversions.
+    #[inline]
+    fn private_new(center: Vec2, scale_x: f64, scale_y: f64, x_rotation: f64) -> Ellipse {
```
cmyr, line 37: 👍

src/affine.rs
derekdreery: Keeping it private for now is good because it does not prevent any public API to be added later.

raphlinus: Agree it should be private, but Colin's separate point is that a struct with named fields would be more readable than a tuple.

cmyr: yea sorry if my point was confusing; I think it's good that this is private, but if it *weren't* I would prefer a struct with named fields to a tuple. I think this is fine as is!

src/ellipse.rs
cmyr: @derekdreery I think this comment and the one below need resolution and then we're good to go. :)

derekdreery: You're right, and even then there are multiple representations of the ellipse (e.g. changing the sign of the radii is the identity). I will just use the derived impl.

src/affine.rs
derekdreery: I actually tried creating a `Svd2x2` struct for the result, but I got lost down a rabbit hole.

derekdreery:
Thanks for letting me leave the segments in ;). They do have a few uses, for example pie charts.

cmyr:
Great, thanks! will get this into the next druid release.

PR #115, opened 2020-05-14
Add `fix_aspect_ratio` fn

derekdreery:
This is a function that I would find useful. It gives you a `Rect` to draw on that has a specific aspect ratio, is centred in the containing `Rect`, and is the maximum size it can be. Name is bikesheddable if you like the function.

### DIFF
--- a/src/rect.rs
+++ b/src/rect.rs
@@ -420,6 +420,50 @@ impl Rect {
     pub fn to_rounded_rect(self, radius: f64) -> RoundedRect {
         RoundedRect::from_rect(self, radius)
     }
+
+    /// Returns the largest possible `Rect` that is fully contained in `self` with the given
+    /// `aspect_ratio`.
+    ///
+    /// The aspect ratio is specified fractionally, as `height / width`.
+    ///
+    /// The resulting rectangle will be centered if it is smaller than the input rectangle.
+    ///
+    /// For the special case where the aspect ratio is `1.0`, the resulting `Rect` will be square.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use kurbo::Rect;
+    /// let outer = Rect::new(0.0, 0.0, 10.0, 20.0);
+    /// let inner = outer.contained_rect_with_aspect_ratio(1.0);
+    /// // The new `Rect` is a square centered at the center of `outer`.
+    /// assert_eq!(inner, Rect::new(0.0, 5.0, 10.0, 15.0));
+    /// ```
+    ///
+    pub fn contained_rect_with_aspect_ratio(&self, aspect_ratio: f64) -> Rect {
+        let (width, height) = (self.width(), self.height());
+        let self_aspect = height / width;
+
+        // TODO the parameter `1e-9` was chosen quickly and may not be optimal.
+        if (self_aspect - aspect_ratio).abs() < 1e-9 {
+            // short circuit
+            *self
+        } else if self_aspect.abs() < aspect_ratio.abs() {
+            // shrink x to fit
+            let new_width = height * aspect_ratio.recip();
+            let gap = (width - new_width) * 0.5;
+            let x0 = self.x0 + gap;
+            let x1 = self.x1 - gap;
+            Rect::new(x0, self.y0, x1, self.y1)
+        } else {
+            // shrink y to fit
+            let new_height = width * aspect_ratio;
+            let gap = (height - new_height) * 0.5;
+            let y0 = self.y0 + gap;
+            let y1 = self.y1 - gap;
+            Rect::new(self.x0, y0, self.x1, y1)
+        }
+    }
 }
 
 impl From<(Point, Point)> for Rect {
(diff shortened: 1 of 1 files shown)


derekdreery:
It's quite hard to explain in text what this function does, but it's very easy in pictures. I made it because to draw a pie chart I need, given any containing rect, a square that looks sensible and fits inside the rect.

derekdreery:
Ok so I think this PR is better now. It doesn't panic and it has a better name. I've added tests to demonstrate how it performs with edge cases.

src/rect.rs
```diff
@@ -615,4 +656,38 @@ mod tests {
         );
     }
     */
+
+    #[test]
+    fn fix_aspect_ratio() {
+        use std::f64;
```
xStrom, line 662: The Travis CI here uses a rather old rust compiler version. I think a better approach would be to improve the CI situation. I'll see if I can cook something up.

derekdreery: I'm happy to do this as an alternative, what are others' thoughts?

cmyr: @raphlinus was it an intentional decision to support older rust compilers? I seem to recall some conversation to that effect but I might be confused.

raphlinus: I think not so much an intentional decision to support older compilers, but so we know when to explicitly raise the MSRV. I'm fine with requiring latest stable actually.

At some point in the future (when people want to distribute druid apps on Debian, for example), we might want to have a more conservative MSRV policy.

src/rect.rs
```diff
@@ -420,6 +420,47 @@ impl Rect {
...
+    /// let inner = outer.fix_aspect_ratio(1.0);
+    /// // The new `Rect` is a square centered at the center of `outer`.
+    /// assert_eq!(inner, Rect::new(0.0, 5.0, 10.0, 15.0));
+    /// ```
+    ///
+    pub fn fix_aspect_ratio(&self, aspect_ratio: f64) -> Rect {
+        let (width, height) = (self.width(), self.height());
+        let self_aspect = height / width;
+
+        if self_aspect == aspect_ratio {
```
cmyr, line 445: This seems like a place where floating point error could come up, although the downside is negligible.

src/rect.rs
```diff
@@ -420,6 +420,47 @@ impl Rect {
     pub fn to_rounded_rect(self, radius: f64) -> RoundedRect {
         RoundedRect::from_rect(self, radius)
     }
+
+    /// Returns the largest possible `Rect` that is fully contained in this rect with the fixed
+    /// aspect ratio given by `aspect_ratio`.
```
cmyr, line 425:
```suggestion
```suggestion
    /// Returns the largest possible `Rect` that is fully contained in `self` with the given `aspect_ratio`.
    ///
    /// The aspect ratio is specified fractionally, as `height / width`.
```
    /// aspect ratio given by `aspect_ratio`.
```

File: src/rect.rs [hunk 4]
```diff
@@ -420,6 +420,47 @@ impl Rect {
...
+    ///
+    /// ```
+    /// # use kurbo::Rect;
+    /// let outer = Rect::new(0.0, 0.0, 10.0, 20.0);
+    /// let inner = outer.fix_aspect_ratio(1.0);
+    /// // The new `Rect` is a square centered at the center of `outer`.
+    /// assert_eq!(inner, Rect::new(0.0, 5.0, 10.0, 15.0));
+    /// ```
+    ///
+    pub fn fix_aspect_ratio(&self, aspect_ratio: f64) -> Rect {
```
cmyr, line 441: The only real Q for me about this PR is about naming.

`constrain_aspect_ratio`? `contained_rect_with_aspect_ratio`? `with_fixed_aspect_ratio`?

derekdreery: i think probably a longer name that fully describes the function would be better - so `contained_rect_with_aspect_ratio`, what do others think?

cmyr: I agree that clarity is the thing to optimize for here, that name is meh but also it's fine, happy to get this merged with that change (and optionally something along the lines of the doc suggestion above) and then we can get it out and into the next piet & druid releases.

src/rect.rs
derekdreery: Is there a `fuzzy_eq` or similar?

src/rect.rs
derekdreery: It can always be changed (or removed) before 1.0.

src/rect.rs
derekdreery: I've done `(x - y).abs() < 1e-6` for now - is there a standard negligible error you use?

cmyr: no, you'd have to just pick a threshold and check against that. We do really want to have some sort of comprehensive fuzzy_eq implementation but we haven't done it yet.

derekdreery:
I've changed the name, and also rebased + squashed.

cmyr:
hmm I screwed something up with that suggestion 🤔

derekdreery:
Right I think it's good now. :)

derekdreery:
Ohhh ok clippy doesn't want == with float - I will use 1e-9 and make a note that the parameter hasn't been tuned.

src/rect.rs
xStrom: I've found [float-cmp](https://crates.io/crates/float-cmp) to be pretty good, but admittedly my floating point understanding is superficial.

cmyr:
sure, you can also just do a `#[allow(_)]` on the method.

derekdreery:
I put a pathological example in and apparently the output has changed in master. I've just removed it as it was a 0/infinity situation, where any answer is meaningless. The fact that it is now `NaN` is probably better than before.

cmyr:
Can you clarify what you mean about the output having changed?

cmyr:
okay let's get this in, would like to get a new kurbo out for druid 0.6. Thanks again @derekdreery!

derekdreery:
In the case where the width was 0 and the aspect ratio was infinity, some
of the results became NaN rather than 0. It was a pathological case though
so any results would be meaningless

PR #120, opened 2020-05-28
Path segment intersection against line

raphlinus:
Adds a method to test intersection of any path segment against a line,
and report the parameter on both the segment and the line for each
intersection found.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -735,8 +735,9 @@ impl PathSeg {
                 // t = position on self
                 let t = t / det;
                 if t >= -EPSILON && t <= 1.0 + EPSILON {
-                // u = position on probe line
-                    let u = (l.p0.x - p0.x) * (l.p1.y - l.p0.y) - (l.p0.y - p0.y) * (l.p1.x - l.p0.x);
+                    // u = position on probe line
+                    let u =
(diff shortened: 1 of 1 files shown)


raphlinus:
If you're interested in following along with the math, I basically followed the blog [Computing Intersections Between a Cubic Bezier Curve and a Line](https://www.particleincell.com/2013/cubic-line-intersection/), but did a little of my own spin. It also helps a *lot* that we already have robust quadratic and cubic solvers, these methods would be really flaky in the presence of degenerate curves otherwise.

src/bezpath.rs
```diff
@@ -706,6 +706,106 @@ impl PathSeg {
...
+    ///
+    /// Returns a vector of the intersections. For each intersection, the `t` value of the
+    /// segment and line are given.
+    ///
+    /// Note: This test is designed to be inclusive of points near the endpoints of
+    /// the segment. This is so that testing a line against multiple contiguous segments
+    /// of a path will be guaranteed to catch at least one of them. In such cases, use
+    /// higher level logic to coalesce the hits (the `t` value may be slightly outside
+    /// the range of 0..1).
+    pub fn intersect_line(&self, line: Line) -> ArrayVec<[(f64, f64); 3]> {
```
cmyr, line 720: I think it's worth having a `LineIntersection` struct here, but I'm happy to do that as follow-up.

PR #121, opened 2020-05-29
Add LineIntersection type and tweak some docs

cmyr:
This also bumps our version to 0.6.2


src/bezpath.rs
```diff
@@ -48,6 +48,28 @@ pub enum PathSeg {
...
+/// This can be generated with the [`PathSeg::intersect_line`] method.
+///
+/// [`Line`]: struct.Line.html
+/// [`PathSeg`]: enum.PathSeg.html
+/// [`PathSeg::intersect_line`]: enum.PathSeg.html#method.intersect_line
+#[derive(Debug, Clone, Copy)]
+pub struct LineIntersection {
+    /// The 'time' that the intersection occurs, on the line.
+    ///
+    /// This value is nominally in the range 0..1, although it may slightly excede
```
raphlinus, line 62: In the current implementation, it's actually hard bounded by 0..1 inclusive. But it might not be a bad idea for the documentation to be more relaxed. Also, "exceed".

PR #123, opened 2020-06-04
Add aspect ratio function to some types.

derekdreery:
These are some more helper methods for the basic types. They would be useful in my case for trying to match a given aspect ratio when constraining during layout.

### DIFF
--- a/src/rect.rs
+++ b/src/rect.rs
@@ -429,6 +429,16 @@ impl Rect {
         Ellipse::from_rect(self)
     }
 
+    /// The aspect ratio of the `Rect`.
+    ///
+    /// This is defined as the height divided by the width. It measures the "squareness" of the
+    /// rectangle (a value of `1` is square). If the width is `0` the output will be `sign(y1 - y0)
+    /// * infinity`. If The width and height are `0`, the result will be `NaN`.
+    #[inline]
+    pub fn aspect_ratio(&self) -> f64 {
+        self.size().aspect_ratio()
+    }
+
     /// Returns the largest possible `Rect` that is fully contained in `self` with the given
     /// `aspect_ratio`.
     ///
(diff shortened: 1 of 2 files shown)


src/rect.rs
```diff
@@ -421,6 +421,15 @@ impl Rect {
         RoundedRect::from_rect(self, radius)
     }
 
+    /// The aspect ratio of the `Rect`.
+    ///
+    /// This is defined as the height divided by the width. It measures the "squareness" of the
+    /// rectangle (a value of 1 is square).
+    #[inline]
+    pub fn aspect_ratio(&self) -> f64 {
+        (self.y1 - self.y0) / (self.x1 - self.x0)
```
cmyr, line 430: an observation: I find with methods like this if I can define them in one place and then call out to that I prefer that approach, and so I'd have probably just done `self.size().aspect_ratio()`.

src/size.rs
```diff
@@ -178,6 +178,11 @@ impl Size {
         Size::new(self.width.trunc(), self.height.trunc())
     }
 
+    /// Returns the aspect ratio of a rectangle with the given size.
+    pub fn aspect_ratio(self) -> f64 {
```
cmyr, line 182: probably also worth an `#[inline]`

src/size.rs
```diff
@@ -178,6 +178,11 @@ impl Size {
...
+        self.height / self.width
```
cmyr, line 183: I wonder if it's worth handling the zero cases specially.

What are the aspect ratios of `(0, 10)`, `(0, 0)` and `(5, 0)`? If we're returning `NaN` that should probably be documented.

derekdreery: I think NaN is probably the correct result of (0, 0), since 0 / 0 really doesn't mean anything. For the others they should just be 0 and infinity.

cmyr: I could imagine (0, 0) being 1? I know this is going to offend everyone, and I'm not going to fight for it. `NaN` is fine. :)

derekdreery: to be honest it could be anything, so if 1 causes fewer problems then fine. 0 = 1 x 0 after all. :)

cmyr: I think @raphlinus will get mad if `0, 0` is 1. 😢

derekdreery:
OK should be ready for review.

PR #124, opened 2020-06-04
CubicBez::to_quads always returns at least one quad

cmyr:
This fixes a crash in CubicBez::nearest.

### DIFF
--- a/src/cubicbez.rs
+++ b/src/cubicbez.rs
@@ -50,8 +50,7 @@ impl CubicBez {
     /// Note that the resulting quadratic Béziers are not in general G1 continuous;
     /// they are optimized for minimizing distance error.
     ///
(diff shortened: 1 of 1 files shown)


src/cubicbez.rs
```diff
@@ -408,6 +407,14 @@ mod tests {
         verify((a * c).nearest(a * Point::new(0.1, 0.001), 1e-6), 0.1);
     }
 
+    // ensure to_quads returns something given colinear points
+    #[test]
+    fn degenerate_to_quads() {
+        let c = CubicBez::new((0., 9.),(6., 6.),(12., 3.0 ), (18., 0.0));
```
raphlinus, line 413: cargo fmt wants spaces here

cmyr: on clue why CI is failing, seems random

PR #134, opened 2020-09-28
Add Rect::intersects

laurmaedje:
This adds a method to check whether two rectangles overlap. 

Small note: In `Rect::contains` (for points) the left and top edges are inclusive, while the right and bottom edges are exclusive. Here, however, I think all comparison should be less/greater-than since even if the edges coincide (and the rectangles just touch each other), the intersected area is still zero and the method should return `false`.

### DIFF
--- a/src/rect.rs
+++ b/src/rect.rs
@@ -218,6 +218,12 @@ impl Rect {
         Rect::new(x0, y0, x1.max(x0), y1.max(y0))
     }
 
+    /// Returns `true` if `self` and `other` intersect.
+    #[inline]
+    pub fn intersects(&self, other: Rect) -> bool {
+        self.x0 < other.x1 && self.x1 > other.x0 && self.y0 < other.y1 && self.y1 > other.y0
+    }
+
     /// Expand a rectangle by a constant amount in both directions.
     ///
     /// The logic simply applies the amount in each direction. If rectangle
@@ -656,6 +662,12 @@ mod tests {
         let p_flip = r_flip.into_bez_path(1e-9);
         assert_approx_eq(r_flip.area(), p_flip.area());
         assert_eq!(r_flip.winding(center), p_flip.winding(center));
+    }
+
+    #[test]
+    fn intersects() {
+        assert!(Rect::new(0.0, 0.0, 10.0, 10.0).intersects(Rect::new(-5.0, -1.0, 1.0, 2.0)));
+        assert!(!Rect::new(5.0, 5.0, 10.0, 10.0).intersects(Rect::new(12.0, 3.0, 13.0, 4.0)));
     }
 
     #[test]


derekdreery:
What is the use-case you have in mind for this function? Or are you adding it for completeness?

laurmaedje:
I implemented a recursive bounding-box-based curve-curve intersection algorithm and need to stop when the curve's bounding boxes don't overlap. Something like `a.intersect(b)` and then somehow checking whether that rect it is empty would probably work, but I think this way it would be cleaner. It's also probably faster (unless the compiler is smart enough). Currently, I have a `RectExt` extension trait for this, but I thought it might be useful to others.

cmyr:
This to me mostly highlights the surprise that `Rect` doesn't implement `is_empty`. I guess the easiest way to check right now would be something like `my_rect.area() == 0.0`, which doesn't feel great.

So... I'm not sure? I *think* my preferred solution here would be a better way to do `x.intersect(y).is_empty()`, but I don't have super strong opinions, but I'd like us to have a generally consistent API (across the various shapes) so that users can intuit what will be available.

But also this is harmless and useful.

I'd be curious to hear anyone else's thoughts.

derekdreery:
My only concern is whether the edges should count. In a strictly mathematical sense, if the rects are closed sets, then they would, but I think in practice what really matters is the area of the intersection, and if this is 0.0 then they can be considered to not have intersected.

derekdreery:
Either way I'd recommend putting a note on the doc comment for the method explaining the choice.

cmyr:
Perhaps that's an argument for going in the direction of `is_empty`, which can be defined strictly in terms of area.

laurmaedje:
I agree that going with `is_empty` would be quite elegant, but (at least if I have done this correctly - never used GodBolt before really) it also seems that notably more code is emitted by the compiler, which will probably make things slower: https://godbolt.org/z/YPMcss

Personally, when I needed this, I was surprised that it didn't exist yet. I would have thought that bounding-box intersection testing is a rather common thing to do (maybe more so in physics-related code).

laurmaedje:
Though, no matter how the intersection problem is dealt with, I think having `is_empty` would definitely be useful on its own.

derekdreery:
I thin `#[no_mangle]` prevents inlining. Here's my version: https://godbolt.org/z/jz5b48. I don't understand the LLVM IR. The asm looks similar: the direct way has more branching, because it short-circuits on empty. They're both fast I think.

cmyr:
I think the main reason that `intersection(y).is_empty()` is producing more code is because it is robust to the fact that a rect may have negative size; that is, it is not an invariant that x0 <= x1. I think we would want a possible `intersects` method to be similarly robust.

laurmaedje:
Hmm, that's a good point. I'd also be fine with just going the `is_empty` route. Then, we could maybe add a doc-comment to `intersect` telling that the canonical way to just check for intersection is `a.intersect(b).is_empty()`.

cmyr:
That sounds good to me, I would be happy for a PR that adds `to_empty` to `Rect`, and maybe also to `Size`? In both cases this should be defined as area == 0.0.

raphlinus:
Sounds reasonable to me. I'm pretty the subtle differences between different interpretations of what "empty" means are a reason I didn't implement it already, but if there's discussion and consensus, I'm happy.

Also, yay Godbolt, and yay Rust + LLVM for generating impressively good code for the `is_empty()` version of this logic.

laurmaedje:
Since we decided on `a.intersect(b).is_empty()` I'm going to close this.

PR #137, opened 2020-09-28
Add Rect::is_empty and Size::{area, is_empty}

laurmaedje:
This adds `is_empty` to `Rect` and `Size` as discussed in #134. I also added `Size::area` to implement `is_empty` in terms of it.

### DIFF
--- a/src/rect.rs
+++ b/src/rect.rs
@@ -154,6 +154,14 @@ impl Rect {
         self.width() * self.height()
     }
 
+    /// Whether this rectangle has zero area.
+    ///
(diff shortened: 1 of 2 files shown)


laurmaedje:
Ah, I forgot, should I add a comment to `intersect` about using `is_empty` to this PR?

derekdreery:
> Ah, I forgot, should I add a comment to intersect about using is_empty to this PR?

I think it's fine either way, so up to you. :)

cmyr (approved):
Thanks!

Issue #136, opened 2020-09-28
Converting iterator of path elements to segments


laurmaedje:
There is currently no built-in way to transform an iterator of path elements to an iterator of path segments. My use case is to take an argument of type `impl Shape` and to iterate over its segments. `Shape::to_bez_path` only returns an iterator over path elements. I could, of course, collect into a `BezPath` and the call `segments()` on that, but I'd like to avoid the intermediate allocation.

Looking into the source code, there is the private function `BezPath::segments_of_slice`, which has a todo that it should maybe be public, and there is the underlying `BezPathSegs` iterator. Unfortunately, `segments_of_slice` still wouldn't work with general iterators. 

To fix this, how about adding a freestanding function like this:
```rust
pub fn segments(path: impl IntoIterator<Item = PathEl>) -> impl Iterator<Item = PathSeg> { ... }
```
This function would be nicely parallel to `flatten` and could similarly link to `BezPath::segments`, which would then be implemented in terms of it. `segments_of_slice` could then be removed, I think.

If you think this sounds like a good way to do it, I'd be happy to open a PR!

(An alternative, also stated in `segments_of_slice`'s comment, would be to create a trait such that `iter.segments()` works on any iterator. Personally, I think this wouldn't be worth it and the parallelity with `flatten` would also be lost.)

cmyr:
I think this makes sense. In terms of the implementation, couldn't we add a `segments` method to `Shape`, and have a default impl that just wraps `to_bez_path`? This could be backed by a public helper method in case the user wanted to iterate the segments of a sequence of `PathEl` that wasn't a `Shape`.

Looking at the `Shape` trait again, the api nerd in me really thinks that `to_bez_path` should be renamed `iter_bezier_elements` or `iter_elements` or `bezier_elements`, and then a separate `to_bezier` method should return a constructed `BezPath`. 🤷

laurmaedje:
Adding it to `Shape` sounds good to me. I agree that the naming of `to_bez_path` is confusing, I tripped over this multiple times. If we would rename `to_bez_path`, I think both methods should be named similarly, like `elements` and `segments` or `iter_elements` and `iter_segments`.

cmyr:
yes, I agree. I also think this can be done in a pretty painless way, by marking `to_bez_path` and `into_bez_path` as deprecated, adding `elements` and `segments`, and then adding a new `to_bezpath` or `to_bezier` method that returns a `BezPath`. It's the name of this latter method that I'm unsure about.

[...]

PR #139, opened 2020-09-29
Restructured path handling in Shape and segments iterator

laurmaedje:
This implements the plans laid out in #136.

`BezPathSegs` got a bit refactored and renamed to `Segments` created by the public `segments` function.

`Shape` gets the following changes:
- `BezPathIter` -> `PathElementsIter`
- `to_bez_path` -> `to_path_elements`
- `into_bez_path` -> `into_path` and `BezPath` actually uses this for zero-cost conversion
- adds new `to_path` that takes `&self`

Apart from updating the primitive shapes to the new `Shape` interface, I was able to make some functions for `Arc` and `Ellipse` zero-allocation by using methods directly in `Segments` instead of creating an intermediate `BezPath`. I marked the methods on `Segments` as `pub(crate)` for now, but maybe (as a previous TODO also already indicated) it would be a good idea to make them fully public.

Note: This contains breaking changes and needs kurbo `0.7`.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -1,13 +1,13 @@
 //! A generic trait for shapes.
 
-use crate::{BezPath, Circle, Line, PathEl, Point, Rect, RoundedRect};
+use crate::{segments, BezPath, Circle, Line, PathEl, Point, Rect, RoundedRect, Segments};
 
 /// A generic trait for open and closed shapes.
 pub trait Shape: Sized {
-    /// The iterator resulting from `to_bez_path`.
-    type BezPathIter: Iterator<Item = PathEl>;
+    /// The iterator resulting from `to_path_elements`.
+    type PathElementsIter: Iterator<Item = PathEl>;
 
-    /// Convert to a Bézier path, as an iterator over path elements.
+    /// Convert to an iterator over Bézier path _elements_.
     ///
     /// Callers should exhaust the `as_` methods first, as those are
     /// likely to be more efficient; in the general case, this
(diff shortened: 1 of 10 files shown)


src/shape.rs
```diff
@@ -27,22 +27,36 @@ pub trait Shape: Sized {
     /// iterators from complex shapes without cloning.
     ///
     /// [GAT's]: https://github.com/rust-lang/rust/issues/44265
-    fn to_bez_path(&self, tolerance: f64) -> Self::BezPathIter;
+    fn to_path_elements(&self, tolerance: f64) -> Self::PathElementsIter;
```
cmyr, line 30: I might have missed this in discussion, but I would not include the `to_` prefix for methods that return iterators.

cmyr (commented):
This looks great, my only note is that I would use `path_segments` and `path_elements` instead of `to_path_segments` and `to_path_elements`.

I think a `DEFAULT_TOLERANCE` `const` would be nice, but we can do that as follow-up.

src/shape.rs
laurmaedje: Changed. I've also reworded the documentation a little because it said "Convert to an iterator ..." previously, which is not really accurate.

I changed it to the follwing:
> Returns an iterator over this shape expressed as Bézier path elements.

I'm not _super_ happy with the wording, but it's the best I've got. I'm open to suggestions.

cmyr: Maybe something like,

```
    /// Returns an iterator over the individual ([`PathEl`])s of a [`BezPath`] that
    /// represents this shape.
    ///
    /// This method should not allocate, and is suitable for things like drawing,
    /// as the returned elements map closely to many 2D graphics APIs.
    ///
    /// If you are are more interested in the *segments* of the path (that is, the
    /// constituent lines and curves) then you should use [`Shape::path_segements`]
    /// instead. If you would like to retain the [`BezPath`], use [`Shape::to_path`].
```

In any case I'm happy to touch up the docs in a subsequent PR, I think this is in a good place to merge.

cmyr (approved):
This looks good to me, will leave up for a little bit in case @raphlinus or @jneem would like to take a pass as well.

src/shape.rs
laurmaedje: I like that, though I'm not sure about the should-not-allocate part since it _does_ allocate for `<BezPath as Shape>::path_elements` (at least until GATs land).

> In any case I'm happy to touch up the docs in a subsequent PR, I think this is in a good place to merge.

Sounds good!

raphlinus (approved):
Looks good to me also, thanks for your work on this.

PR #145, opened 2020-10-13
Give BezPath::nearest a concrete return type

cmyr:
In the course of this work I had a few questions about whether or
not this method is pulling its weight; I have not found it useful
as written, and where I have need this sort of functionality I
have ended up just iterating the segments and working with them
directly.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -262,18 +262,40 @@ impl BezPath {
     /// followed by `Lineto/Quadto/Cubicto` elements, the element index is the
     /// segment index + 1.
     ///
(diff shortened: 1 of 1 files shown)


raphlinus:
Right. I've also been iterating over segments for very similar reasons - basically, to use this method in Runebender, we'd need to map the segment index back to an EntityId. The way it works now is arguably better; it iterates over Runebender PathSeg, not kurbo BezPath segments, and the former object has direct access to EntityId.

Since this method is providing little value over iterating segments, I'd be comfortable removing it altogether.

cmyr:
I'm happy to remove this.

PR #146, opened 2020-10-13
Improve docs in lib.rs and shape.rs

cmyr:
@raphlinus a few questions in here for you.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,18 +12,74 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! A garden of data structures for manipulating 2D shapes and curves.
+//! 2D geometry, with a focus on curves.
 //!
 //! The kurbo library contains data structures and algorithms for curves and
-//! vector paths. It is probably most appropriate for creative tools,
-//! but is general enough it might be useful for other applications.
+//! vector paths. It was designed to serve the needs of 2D graphics applications,
+//! but it is intended to be general enough to be useful for other applications.
+//!
+//! Kurbo is designed to be used by [`Piet`], a crate for drawing 2D graphics,
+//! and is in turn used by [`Druid`], a cross-platform GUI toolkit.
+//!
+//! # Examples
+//!
+//! Basic UI-style geometry:
+//! ```
+//! use kurbo::{Insets, Point, Rect, Size, Vec2};
+//!
+//! let pt = Point::new(10.0, 10.0);
+//! let vector = Vec2::new(5.0, -5.0);
+//! let pt2 = pt + vector;
+//! assert_eq!(pt2, Point::new(15.0, 5.0));
+//!
+//! let rect = Rect::from_points(pt, pt2);
+//! assert_eq!(rect, Rect::from_origin_size((10.0, 5.0), (5.0, 5.0)));
+//!
+//! let insets = Insets::uniform(1.0);
+//! let inset_rect = rect - insets;
+//! assert_eq!(inset_rect.size(), Size::new(3.0, 3.0));
+//! ```
+//!
+//! Finding the closest position on a [`Shape`]'s perimeter to a [`Point`]:
+//!
+//! ```
+//! use kurbo::{Circle, ParamCurve, ParamCurveNearest, Point, Shape};
+//!
+//! const DESIRED_ACCURACY: f64 = 0.1;
+//!
+//! /// Given a shape and a point, returns the closest position on the shape's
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -12,11 +12,64 @@
...
+//! ```
+//!
+//! Finding the position on a [`Shape`]'s perimeter to a [`Point`]:
+//!
+//! ```
+//! use kurbo::{Point, Rect, Circle, Shape, ParamCurve, ParamCurveNearest};
+//!
+//! /// Given a shape and a point, returns the closest position on the shape's
+//! /// parimeter, or `None` if the shape is malformed.
+//! fn closest_parimeter_point(shape: impl Shape, pt: Point) -> Option<Point> {
```
raphlinus, line 50: perimeter

src/lib.rs
```diff
@@ -12,11 +12,64 @@
...
+//!         }
+//!     }
+//!     best.map(|(point, _)| point)
+//! }
+//!
+//! let circle = Circle::new((5.0, 5.0), 5.0);
+//! let hit_point = Point::new(5.0, -2.0);
+//! let expectation = Point::new(5.0, 0.0);
+//! let hit = closest_parimeter_point(circle, hit_point).unwrap();
+//! assert!(hit.distance(expectation) <= f64::EPSILON.sqrt());
```
raphlinus, line 66: I think this should be based on DESIRED_ACCURACY, not EPSILON.

src/shape.rs
```diff
@@ -3,16 +3,35 @@
...
-    /// Returns an iterator over this shape expressed as Bézier path
-    /// _elements_.
+    /// Returns an iterator over this shape expressed as [`PathEl`]s;
+    /// that is, as Bézier path _elements_.
+    ///
+    /// All shapes can be represented as Béziers, but in many situations
+    /// (such as when interfacing with a platform drawing API) there are more
+    /// efficient native types for specific concrete shapes. In this case,
+    /// the user should exhaust the `as_` methods ([`as_rect`], [`as_line`], etc)
+    /// before converting to a bezier, as those are likely to be more efficient.
```
raphlinus, line 28: Do we want to be consistent about the spelling (title case and with accent)? I do use "bezier" myself a lot, especially to refer to a concrete representation.

src/shape.rs
```diff
@@ -48,15 +75,18 @@ pub trait Shape: Sized {
     ///
     /// The `tolerance` parameter is the same as for
     /// [`path_elements()`](#tymethod.path_elements).
+    //TODO: Does this pull its weight? It seems marginal at best.
```
raphlinus, line 78: Yes, it avoids allocating in the `BezPath` implementation (there, it's free). This was discussed some in #136, see in particular the tables. I think it's worth capturing at least some of that in docs.

src/lib.rs
cmyr: ha I agonized over this, that is a very reasonable solution 😅

src/shape.rs
cmyr: Does it duplicate the functionality provided by `Shape::as_path_slice`?

Alternatively, we could also have a `as_bezpath` method, similar to `as_rect` or `as_circle`?

cmyr: .. which I guess is what `as_path_slice` is doing, really..

raphlinus: It doesn't duplicate that functionality, as that would require allocation to create a BezPath, and this doesn't.

The choice between `as_path_slice` and `as_bezpath` is subtle, and I think boils down to the same reason we generally prefer `&str` in APIs rather than `&String`, though the analogy is not exact; there are some potentially useful methods on `BezPath` that are not implemented on a path slice.

cmyr:
okay, fixed the typos and removed that comment.

src/shape.rs
```diff
@@ -84,6 +115,7 @@ pub trait Shape: Sized {
     /// magnitude values are also possible when the shape is more complex.
     ///
     /// [`area`]: #tymethod.area
+    /// [winding number]: https://mathworld.wolfram.com/ContourWindingNumber.html
```
raphlinus, line 118: Any particular reason why the mathworld link here? Wikipedia also seems good. I'm ok with either, though.

cmyr: No particular reason, we've used mathworld in other places and the explanations are concise.

raphlinus (approved):
This looks good to me now. Thanks!

PR #148, opened 2020-10-14
Add deprecations for old Shape methods

cmyr:
This isn't strictly necessary but will make updating easier and
is otherwise harmless.


raphlinus (approved):
It's still a breaking change, but less breaking, yes?

cmyr:
Yea, we have other breaking changes anyway; this will just make it easier for people to move to the new API.

PR #150, opened 2020-10-20
Impl Shape for QuadBez, CubicBez, & PathSeg

cmyr:


src/bezpath.rs
```diff
@@ -1040,6 +1040,68 @@ impl<'a> Shape for &'a [PathEl] {
...
+
+impl Iterator for PathSegIter {
+    type Item = PathEl;
+
+    fn next(&mut self) -> Option<PathEl> {
+        self.ix += 1;
+        match (self.ix, self.seg) {
+            // yes I could do some fancy bindings thing here but... :shrug:
+            (1, PathSeg::Line(line)) => Some(PathEl::MoveTo(line.p0)),
+            (1, PathSeg::Quad(line)) => Some(PathEl::MoveTo(line.p0)),
```
raphlinus, line 1095: Don't use "line". I think "seg" would be ok if we want consistency.

src/quadbez.rs
```diff
@@ -92,6 +92,55 @@ impl QuadBez {
...
+    type PathElementsIter = QuadBezIter;
+
+    #[inline]
+    fn path_elements(&self, _tolerance: f64) -> QuadBezIter {
+        QuadBezIter { quad: *self, ix: 0 }
+    }
+
+    /// The area under the curve.
+    ///
+    /// We could just return 0, but this seems more useful.
```
raphlinus, line 111: I think I'd prefer 0, to be consistent with Line. The idea of these individual signed_area methods is that you can add them up for all the segments of a closed path.

I'm wondering now whether it might be better do implement it for all such segments, that way you get the "shoelace algorithm" for polygon area. In any case, I'd prefer that Line, QuadBez, and CubicBez are consistent. I'm fine with PathSeg having the same behavior as a single element BezPath though.

raphlinus (approved):
Generally looks good, a few concerns inline.

PR #151, opened 2020-10-27
Minor docstring cleanup

cmyr:
These were showing up with weird line breaking in vim + rust-analyzer.


raphlinus (approved):
Not sure why the old version wasn't working, but seems harmless.

PR #152, opened 2020-11-04
Fix ParamCurveArclen::inv_arclen iteration count

chubei-oppen:


cmyr:
Playing around, I agree that on larger curves the current implementation does strange things as the accuracy creeps towards 1.0, and we crash after falling through the loop at 1.0. I will defer to @raphlinus over whether this is our best solution. :)

src/param_curve.rs
```diff
@@ -92,7 +92,7 @@ pub trait ParamCurveArclen: ParamCurve {
         let mut t_last = 0.0;
         let mut t0 = 0.0;
         let mut t1 = 1.0;
-        let n = (-accuracy.log2()).ceil();
+        let n = (self.arclen(accuracy) / accuracy).log2().ceil();
```
raphlinus, line 95: Yes, I think this is better logic. But to avoid a loop count of 0, I think it would be better to add a .max(1) on it.

chubei-oppen: Yes, this code panics on large accuracy values. Pull request updated.

cmyr (approved):
Thanks, I'll cut a release shortly.

PR #155, opened 2020-11-14
Add Affine::map_unit_square method.

derekdreery:
I reached for this method today, so I thought I'd post it for consideration.

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -74,6 +74,14 @@ impl Affine {
         Affine([1.0, 0.0, 0.0, 1.0, p.x, p.y])
     }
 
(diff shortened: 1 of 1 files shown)


src/affine.rs
```diff
@@ -74,6 +74,14 @@ impl Affine {
         Affine([1.0, 0.0, 0.0, 1.0, p.x, p.y])
     }
 
+    /// Creates an affine transformation that takes the unit sqaure to the given rectangle.
```
raphlinus, line 77: square

raphlinus (approved):
Yes, I can see this being a useful method.

PR #157, opened 2020-11-22
Add `is_finite` and `is_nan` to most types.

derekdreery:
I didn't add it for `Arc` because I'm planning to redo that anyway based on `Ellipse`. I also skipped svg stuff.

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -123,6 +123,28 @@ impl Affine {
         Rect::from_points(p00, p01).union(Rect::from_points(p10, p11))
     }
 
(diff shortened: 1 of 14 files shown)


raphlinus (approved):
I'm not 100% sure about the need yet, but the methods do make sense, and it's likely that at some point we'll be glad they're there. Thanks!

derekdreery:
In case you're interested my use case is to check a `Rect` is finite before I do some math with it. If I didn't assert it's finiteness, the errors would occur somewhere else and be more cryptic.

PR #158, opened 2020-11-22
Clean up intra-crate docs now they are stable.

derekdreery:
Older versions of rustc will still work but might get warnings. This was the case before this patch, as the crate already uses intra-doc references.


derekdreery:
This should be an easy review - no code it touched.

raphlinus (approved):
Definitely makes the source cleaner (and I'm sure this will increase the motivation to link properly, as it doesn't involve the friction of figuring out the syntax).

It does increase MSRV, but I think for kurbo we're not worrying about that.

derekdreery:
Hopefully the MSRV is the same: it will just print warnings in earlier versions of rust. It should still work.

PR #160, opened 2020-12-05
Fix quadratic bezier flattening

raphlinus:
There was an off-by-one bug in the iteration of the quad bez flatting
loop, resulting in missing the last point and thus getting overly coarse
flattening.


cmyr (approved):
Not a surprise to have one or two of these slip through, sure we'll shake out a few more over time. :)

PR #161, opened 2020-12-20
Using `From` is more general

derekdreery:


derekdreery:
Or is it still going to be removed?

cmyr:
Oops, yes, I want to remove this. I really find that forcing explicit conversions between the 2-tuple types (`to_vec2`, `to_size`, `to_point` etc) make it harder to write an annoying class of bugs.

PR #162, opened 2021-01-08
Add a function to scale a rect about the origin.

derekdreery:


derekdreery:
I could instead implement a general scale that took a center point as well as the scale factor.

raphlinus (approved):
Seems good to me. It's *almost* a case where we could overload `std::ops::Mul`, but I think it's better to have an explicit function.

PR #164, opened 2021-01-26
Give ParamCurveNearest::nearest a struct return type

cmyr:
This has been a minor papercut since forever; I always forget which
argument is which.

Not sure about naming the one field 't'; it might make more sense to call it 'param' or something.

This is a breaking change, and we should do any other breaking we can think of before doing a release.

### DIFF
--- a/src/param_curve.rs
+++ b/src/param_curve.rs
@@ -137,12 +137,27 @@ pub trait ParamCurveArea {
     fn signed_area(&self) -> f64;
 }
 
+/// The nearest position on a curve to some point.
+///
+/// This is returned by [`ParamCurveNearest::nearest`]
+#[derive(Debug, Clone, Copy)]
+pub struct Nearest {
+    /// The square of the distance from the nearest position on the curve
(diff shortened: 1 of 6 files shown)


src/param_curve.rs
```diff
@@ -137,12 +137,27 @@ pub trait ParamCurveArea {
...
 }
 
+/// The nearest position on a curve to some point.
+///
+/// This is returned by [`ParamCurveNearest::nearest`]
+#[derive(Debug, Clone, Copy)]
+pub struct Nearest {
+    /// The square of the distance from the nearest position on the curve
+    /// to the given point.
+    pub distance: f64,
```
raphlinus, line 147: Because this is distance squared (to save a sqrt, as most often in hit testing scenarios a lot of these will be compared), maybe `distance_sq` or `distance2`? Calling it "distance" to me raises the risk it will be misinterpreted.

raphlinus (approved):
Generally good but I do wonder about the name.

PR #165, opened 2021-01-27
Remove Into<Size> for Vec2

cmyr:
We want to force these sorts of conversions to be explicit.

This is itself a breaking change, so should do it with the next version bump.


raphlinus (approved):
Agree, this should be explicit.

PR #166, opened 2021-01-30
Allow independent corner radii for RoundedRect

### LINKED ISSUE #138: Enable `RoundedRect` to change the radius on all four corners
Currently, to do this, you'd have to do the same thing as the `Shape` implementation for `RoundedRect` does, with the exception that each corner could have a different radius (or none). It would also make it much easier to implement styling of all corners in `druid`.

The overhead on `RoundedRect` would be minimal. This could easily be implemented by making `RoundedRectPathIter::arcs` `Option<T>`s and adding three more `f64` for each corner in `RoundedRect`. I'm not sure how a non-rounded corner (ie. zero radius) should be specified. Two possibilities are making the radius variables in `RoundedRect` also `Option` or when creating the `RoundedRectPathIter` comparing the radius to some minimum value.

I will implement this if I get the Ok and some feedback on the problem I mentioned.

SecondFlight:
Closes #138.

This PR allows `RoundedRect` to be created with four radius values instead of one, and updates the `Shape` impl accordingly.

All functions that take a `f64` radius value now take an `impl Into<RoundedRectRadii>` value. `RoundedRectRadii` implements the `from` trait for `f64`, so existing code should not need to change in most cases.

Supporting this in Piet is not trivial as best I can tell. I took a cursory look at the breaking changes and noticed that, while Direct2D supports rounded rectangles, it only supports them with a single radius for all corners. I haven't looked at other platforms.

I am taking the existing unit test for `to_path` as evidence of correctness, but I'd like to ask my reviewer(s) to double-check the ordering of the corners in that function. If I messed up the ordering, then I believe the unit test would still pass.

Please let me know if there's anything I can fix or improve!


src/rounded_rect.rs
```diff
@@ -1,57 +1,73 @@
...
 
 impl RoundedRect {
     /// A new rectangle from minimum and maximum coordinates.
     ///
-    /// The result will have non-negative width, height and radius.
+    /// The result will have non-negative width, height and radii.
     #[inline]
-    pub fn new(x0: f64, y0: f64, x1: f64, y1: f64, radius: f64) -> RoundedRect {
-        RoundedRect::from_rect(Rect::new(x0, y0, x1, y1), radius)
+    pub fn new(x0: f64, y0: f64, x1: f64, y1: f64, radii: RoundedRectRadii) -> RoundedRect {
```
raphlinus, line 43: A possibility here is to have `impl Into<RoundedRectRadii>` here, which would be more ergonomic for allowing a simple scalar radius.

src/rounded_rect.rs
```diff
@@ -149,74 +161,178 @@ impl Shape for RoundedRect {
...
+        ]
+        .iter()
+        {
+            // The following would subtract the square and add the quarter-
+            // circle...
+            // area -= radius * radius;
+            // area += FRAC_PI_4 * radius * radius;
+
+            // ...but it can be simplified as:
+            area += (FRAC_PI_4 - 1.0) * radius * radius;
```
raphlinus, line 250: I believe using `sum` instead of `for` would be slightly more ergonomic, but this is also ok.

src/rounded_rect.rs
```diff
@@ -311,31 +427,32 @@ mod tests {
 
         // Extremum: 0.0 radius corner -> rectangle
         let rect = Rect::new(0.0, 0.0, 100.0, 100.0);
-        let rounded_rect = RoundedRect::new(0.0, 0.0, 100.0, 100.0, 0.0);
+        let rounded_rect = RoundedRect::new(0.0, 0.0, 100.0, 100.0, 0.0.into());
```
raphlinus, line 430: This "into" would not be necessary if we went the "impl" approach above.

src/rounded_rect_radii.rs
```diff
@@ -0,0 +1,104 @@
...
+
+//! A description of the radii for each corner of a rounded rectangle.
+
+use std::convert::From;
+
+/// Radii for each corner of a rounded rectangle.
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct RoundedRectRadii {
+    /// The radius of the top-left corner.
```
raphlinus, line 23: There's a minor pedantic point here. I believe the use of the word "top" here is essentially assuming a y-down coordinate space, but kurbo is also supposed to work in y-up (mathematics) coordinate spaces. I don't think we have to do any kind of awkward renaming, but I do think it would be useful to at least add a comment to the struct explaining the story (I think "top" is always the smaller coordinate, so top and bottom are inverted in a y-up space).

src/rounded_rect_radii.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// Returns true if any corner radius value is NaN.
+    pub fn is_nan(&self) -> bool {
+        self.top_left.is_nan()
+            || self.top_right.is_nan()
+            || self.bottom_right.is_nan()
+            || self.bottom_left.is_nan()
+    }
+}
```
raphlinus, line 92: I think it would also be useful to have `as_single_radius(&self) -> Option<f64>` that succeeds when the radii are all equal. There may well be more efficient rendering paths when that is the case.

raphlinus (commented):
Generally this looks pretty good to me. I'm a little worried about all the edge cases (radii adding up to more than the rectangle side), but overall this looks like a careful job.

SecondFlight:
> I'm a little worried about all the edge cases (radii adding up to more than the rectangle side)

I don't believe this is possible.

`RoundedRectRadii` doesn't know how big the associated rectangle is, so the code to handle that edge-case is in `rounded_rect.rs`, specifically [line 56](https://github.com/linebender/kurbo/blob/3b653957d5e285b9ccf9a747f73c7334a37adadf/src/rounded_rect.rs#L56):
```rust
let radii = radii.abs().clamp(shortest_side_length / 2.0);
```

SecondFlight:
I'll address the remaining comments next time I work on this.

raphlinus:
> I believe this addresses the edge-case

Ok, I think it means nothing can go wrong, but it's also more conservative than it needs to be. For example, in a w x h rectangle, radii of (w, w, 0, 0) are possible, and this is a useful shape to draw. The logic to enforce that is not trivial, and there may not be an unambiguous "correct" solution as in the case of a single radius.

tirix:
> The logic to enforce that is not trivial 

I think *enforcing* would not be a big issue. It may either mean panicking:
```
assert radii.top_left + radii.top_right <= self.width;
assert radii.bottom_left + radii.bottom_right <= self.width;
assert radii.top_left + radii.bottom_left <= self.height;
assert radii.top_right + radii.bottom_right <= self.height;
```
Or clamping in a different way, arbitrarily giving priority to some corner over others:
```
radii.top_left.clamp(width);
radii.top_left.clamp(height);
radii.top_right.clamp(width - radii.top_left);
radii.top_right.clamp(height);
radii.bottom_left.clamp(width);
radii.bottom_left.clamp(height - radii.top_left);
radii.bottom_right.clamp(width - radii.bottom_left);
radii.bottom_right.clamp(height - radii.top_right);
```
However allowing the radii to go past the half dimensions will make several other calculations invalid, this is where it will not be trivial anymore.

src/rounded_rect.rs
SecondFlight: Ah of course, my lack of Rust knowledge is showing here. Thanks for the suggestion!

src/rounded_rect_radii.rs
SecondFlight: Okay, I've documented this at the top of `RoundedRectRadii`.

SecondFlight:
I believe I've addressed all the review comments. Please let me know if I can add or change anything else.

cmyr (approved):
I think this looks good, and I think the API is the right choice; I had initially thought it would make more sense to have separate methods for the non-uniform radii case, but this feels good.

cmyr:
I'll look into doing a kurbo release shortly.

PR #167, opened 2021-02-03
Update manifest to 0.8.0

cmyr:


raphlinus:
The `rand` thing is messy, as it's not needed for kurbo proper. I think it *is* possible to fix by setting the "js" feature when doing wasm builds (see https://docs.rs/getrandom/0.2.2/getrandom/#webassembly-support), but it's also possible the rand-needing examples should be factored out into a separate crate.

A quick fix might be to not run the examples on wasm, but I haven't really investigated.

cmyr:
okay I'm just going to not touch rand for now, this doesn't seem like a major concern.

PR #169, opened 2021-02-16
Add Euler spiral curve representation

raphlinus:
This commit adds support for piecewise Euler spirals, gated under the `euler` feature. A primary goal is to support parallel curves, as described in https://github.com/raphlinus/raphlinus.github.io/issues/46.

There are some rough edges, and there's not an end-to-end parallel curve yet, but it's progress. Here's what's currently implemented:

* The `EulerSeg` curve, along with some `ParamCurve` traits.
* Geometric Hermite interpolation of an Euler spiral segment
* Conversion of `CubicBez` to an iterator of Euler spiral segments
* Conversion of an Euler spiral segment to an iterator of CubicBez

Here's what's missing:
[ ] The actual parallel curve algorithm at the heart of the blog post
[ ] An optimized version of Geometric Hermite interpolation (see blog post)
[ ] Error bounds on the above
[ ] `ParamCurveNearest` and `ParamCurveExtrema` on `EulerSeg`
[ ] Most of `EulerPath` including the `Shape` trait.
[ ] And end-to-end driver

It's *slightly* unclear that we'll even develop `EulerPath` at this point. It's possible to exercise the main parallel curve flow by composing iterators, and ending up with a `BezPath`. Possibly that should be taken out of this PR and added later if/when needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,6 @@ features = ["derive"]
 # This is used for research but not really needed; maybe refactor.
 [dev-dependencies]
 rand = "0.7.2"
+
+[features]
+euler = []
--- a/examples/euler.rs
+++ b/examples/euler.rs
@@ -1,3 +1,17 @@
+// Copyright 2021 The kurbo Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


cmyr:
@raphlinus is this ready? going to do a piet release soon, might as well get whatever outstanding kurbo work merged as makes sense.

raphlinus:
I don't think we should block a piet release on it. I want to make some tweaks to the API (take an accuracy parameter pretty much everywhere), put the euler stuff into an euler namespace instead of top level, and make sure things are labeled as experimental. I'm also increasingly unsure about the "euler" feature - it feels like there should be an "advanced" or "full" feature that would encompass this, #174, and other future stuff along similar lines.

I *can* get this ready but it would be tomorrow.

cmyr:
no stress, I'm not going to get the release done today, although we're not too far off.

PR #170, opened 2021-02-22
Add Affine::IDENTITY constant

cmyr:
🤷


raphlinus (approved):
Sure, looks good.

In slightly related news, I was disappointed to find that Rust doesn't optimize `0.0 +` or `0.0 *` (though it does optimize `1.0 *`), because of NaN handling. There are fast-math intrinsics in unstable, which will help with this problem if and when these stabilize.

See https://www.reddit.com/r/rust/comments/e5ge5k/rust_and_ffastmath/ for more background on that.

PR #171, opened 2021-03-03
Impl Add/Sub Vec2 for RoundedRect

cmyr:


raphlinus (approved):
It just feels so good to me to express these ops using simple mathematical notation :)

cmyr:
doesn't it just?

PR #172, opened 2021-03-10
No std

derekdreery:
I needed a geometry library for a play that I've been having with my PineTime, and so have modified kurbo to work with no_std. I've done this by using a feature `std` that is on-by-default. When in `no_std`, I use `libm` for math routines, and cfg-out things that require an allocator (which is surprising little). I re-export core or std from a private module to enable the rest of the crate to just use `std`.

One change that was required is changing a sort to a sort_unstable, because the former allocates. In practice the change should make little or no difference.

Sharing here in case you are interested in upstreaming no_std support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,11 @@ features = ["derive"]
 # This is used for research but not really needed; maybe refactor.
 [dev-dependencies]
 rand = "0.7.2"
+
+[[example]]
+name = "ellipse"
+required-features = ["std"]
+
+[[example]]
+name = "circle"
+required-features = ["std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,11 +10,16 @@ description = "A 2D curves library"
 readme = "README.md"
 categories = ["graphics"]
 
+[features]
+std = ["arrayvec/std"]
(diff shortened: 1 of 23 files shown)


derekdreery:
This PR should probably not be merged until CI has been turned on for no_std.

cmyr:
Would you like to add a no-std build to the CI as part of this PR? That would make me more comfortable merging..

derekdreery:
I'm conflicted on whether this is worth merging. In places where `no_std` is used, having a FPU isn't a given, and one often wants to work in integers when programming displays. Anything like beziers would be pre-rasterized and stuck in ROM somewhere. Probably would favor simply closing this PR, if you think that's best.

cmyr:
I'd want to have a clear sense that this was going to be used somewhere before merging, because I think it adds fragility to the codebase; there's a full other tree of codepaths to test, etc. If this *does* have a clear use-case then I'm happy to review, although I've never really played around with `no_std` before. 🤷

derekdreery:
Let's close for now. I will keep the branch around in case its useful in the future.

cmyr:
sounds good!

PR #173, opened 2021-03-13
Fancier solver for inverse arclength

raphlinus:
Use the ITP method for solving inverse arclength, which converges more
quickly than bisection but is just as robust.

This patch makes the ITP solver available in `common` as well, as it is
likely to be of general use. It's also nice that the solver is factored
out from the arclength-specific logic.

Also use the simple analytical solution for inverse arc length of a
line.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -148,6 +148,94 @@ pub fn solve_quadratic(c0: f64, c1: f64, c2: f64) -> ArrayVec<[f64; 2]> {
     result
 }
 
+/// Solve an arbitrary function for a zero-crossing.
+///
+/// This uses the [ITP method], as described in the paper
+/// [An Enhancement of the Bisection Method Average Performance Preserving Minmax Optimality].
+///
+/// The values of `ya` and `yb` are given as arguments rather than
+/// computed from `f`, as the values may already be known, or they may
+/// be less expensive to compute as special cases.
+///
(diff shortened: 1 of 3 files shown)


src/common.rs
```diff
@@ -148,6 +148,94 @@ pub fn solve_quadratic(c0: f64, c1: f64, c2: f64) -> ArrayVec<[f64; 2]> {
...
+///
+/// This uses the [ITP method], as described in the paper
+/// [An Enhancement of the Bisection Method Average Performance Preserving Minmax Optimality].
+///
+/// The values of `ya` and `yb` are given as arguments rather than
+/// computed from `f`, as the values may already be known, or they may
+/// be less expensive to compute as special cases.
+///
+/// It is assumed that `ya < 0.0` and `yb > 0.0`, otherwise unexpected
+/// results may occur.
```
cmyr, line 161: is it worth enforcing this with an assert?

cmyr (approved):
I am going to reluctantly recuse myself from actually spending the time with the paper and trying to bench-check your work, which doesn't feel like a very good use of a day for me right now; feel free to merge when you're satisfied. :)

src/common.rs
raphlinus: I don't think it should be an assert, because weird stuff happens sometimes and crashing would be unfriendly. The client of this function is expected to be pretty sophisticated about what they want, and the point of this comment is to set expectations.

PR #181, opened 2021-06-10
Add derive PartialEq to BezPath

0HyperCube:
Add derive `PartialEq` for `BezPath` for comparing `BezPath`s

(for [Graphite](https://github.com/GraphiteEditor/Graphite))


cmyr:
Sorry for letting these sit.

These were originally omitted out of concern about floating point comparisons, and I would really like to figure out a better long-term solution, but given that we have already caved and added `PartialEq` for a bunch of other types I'm happy to also add it here.

cmyr (approved):
Thanks!

Issue #182, opened 2021-06-12
Some desired methods


simoncozens:
```rust
impl PathSeg {
fn start(&self) -> Point;
fn end(&self) -> Point;
// Because pattern matching just to get start/end points is annoying, and using to_cubic is hacky.
}

impl Point {
fn square_distance(&self, other: Point) -> f64;
fn unitize(&self) -> Point;
fn angle(&self) -> f64;
fn dot(&self, other: Point) -> f64;
fn cross(&self, other: Point) -> f64;
}
```

Probably more as I think of them. I don't know if you want to turn every struct into a Ruby-like grab-bag of methods, but these are things I end up reaching for frequently.

raphlinus:
The first Point method makes sense but it can also be written as (p1 - p2).hypot2(). The other methods are not mathematically meaningful because they are operations on Vec2, not Point.

Early on we did not have a separate Point type, rather used Vec2 for everything. Sometime later, I was swayed by the argument that we should use the type system. That causes a bunch of .to_vec2() and .to_point() calls, which don't do anything. Was that a mistake? Should we go back?

raphlinus:
Thinking about this a little more, I am open to taking a different point on the ergonomics/correctness tradeoff. We keep Vec2 and Point as separate types for essentially documentation reasons, but implement basically the same functionality. This would have the advantage of making 0.5 * (p1 + p2) work, which is currently annoying. There would be some minor strangeness, like (p1 + p2) - p3 being a Vec2 (currently a type error), while p1 + (p2 - p3) is a Point.

One of the differences between Point and Vec2 is that, in theory, Affine * Vec2 might multiply just the 2x2 scale/rotation matrix and ignore the translation. However, that isn't implemented and I think to avoid confusion we should probably keep it that way.

cmyr:
In my own work I've found the distinction between vec and point to be generally more helpful than harmful, and I'd be inclined to keep it?

Given that we've added `Point::distance` I think `Point::distance_squared` is a reasonable addition.

Other things being written with `to_vec2()` conversions feels *okay*:


```rust
// unitize:
point.to_vec2().normalize().to_point();
// angle
point.to_vec2().atan2();
// dot
point.to_vec2().dot(other.to_vec2())
// cross
point.to_vec2().cross(other.to_vec2())
```
```

derekdreery:
[...]

PR #184, opened 2021-06-25
Add Point::distance_squared

cmyr:
from #182


raphlinus (approved):
Seems good to me, thanks!

PR #186, opened 2021-07-19
Fix some clippy warnings

remilauzier:


remilauzier:
I have no idea on how to fix wasm test.

cmyr (commented):
iirc there was an issue with `rand` and WASM which was the reason we're still on 0.7.x. I would be happy to merge the clippy changes, though, if you wanted to just PR those?

cmyr (approved):
Thanks!

PR #187, opened 2021-07-30
Fix some typo and warning found by intellij idea

remilauzier:


src/common.rs
```diff
@@ -165,7 +165,7 @@ pub fn solve_quadratic(c0: f64, c1: f64, c2: f64) -> ArrayVec<[f64; 2]> {
 /// represent the lower and upper bounds of the bracket searched for a
 /// solution.
 ///
-/// The ITP method has tuning parameters. This implementation hardwires
+/// The ITP method has tuning parameters. This implementation hardwired
```
raphlinus, line 168: In this case, "hardwires" is better, I think.

raphlinus (approved):
Thanks! Most of these (with one small exception) are improvements. I think many of the commas are judgment calls, but are fine.

src/arc.rs
```diff
@@ -136,16 +136,16 @@ impl Shape for Arc {
         iter::once(PathEl::MoveTo(self.center + p0)).chain(self.append_iter(tolerance))
     }
 
-    /// Note: shape isn't closed so area is not well defined.
+    /// Note: shape isn't closed so area is not well-defined.
```
cmyr, line 139:
```suggestion
    /// Note: shape isn't closed so area is not well defined.
```

https://english.stackexchange.com/questions/410283/hyphen-well-defined-vs-well-defined

src/circle.rs
```diff
@@ -103,7 +103,7 @@ impl Shape for Circle {
         } else {
             // This is empirically determined to fall within error tolerance.
             let n = (1.1163 * scaled_err).powf(1.0 / 6.0).ceil() as usize;
-            // Note: this isn't minimum error, but it is simple and we can easily
+            // Note: this isn't minimum error, but it is simple, and we can easily
```
cmyr, line 106:
```suggestion
            // Note: this isn't minimum error, but it is simple and we can easily
```

this works well enough without the comma splice :)

src/cubicbez.rs
```diff
@@ -235,7 +235,7 @@ impl ParamCurveArclen for CubicBez {
         // A degenerate curve where all points are identical will cause infinite recursion in the rec function (well, until MAX_DEPTH at least) in all branches.
         // This test will in addition be true if the bezier curve is just a simple line (i.e. p0=p1 and p2=p3).
         // The constant 0.5 has not been mathematically proven to be small enough, but from empirical tests
-        // a value of about 0.87 should be enough. Thus 0.5 is a conservative value.
+        // a value of about 0.87 should be enough. Thus, 0.5 is a conservative value.
```
cmyr, line 238:
```suggestion
        // a value of about 0.87 should be enough. Thus 0.5 is a conservative value.
```

ditto

src/ellipse.rs
```diff
@@ -13,8 +13,8 @@ use crate::{Affine, Arc, ArcAppendIter, Circle, PathEl, Point, Rect, Shape, Size
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Ellipse {
     /// All ellipses can be represented as an affine map of the unit circle,
-    /// centred at (0, 0). Therefore we can store the ellipse as an affine map,
-    /// with the implication it be applied to the unit circle to recover the
+    /// centred at (0, 0). Therefore, we can store the ellipse as an affine map,
```
cmyr, line 16:
```suggestion
    /// centred at (0, 0). Therefore we can store the ellipse as an affine map,
```

src/ellipse.rs
```diff
@@ -13,8 +13,8 @@ use crate::{Affine, Arc, ArcAppendIter, Circle, PathEl, Point, Rect, Shape, Size
...
+    /// with the implication it is applied to the unit circle to recover the
```
cmyr, line 17:
```suggestion
    /// with the implication that it be applied to the unit circle to recover the
```

src/ellipse.rs
```diff
@@ -117,7 +117,7 @@ impl Ellipse {
 
     /// Returns the two radii of this ellipse.
     ///
-    /// The first number is the horizontal radius and the second is the vertical
+    /// The first number is the horizontal radius, and the second is the vertical
```
cmyr, line 120:
```suggestion
    /// The first number is the horizontal radius and the second is the vertical
```

The trailing comma is to join the dependent clause "before rotation", but a simple "and" works here

src/shape.rs
```diff
@@ -55,7 +55,7 @@ pub trait Shape: Sized {
     /// Convert to a Bézier path.
     ///
     /// This always allocates. It is appropriate when both the source
-    /// shape and the resulting path are to be retained.
+    /// shape, and the resulting path are to be retained.
```
cmyr, line 58:
```suggestion
    /// shape and the resulting path are to be retained.
```

cmyr (approved):
Thanks, you've caught a lot of typos here.

I'm going to pass on some of the additional commas, generally where the word "and" is already serving to join two clauses. I'll go ahead and change those back myself and then merge.

Thanks again!

src/arc.rs
```diff
@@ -154,7 +154,7 @@ impl Shape for Arc {
         self.path_segments(0.1).perimeter(accuracy)
     }
 
-    /// Note: shape isn't closed so a point's winding number is not well defined.
+    /// Note: shape isn't closed, so a point's winding number is not well-defined.
```
cmyr, line 157:
```suggestion
    /// Note: shape isn't closed, so a point's winding number is not well defined.
```

PR #188, opened 2021-08-04
Overwrite `contains` shape method on rect

JAicewizard:
The normal `contains` is slow because it uses winding. We can do better.

### DIFF
--- a/src/rect.rs
+++ b/src/rect.rs
@@ -621,6 +621,11 @@ impl Shape for Rect {
     fn as_rect(&self) -> Option<Rect> {
         Some(*self)
     }
+
+    #[inline]
+    fn contains(&self, pt: Point) -> bool {
+        self.contains(pt)
+    }
 }
 
 // This is clockwise in a y-down coordinate system for positive area.


src/rect.rs
```diff
@@ -621,6 +621,11 @@ impl Shape for Rect {
     fn as_rect(&self) -> Option<Rect> {
         Some(*self)
     }
+
+    #[inline]
+    fn contains(&self, pt: Point) -> bool {
+        self.contains(pt)
```
cmyr, line 627: I'm surprised this works without complaint, because `contains` is ambiguous? I would expect to have to do `Rect::contains(pt)`.

In any case, I agree this makes sense, although it will only matter in very specific circumstances; normally a method call that exists on a trait and a concrete type will resolve to the impl on the concrete type, so this would only be used in situations where something was passed as a generic `Shape`. Still worth doing though!

JAicewizard: I am using generic shapes and this was a semi-major bottleneck for me. Not a real bottleneck, still waiting on my `image` improvements to land but still a good 30% or so. This will only help when I am using rects, but that will be the majority of the time.

For other types `winding` is basically `contains` so those dont have as big of a performance hit.

> I'm surprised this works without complaint

IIRC the default is always on the type itself if it exists, only when it cant find that traits are considered.

cmyr (approved):
Thanks, will include this in the next release. :)

PR #189, opened 2021-08-25
Compute the minimum distance between two pathsegs

simoncozens:
This implements the [Chen et al.](https://www.sciencedirect.com/science/article/pii/S0377042708005785) sweeping sphere algorithm for minimum distance finding; works on all Bézier orders, including lines.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -9,10 +9,11 @@ use std::ops::{Mul, Range};
 use arrayvec::ArrayVec;
 
 use crate::common::{solve_cubic, solve_quadratic};
+use crate::mindist::min_dist_param;
 use crate::MAX_EXTREMA;
 use crate::{
     Affine, CubicBez, Line, Nearest, ParamCurve, ParamCurveArclen, ParamCurveArea,
-    ParamCurveExtrema, ParamCurveNearest, Point, QuadBez, Rect, Shape, TranslateScale,
+    ParamCurveExtrema, ParamCurveNearest, Point, QuadBez, Rect, Shape, TranslateScale, Vec2,
 };
 
 /// A Bézier path.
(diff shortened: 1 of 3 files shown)


src/bezpath.rs
```diff
@@ -942,6 +943,35 @@ impl PathSeg {
             PathSeg::Cubic(cubic_bez) => cubic_bez.is_nan(),
         }
     }
+
+    #[inline]
+    fn as_vec2_vec(&self) -> Vec<Vec2> {
```
raphlinus, line 948: Returning an ArrayVec would avoid the allocation.

src/mindist.rs
```diff
@@ -0,0 +1,274 @@
...
+    let (vmin, vmax) = v;
+    let umid = (umin + umax) / 2.0;
+    let vmid = (vmin + vmax) / 2.0;
+    let svalues: [(f64, f64, f64); 4] = [
+        (S(umin, vmin, bez1, bez2), umin, vmin),
+        (S(umin, vmax, bez1, bez2), umin, vmax),
+        (S(umax, vmin, bez1, bez2), umax, vmin),
+        (S(umax, vmax, bez1, bez2), umax, vmax),
+    ];
+    let alpha: f64 = svalues.iter().map(|(a, _, _)| *a).fold(f64::NAN, f64::min);
```
raphlinus, line 28: I think `reduce` is slightly nicer than `fold` here, because you avoid the nan (there is an unwrap but it should be compiled out because the size is known).

src/bezpath.rs
```diff
@@ -942,6 +943,35 @@ impl PathSeg {
...
+                    c.p0.to_vec2(),
+                    c.p1.to_vec2(),
+                    c.p2.to_vec2(),
+                    c.p3.to_vec2(),
+                ]
+            }
+        }
+    }
+
+    /// Minimum distance between two PathSegs
```
raphlinus, line 963: Some questions. Do the pathsegs need to be of the same type (line/quad/cubic), or can it do the degree raising as needed?

Also, I see it returning the distance, t1, and t2. This should be stated in the docstring.

src/mindist.rs
```diff
@@ -0,0 +1,274 @@
...
+            (new_umid, umax),
+            (new_vmid, vmax),
+            epsilon,
+            Some(alpha),
+        ),
+    ];
+
+    *results
+        .iter()
+        .min_by(|a, b| a.0.partial_cmp(&b.0).unwrap_or(Less))
```
raphlinus, line 140: I would slightly prefer `Ordering::Less` here.

raphlinus (commented):
Generally this looks good but I had a few comments. I haven't had a chance to dig deeply into the algorithm or compare it against the paper, but it looks like a

Also, I mentioned on Zulip that I preferred an L2 norm, but I was confusing Hausdorff-style metrics with what's being calculated here, which as I understand is a generalization of `nearest` to two curves, as opposed to the existing point/curve.

This looks like a very useful feature, thanks!

src/bezpath.rs
simoncozens: The segs don't need to be the same order, but it doesn't do degree raising either - it works in terms of basis functions, and it has sums that do things like `∑i=0..order(Curve1) ∑j=0..order(Curve2) f(Curve1, i) ⋅ g(Curve2, j)`.

PR #191, opened 2021-08-30
Fix wrapping of subtraction. Fixes #190

### LINKED ISSUE #190: Overflow in mindist
I've done something bad here:

```
Cubic(CubicBez { p0: (232.0, 126.0), p1: (134.0, 126.0), p2: (139.0, 232.0), p3: (141.0, 301.0) }) <-> Line(Line { p0: (359.0, 416.0), p1: (367.0, 755.0) })
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /Users/simon/.cargo/git/checkouts/kurbo-aa548d4979964553/e964d99/src/mindist.rs:206:49
```

https://github.com/linebender/kurbo/blob/e964d99f9e0643e4f3457cceb0166d39e91e5b80/src/mindist.rs#L206

simoncozens:


src/mindist.rs
```diff
@@ -271,4 +272,32 @@ mod tests {
...
+    fn test_out_of_order() {
+        let bez1 = PathSeg::Cubic(CubicBez::new(
+            (287.0, 182.0),
+            (346.0, 277.0),
+            (356.0, 299.0),
+            (359.0, 416.0),
+        ));
+        let bez2 = PathSeg::Line(Line::new((141.0, 301.0), (152.0, 709.0)));
+        let (dist1, _t1, _t2) = bez1.min_dist(bez2, 0.5);
+        println!("\n\n\n");
```
cmyr, line 299: maybe unintentional?

src/mindist.rs
```diff
@@ -203,7 +203,7 @@ fn D_rk(r: usize, k: usize, bez1: &[Vec2], bez2: &[Vec2]) -> f64 {
 
 // Bezier basis function
 fn basis_function(n: usize, i: usize, u: f64) -> f64 {
-    choose(n, i) as f64 * (1.0 - u as f64).powi((n - i) as i32) * u.powi(i as i32)
+    choose(n, i) as f64 * (1.0 - u as f64).powi(n as i32 - i as i32) * u.powi(i as i32)
```
cmyr, line 206: What expectation do we have for inputs, and what is the desired behaviour? are we happy to see negative powers here?

src/mindist.rs
```diff
@@ -149,7 +149,7 @@ fn S(u: f64, v: f64, bez1: &[Vec2], bez2: &[Vec2]) -> f64 {
     let m = bez2.len() - 1;
```
cmyr, line 149: Just out of curiousity, is it possible for this to receive 0-length inputs?

src/mindist.rs
```diff
@@ -225,6 +225,7 @@ fn choose(n: usize, k: usize) -> u32 {
 mod tests {
     use crate::mindist::A_r;
     use crate::mindist::{choose, D_rk};
+    use crate::Line;
```
cmyr, line 228: can combine this with the next import? 🤷

cmyr (approved):
Looks good, a few little things you are free to ignore.

src/mindist.rs
simoncozens: These are fed from a loop which ranges over 0..order_of_bezier

src/mindist.rs
simoncozens: Nope, because it's called with the results of `as_vec2_vec`. But some asserts wouldn't be bad.

src/mindist.rs
raphlinus: So i >n is possible? That really shouldn't be, it would also be a divide by zero when u is 0 or 1. Can you explain how this might happen?

simoncozens: I think this was fixing the symptom, whereas the next commit actually fixed the disease.

raphlinus (approved):
Looks good, thanks!

Issue #192, opened 2021-09-04
Interpolatable cubic-to-quadratic


simoncozens:
One facility which would be very useful (all right, *indispensable*) for font work is the ability to convert *multiple* cubic curves to quadratics together, with the guarantee that the returned curves have the same structure (same number of quadratics and, ideally, the same elidable "implied oncurves").

fontmake's [cu2qu](https://github.com/googlefonts/cu2qu/blob/main/Lib/cu2qu/cu2qu.py) has an algorithm with these guarantees; we could steal it, or do something clever ourselves.

raphlinus:
This is an interesting problem for sure! I think it's possible to do better than cu2qu, but of course there are complex tradeoffs. I have no objection to making the cu2qu algorithm or something like it available (especially to make results that reproduce existing tools), but I am also interested in something that is more "interpolation compatible" by construction, and plan to spend some research time on it. I can move up the priority of that if it's starting to block production of variable fonts.

yisibl:
Does something similar to this function do?https://docs.rs/lyon/0.11.0/lyon/algorithms/geom/cubic_to_quadratic/fn.cubic_to_quadratics.html

raphlinus:
Not very similar, as that function takes *one* cubic, and the method here takes a slice of cubics, mapping them to a sequence of quadratic b-splines that are "interpolation compatible." This is really something extremely specialized to the production of variable fonts.

PR #193, opened 2021-09-07
Cubic to quadratic spline conversion

simoncozens:
As mentioned in #192. This is a straight port of the algorithms in `cu2qu`. *I do not claim to understand them*.

For discussion...


raphlinus (commented):
Overall looks good, and a welcome addition. I do have a bunch of comments, most having to do with Rust style. The biggest ask is a newtype for quadratic B-splines.

I'm curious about the 3/8 subdivision also. I recall talking with Behdad about some of this stuff, and it rings a bell, but I can't reconstruct it. Perhaps worth asking.

src/line.rs
```diff
@@ -36,6 +36,19 @@ impl Line {
         self.arclen(DEFAULT_ACCURACY)
     }
 
+    /// Computes the intersection point of two lines.
```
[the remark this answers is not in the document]
simoncozens: It does indeed. I think I would like us to think more about intersections. I'll raise a separate issue for that.

I shall call this one `crossing_point` or something.

simoncozens:
> I'm curious about the 3/8 subdivision also. I recall talking with Behdad about some of this stuff, and it rings a bell, but I can't reconstruct it. Perhaps worth asking.

In all cases I've found, the result is the same.

src/cubicbez.rs
```diff
@@ -73,6 +76,203 @@ impl CubicBez {
...
+            .fit_inside(accuracy)
+            {
+                return None;
+            }
+            return Some(QuadBez::new(self.p0, q1, self.p3));
+        }
+        None
+    }
+
+    fn split_into_n(&self, n: usize) -> Vec<CubicBez> {
```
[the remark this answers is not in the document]
simoncozens: If you remove the case analysis, the tests start to fail. So something else is going on here...

src/cubicbez.rs
```diff
@@ -73,6 +76,197 @@ impl CubicBez {
...
+            .fit_inside(accuracy)
+            {
+                return None;
+            }
+            return Some(QuadBez::new(self.p0, q1, self.p3));
+        }
+        None
+    }
+
+    fn split_into_n(&self, n: usize) -> impl Iterator<Item = CubicBez> {
```
cmyr, line 163: out of curiosity: do we have a sense of the distribution of `n` in calls to this function? If it is more likely to be a small value than a large value it might be worth getting a little fancy in order to avoid the allocation.

Looking through this a bit more, I think it should be fairly trivial to avoid the allocation, if so desired, and convert this whole method to use `std::iter::from_fn`?

src/cubicbez.rs
```diff
@@ -73,6 +76,197 @@ impl CubicBez {
...
+            mid2,
+            mid2 + deriv2,
+            ((p2 + 2.0 * p3) / 3.0).to_point(),
+            self.p3,
+        );
+        (left, mid, right)
+    }
+
+    /// Does this curve fit inside the given distance from the origin?
+    pub fn fit_inside(&self, distance: f64) -> bool {
```
cmyr, line 255: Is this intended as public API, or is it an implementation detail of the other methods?

src/cubicbez.rs
```diff
@@ -343,11 +537,34 @@ impl Iterator for ToQuads {
...
+    let mut splines: Vec<Option<QuadSpline>> = vec![None; curves.len()];
+    let mut last_unsuccessful_i = 0;
+    let mut split_order = 1;
+    let mut i = 0;
+    while split_order <= MAX_SPLINE_SPLIT {
+        if let Some(spline) = curves[i].approx_spline_n(split_order, accuracy) {
+            splines[i] = Some(spline);
+            i = (i + 1) % curves.len();
+            if i == last_unsuccessful_i {
+                return splines.into_iter().collect::<Option<Vec<_>>>();
```
cmyr, line 553: just to make sure I understand the logic:

- for each number of splits up to MAX, we try to generate a quadratic from the input
- if we succeed for each input, for a given split count, we return the quads
- if we fail for any input at a given split count we try the next split count
- if we try them all and fail, we bail?

Assuming that is true:
- I might prefer a `Result` to an `Option`, but not zealously
- I wonder if there's a more idiomatic approach? Maybe something like,

```rust
pub fn cubics_to_quadratic_splines(curves: &[CubicBez], accuracy: f64) -> Option<Vec<QuadSpline>> {
    let mut result = Vec::new();
    let mut split_order = 0;

    while split_order <= MAX_SPLINE_SPLIT {
        split_order += 1;
        result.clear();

        for curve in curves {
            match curve.approx_spline_n(split_order, accuracy) {
                Some(spline) => result.push(spline),
                None => break,
            }
        }

        if result.len() == curves.len() {
            return Some(result);
        }
    }
    None
}
```

partially this is about 'feel', but there are also potential performance costs to indexing into arrays in a tight loop, since each index op can generate a bounds check, and using iterators avoids this.

src/quadspline.rs
```diff
@@ -0,0 +1,20 @@
+//! Quadratic Bézier splines.
+use crate::Point;
+
+#[derive(Clone, Debug, PartialEq)]
+#[allow(missing_docs)]
```
cmyr, line 5:
```suggestion
/// A quadratic Bézier spline.
```

src/quadspline.rs
```diff
@@ -0,0 +1,20 @@
...
+pub struct QuadSpline(Vec<Point>);
+
+impl QuadSpline {
+    /// Construct a new quadratic spline from an array of points.
```
cmyr, line 9:
```suggestion
    /// Construct a new `QuadSpline` from an array of [`Point`]s.
```

might as well link this to the docs for `Point`

src/quadspline.rs
```diff
@@ -0,0 +1,20 @@
...
+    #[inline]
+    pub fn new(points: Vec<Point>) -> Self {
+        Self(points)
+    }
+
+    /// Return the control points of the spline.
```
cmyr, line 15:
```suggestion
    /// Return the spline's control [`Point`]s.
```

cmyr (commented):
piling on with a few notes & observations 😅

src/cubicbez.rs
simoncozens: As it happens, yes: https://github.com/googlefonts/cu2qu/blob/6812ae48c6b81aee95949d64e2a9489c333d825d/tests/cu2qu_test.py#L70-L77

```python
        expected = {
            2: 6,
            3: 26,
            4: 82,
            5: 232,
            6: 360,
            7: 266,
            8: 28}
```

src/cubicbez.rs
simoncozens: I very often find that if something is a useful routine for an implementor, it ends up also being a useful routine for a user who wants to do something almost, but not quite, what the implementor was doing.

src/cubicbez.rs
cmyr: okay lemme take a poke at this and see if I can get rid of the allocation.

src/cubicbez.rs
simoncozens: You have the logic right. I’ll try reworking with your approach.

src/cubicbez.rs
cmyr: That's a reasonable concern, but it's also worth weighing against the fact that `pub` isn't free, in that it now makes this part of the API for the purposes of semver. If we were to change the name or add a parameter or something in the future, it would be a breaking change, and require a new major version.

My inclination is to keep things private until there is a demonstrated use case. If you think there is a reasonable case for this, then by all means I'm happy to keep it. :)

src/cubicbez.rs
cmyr: Okay I got a bit nerd-sniped here. There's a non-allocating version at https://gist.github.com/cmyr/1d6c316bec50c4d23e4be2c794fe6b89, feel free to take it if you like?

src/cubicbez.rs
```diff
@@ -73,6 +76,203 @@ impl CubicBez {
...
+            let b1 = (3.0 * a * t1 + b) * delta_2;
+            let c1 = (2.0 * b * t1 + c + 3.0 * a * t1_2) * dt;
+            let d1 = a * t1 * t1_2 + b * t1_2 + c * t1 + d;
+
+            r.push(CubicBez::from_parameters(a1, b1, c1, d1))
+        }
+        r
+    }
+
+    fn parameters(&self) -> (Vec2, Vec2, Vec2, Vec2) {
```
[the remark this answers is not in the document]
simoncozens: Finding the cubic polynomial parameters turns out to be useful for the curve-curve intersection code...

src/quadspline.rs
```diff
@@ -0,0 +1,20 @@
+//! Quadratic Bézier splines.
+use crate::Point;
+
+#[derive(Clone, Debug, PartialEq)]
+/// A quadratic Bézier spline.
```
cmyr, line 5:
```suggestion
/// A quadratic Bézier spline.
#[derive(Clone, Debug, PartialEq)]
```

house style: docs above attributes. 🤷

cmyr (approved):
One last nit but I'd be happy to merge this!

src/cubicbez.rs
```diff
@@ -73,6 +76,224 @@ impl CubicBez {
...
+        if self.p2.to_vec2().hypot() <= distance && self.p1.to_vec2().hypot() <= distance {
+            return true;
+        }
+        let mid =
+            (self.p0.to_vec2() + 3.0 * (self.p1.to_vec2() + self.p2.to_vec2()) + self.p3.to_vec2())
+                * 0.125;
+        if mid.hypot() > distance {
+            return false;
+        }
+        // Split in two. Note that cu2qu here uses a 3/8 subdivision. I don't know why.
```
behdad, line 292: Humm. Does it?
https://github.com/fonttools/fonttools/blob/48f68877a8ab5f868f57926b14eb4af1c811dfae/Lib/fontTools/cu2qu/cu2qu.py#L169-L174

looks like split at .5 to me.

src/cubicbez.rs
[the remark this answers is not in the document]
behdad: The case analysis was to speed up the Python version...

Issue #194, opened 2021-09-09
Segment/segment intersections


simoncozens:
We currently have a line/segment intersection method, which is great, but I think a generic segment/segment intersection trait would be very useful: line/line, line/curve, curve/curve. I have [this code in beziers.py](https://github.com/simoncozens/beziers.py/blob/master/beziers/utils/intersectionsmixin.py) and not having it in kurbo is a blocker for me in moving some projects to kurbo. 

The only problem is that segment/segment intersections have *many* gnarly edge cases, and I'm not 100% convinced that my code handles all of them.

simoncozens:
I have got most of this implemented, leaning on `intersect_line` for everything apart from the curve/curve intersection case, but I'm having problems terminating the bisection at the right point. (Which I think was a problem with the Python implementation too.)

tylers-epilog:
Hi Simon.

Is this still being worked on? I'd like to contribute to this library by providing a way to add operators to paths. For example, get the union of two paths, or subtracting filled shapes of one path from another. Intersections between Beziers/line is necessary to do this.

If you have any code that you can provide for a starting point, that would be great. If not, I plan to add the GeoClip intersection algorithm. I have quite a bit of experience in a similar algorithm, BezClip, so I'm not going into this completely blind, but if you have any tips from your past experience on doing this, please let me know

Thanks,
Tyler

simoncozens:
I implemented it in #199; waiting for review...

raphlinus:
Ah, I should take a look at that. On quick skim, it looks pretty good. I'm uneasy about adding a new trait, as those are really intended for user-defined curves (think Euler spirals and similar) and I doubt that trait will get implemented for anything other than existing beziers. (See #6 for thoughts). But that's not a showstopper.

If someone else knowledgeable were to go through that code carefully, I'd be happy to approve after another light skim myself.

simoncozens:
#219 supersedes #199.

PR #195, opened 2021-09-09
Curve curve intersection

simoncozens:
As discussed in #194. Doesn't *quite* work.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -141,6 +141,23 @@ pub struct LineIntersection {
     pub segment_t: f64,
 }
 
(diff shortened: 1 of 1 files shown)


raphlinus (commented):
It's not surprising this doesn't work well - when an intersection is extremely near a subdivision point, then due to floating point roundoff it's possible it can end up in both or neither of the subdivisions.

This is a legitimately tricky problem to get right, the robustness even more so than efficiency. It might be worth taking a look at the lyon code - and also might be good to know that we have permission to adapt that code as long as attribution is given (they adapted some of our flattening code).

PR #196, opened 2021-09-14
Specialize `inv_arclen` for `PathSeg` and `ConstPoint`

Sh4rK:
The default implementation of `ParamCurveArclen::inv_arclen` uses an
iterative solver. For some types we can skip that:
- `PathSeg`: forward it to the inner implementation (useful for `Line`)
- `ConstPoint`: always return 0

As an additional thought, maybe it would be better to not have a default implementation of some of these methods, as it can accidentally create inefficiencies when left unimplemented for specific cases. Instead, there could be easy-to-use functions to call from the methods when this behavior is desired in concrete instances.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -677,6 +677,14 @@ impl ParamCurveArclen for PathSeg {
             PathSeg::Cubic(cubic) => cubic.arclen(accuracy),
         }
     }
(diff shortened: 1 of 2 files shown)


cmyr (approved):
Thanks!

PR #200, opened 2021-09-17
Add a custom return type for BezPath::min_dist

cmyr:
This is part of a bit of cleanup before doing a release, and includes a
few other doc and license fixes. It also makes min_dist_param pub(crate)
instead of pub; mindist was not exported to this isn't available as
public API in any case. If someone specifically needs to use this at
some point we can consider making it public then.

@simoncozens does this make sense for you? The one main question I have is around whether we should be returning distance or distance2 from `min_dist`. [`ParamCurve::nearest`](https://docs.rs/kurbo/0.8.1/kurbo/struct.Nearest.html) returns the distance2, is there a reason to prefer distance here? I think it makes sense to have the same behaviour in both places, though.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -9,7 +9,6 @@ use std::ops::{Mul, Range};
 use arrayvec::ArrayVec;
 
 use crate::common::{solve_cubic, solve_quadratic};
-use crate::mindist::min_dist_param;
 use crate::MAX_EXTREMA;
 use crate::{
     Affine, CubicBez, Line, Nearest, ParamCurve, ParamCurveArclen, ParamCurveArea,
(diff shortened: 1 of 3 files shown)


simoncozens:
I'm not convinced about the analogy with `nearest`; I think there is a difference between `min_dist` and `nearest`.

For `nearest`, the user is asking for the *nearest point*, and gets a point together with the (square) distance returned as a side-effect of computation. For `min_dist`, the user is asking for a *distance*, and gets the points returned as a side-effect of computation.

If I ask the library for a distance, I would be surprised to get a square distance returned by default.

cmyr:
Okay I think that's convincing, thanks!

PR #202, opened 2021-09-17
Fix winding orientation in line case

### LINKED ISSUE #180: BezPath::contains returns incorrect results
## Description

The current implementation of `BezPath::contains` gives incorrect results in many cases.

## Environment

* Rustc - 1.52.0
* Kurbo - 0.8.1 (latest)

## Steps to Reproduce

Compile and run the example below as `main.rs`.  This compares against a different check of points being inside the triangle, though based on [the note on `PathSeg::intersect_line`](https://docs.rs/kurbo/0.8.1/kurbo/enum.PathSeg.html#method.intersect_line), this is not a reliable check.

```rust
use kurbo::{BezPath, Line, Point, Shape};

fn contains_count_intersections(path: &BezPath, point: Point) -> bool {
    let bbox = path.bounding_box();
    let outside_point = kurbo::Point::new(bbox.min_x() - 1.0, bbox.min_y());
    let line = Line::new(point, outside_point);
    let num_intersections = path
        .segments()
        .map(|seg| seg.intersect_line(line).len())
        .sum::<usize>();

    num_intersections % 2 != 0
}

fn main() {
    path.move_to((0.0, 0.0));
    path.line_to((1.0, 1.0));
    path.line_to((2.0, 0.0));
    path.close_path();

    let test_point = Point::new(1.0, 0.5);

    println!(
        "path.contains({}) = {}",
        test_point,
        path.contains(test_point),
    );
    println!(
        "contains_count_intersections({}) = {}",
        test_point,
        contains_count_intersections(&path, test_point),
    );
}
```

## Expected behavior

Program runs and outputs

```
path.contains((1, 0.5)) = true
contains_count_intersections((1, 0.5)) = true
```

## Observed behavior

Program runs and outputs

```
path.contains((1, 0.5)) = false
contains_count_intersections((1, 0.5)) = true
```

raphlinus:
The line case of PathSeg winding number calculation had its sign wrong.

Fixes #180

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -758,7 +758,7 @@ impl PathSeg {
                 let a = end.y - start.y;
                 let b = start.x - end.x;
                 let c = a * start.x + b * start.y;
-                if (a * p.x + b * p.y - c) * (sign as f64) >= 0.0 {
+                if (a * p.x + b * p.y - c) * (sign as f64) <= 0.0 {
                     sign
                 } else {
                     0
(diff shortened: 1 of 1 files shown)


cmyr (approved):
😅

PR #204, opened 2021-09-17
Update arrayvec to use const generics

cmyr:
It's unclear to me whether or not this is a breaking change? I think it isn't, since the return types should behave identically, but I'm not a very good language lawyer..

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ categories = ["graphics"]
 features = ["mint", "serde"]
 
 [dependencies]
-arrayvec = "0.5.1"
+arrayvec = "0.7.1"
 
 [dependencies.mint]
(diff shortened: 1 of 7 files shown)


raphlinus (approved):
I'm in favor of the change. I think it may be a minor breaking change, but happy to put it in the category of "unlikely to break any real code," as I'm sure most if not all clients just use the value directly.

cmyr:
That means you would not merge this patch, and keep the breaking change? Would you go to `0.9.0`? My feeling was that if we could revert this and avoid needing to bump the minor version that would be worth it. 🤷

cmyr:
I am going to take the position that this is not a semver breaking change, and if it turns out to actually break code we can yank. 🤷

PR #210, opened 2021-12-22
Sin cos optimization

### LINKED ISSUE #209: Use `sin_cos` for slightly faster computation
https://github.com/linebender/kurbo/blob/1ca29449bb4e654baa45ee7f9b901d642c1efc1d/src/affine.rs#L69

`sin` and `cos` are calculated using some expansion or other (I assume Taylor series?). There exists a method `f64::sin_cos` that calculates both at the same time, saving a few numeric operations. Using this function here and anywhere else you need both results will result in a slight speedup.

VanSHOE:
Fixes #209 

Replaced all the sin and cos simultaneous instances and tried to keep the same format and style of the code
Please have a look and let me know if this helps
In src/circle.rs I reversed the order of sin and cos and tried to ensure it did not break anything but I request you to verify the same

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -66,8 +66,7 @@ impl Affine {
     /// The angle, `th`, is expressed in radians.
     #[inline]
     pub fn rotate(th: f64) -> Affine {
-        let s = th.sin();
-        let c = th.cos();
+        let (s, c) = th.sin_cos();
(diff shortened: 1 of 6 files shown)


raphlinus (approved):
Looks good to me, thanks!

derekdreery (approved):
LGTM

I'll let someone else check before merging.

Issue #211, opened 2022-01-13
The fact `BezPath` allows "subpaths" is poorly documented, annoying for implementors, and contrary to common sense


ctrlcctrlv:
Since @MatthewBlanchard recommended the gradual switchover of MFEK/math.rlib and MFEK/glifparser.rlib to Kurbo, I have been working on making both libraries more amenable for such a change.

Overall, Kurbo's primitives are better than ours. Except for one—`BezPath`/`PathEl`.

The title pretty much sums up my thoughts, but if it's not enough, please view this file: https://github.com/MFEK/glifparser.rlib/blob/ee7f0e8786cac7328f27064a5575aeb1a6175006/src/outline/kurbo.rs

That's what it takes to split Kurbo paths safely. That's absurd. Please begin guaranteeing `BezPath` won't contain more than one contour—use `Vec<BezPath>` for that. It's true that a lot of that file stems from our (MFEK's) `Point` type having a `prev`/`next` paradigm and not a segment-based paradigm (i.e. it represents points how users think of them, not how computers do), but a significant amount of the frustration—actually, most of it—was from this lack of separation.

Even now, for reasons I can't quite work out, I can't round trip open paths. I admitted defeat:

https://github.com/MFEK/glifparser.rlib/blob/ee7f0e8786cac7328f27064a5575aeb1a6175006/src/outline/kurbo.rs#L1

    /// Kurbo module — warning: only guaranteed to round trip closed contours!

cmyr:
Kurbo is intended for 2D drawing, and the vast  majority of 2D drawing APIs are based on the API of [PostScript](https://www.adobe.com/content/dam/acom/en/devnet/actionscript/articles/PLRM.pdf). This uses the pattern of modeling a Bèzier path as a series of operations with a pen. The common operations are,

- "move to": lift the pen, without drawing, and move it to a location (beginning a new subpath)
- "line to": draw a straight line from the pen's previous location to some new location
- "curve to": draw a cubic from the previous location to a new location
- "close path": draw a line to the first point in the current subpath (the "move to" location) and close the subpath

there are some variations: postscript also has "arc" which adds a circle segment.

[...]

PR #215, opened 2022-01-14
Improve documentation and robustness of BezPath

raphlinus:
Clearly document the correctness invariants of BezPath and the fact that
it can represent multiple subpaths. Add debug asserts to catch construction
of invalid paths.

Note: the test_elements_to_segments_starts_on_quad formerly constructed
an invalid path and expected particular behavior. This is changed to
expect a panic on debug assert.

Responsive to #211

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -202,11 +202,7 @@ impl BezPath {
     /// Will panic with a debug assert when the current subpath does not
     /// start with `move_to`.
     pub fn line_to<P: Into<Point>>(&mut self, p: P) {
-        debug_assert!(!self.0.is_empty(), "Subpath must start with move_to");
-        debug_assert!(
-            self.0.last() != Some(&PathEl::ClosePath),
-            "Subpath must start with move_to"
-        );
+        debug_assert!(self.is_open_subpath(), "no open subpath (missing MoveTo)");
         self.push(PathEl::LineTo(p.into()));
     }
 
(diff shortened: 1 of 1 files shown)


src/bezpath.rs
```diff
@@ -1245,7 +1269,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Can't start a segment on a ClosePath")]
+    #[should_panic(expected = "no open subpath")]
```
cmyr, line 1272: These tests will fail with `cargo test --release`. Not sure what the best approach is, here?

cmyr (approved):
one one inline, but I agree this is helpful, and the current behaviour was clearly wrong in some cases.

One additional thought: there are various other ways to construct a bezpath, such as by pushing elements directly, or via `FromIterator`. Should we validate those as well?

src/bezpath.rs
raphlinus: > One additional thought: there are various other ways to construct a bezpath, such as by pushing elements directly, or via `FromIterator`. Should we validate those as well?

It's a good question. Probably yes, but in some cases (like extending from a slice) it might be expensive. I think the current state is very clear documentation about what's correct, which is the immediate problem.

raphlinus: Regarding `test --release` I'm not sure what's considered best practice. I would prefer the checks not be in release code though, as it might be a performance critical path.

cmyr: yes, I agree with not having the tests in release mode. Can we disable these tests in release?

PR #216, opened 2022-01-14
BezPath: get_seg() && path_segments() equivalence

### LINKED ISSUE #214: [`BezPath::get_seg`] should not produce segment for `ClosePath` element if current position is equal to previous `MoveTo`
I already can't use it in glifparser due to #211, but this issue is yet another blocker.

https://github.com/linebender/kurbo/blob/b9247ed0a872cdd63a5b3c64ca5bc1e1392abee8/src/bezpath.rs#L322-L325

There is truly no line there on e.g. a circle. This is just attempting to work around the rather clumsy PostScript-ism of even having "move" points in non-move shapes (closed shapes).

cmyr:
This ensures these two methods have equivalent behaviour, and fixes up
the docs on the former.

closes #214

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -302,8 +302,14 @@ impl BezPath {
 
     /// Get the segment at the given element index.
     ///
-    /// The element index counts [`PathEl`] elements, so
-    /// for example includes an initial `Moveto`.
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
This looks right to me, thanks!

PR #222, opened 2022-07-12
Compute inflection points of cubic bezier

raphlinus:
It's useful to know where the inflection points are, possibly just to visualize, possibly because some other algorithm might depend on segments being convex.

Note that quadratic beziers have no inflection points.

Algorithm (cited) adapted from caffeineowl, but with I think a particularly nice higher-level expression.


cmyr (approved):
Looks good!

PR #225, opened 2022-08-24
More cubic arclen

raphlinus:


raphlinus:
I'm making this into a PR so we don't need to keep the branch around, and I'll make the relevant blog post point to this PR. The general approach is valid and I'll do another PR to improve arclen further, but I will close this PR.

PR #226, opened 2022-08-24
Switch to better cubic arclength algorithm

raphlinus:
This is the technique in https://raphlinus.github.io/curves/graphics/2019/01/04/followups.html, it just took me a while to land it.

Benchmarks show 3-4x improvement (both Intel and ARM M1) and it passes the accuracy tests. Another benefit is that it makes the Gauss-Legendre quadrature coefficient sets more complete, and I anticipate other uses of that soonish.


cmyr (approved):
A very cursory review from me, but the results are very satisfying :)

PR #228, opened 2022-08-29
change Line::is_finite() to involve both points

### LINKED ISSUE #227: Line::is_finite() does not consider both points
For a line where p0 is finite and p1 is not finite, Line::is_finite() returns `true`. Seems like it should return `false`.

mlwilkerson:
closes #227


raphlinus (approved):
Thanks, this was in fact an oversight.

cmyr:
Thanks!

PR #231, opened 2022-11-04
Implement GATs

derekdreery:
The main gain from this is being able to iterate over `BezPath`s without cloning. Makes MSRV 1.65

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -1108,10 +1108,10 @@ impl From<QuadBez> for PathSeg {
 }
 
 impl Shape for BezPath {
-    type PathElementsIter = std::vec::IntoIter<PathEl>;
+    type PathElementsIter<'iter> = std::iter::Copied<std::slice::Iter<'iter, PathEl>>;
 
-    fn path_elements(&self, _tolerance: f64) -> Self::PathElementsIter {
-        self.0.clone().into_iter()
+    fn path_elements(&self, _tolerance: f64) -> Self::PathElementsIter<'_> {
+        self.0.iter().copied()
     }
 
     fn to_path(&self, _tolerance: f64) -> BezPath {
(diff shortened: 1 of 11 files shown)


src/bezpath.rs
```diff
@@ -1108,10 +1108,10 @@ impl From<QuadBez> for PathSeg {
 }
 
 impl Shape for BezPath {
-    type PathElementsIter = std::vec::IntoIter<PathEl>;
+    type PathElementsIter<'iter> = std::iter::Copied<std::slice::Iter<'iter, PathEl>>;
 
-    fn path_elements(&self, _tolerance: f64) -> Self::PathElementsIter {
-        self.0.clone().into_iter()
```
derekdreery, line 1114: This is the main benefit from this PR - now we don't have to clone the whole object.

Zarenor (approved):
Excited to see this move, now that GATs are stable!

src/bezpath.rs
```diff
@@ -1108,10 +1108,10 @@ impl From<QuadBez> for PathSeg {
...
+    fn path_elements(&self, _tolerance: f64) -> Self::PathElementsIter<'_> {
+        self.0.iter().copied()
```
cmyr, line 1114: so the question I have with all of these is basically: do we even bother with `copy`? We we need to?

cmyr (approved):
One QQ, but otherwise very neat to see how simple a change this is!

src/bezpath.rs
derekdreery: Do you mean "why do we copy the elements rather than returning a ref to them? If so then the answer is because some impls of `path_elements` will create the path elements on-demand, and so need to pass ownership of them to the callee.

If I've misunderstood let me know.

cmyr: Right, that answers my question. LGTM!

derekdreery: I think it's an important question. From a rough calculation, the size of `PathEl` is around 50 bytes - not insignificant! 

Calc: `Point = 2 x f64 = 16 bytes, PathEl = 3 x Point + u8 (discriminant) = 49 ~ 50`.

derekdreery:
They can do that. I think I might experiment when I have time and get some
numbers.

src/bezpath.rs
dfrg: Measurement is important but I strongly suspect that the optimizer will eliminate these copies.

derekdreery:
I've investigated doing benchmarks with criterion.rs, and I think it's really great! But it's quite involved setting it up, so I will merge this PR for now.

PR #232, opened 2022-11-23
Bump version for release

dfrg:
Bump version to 0.9.0 to prepare for publishing.

This also adds a `Shape` impl for `[PathEl; N]` to match the one for `&[PathEl]`.


raphlinus (approved):
Looks good, and const generics FTW.

A couple things I noticed looking at the readme. One is that the CI badge is broken, as it still points to Travis. I'm sure making it point to GitHub Actions is a small change.

Also, it might be friendly to document the MSRV in the readme, as it will affect people on older versions (pre-GAT I don't think we were aggressively using newer features).

Lastly, now wouldn't be a bad time to bump the edition, though it's a pretty low priority thing.

dfrg:
Changed the badge to GH actions (and renamed the workflow to CI), added a note about MSRV to the readme and bumped the edition to 2021.

PR #233, opened 2023-01-23
Satisfy clippy up to 1.67 beta.

xStrom:
Clippy lints addressed:
* [`unnecessary_cast`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast)
* [`uninlined_format_args`](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args)
* [`manual_clamp`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_clamp)

The `clamp` one is a bit tricky. It introduces a panic condition to the `Size::clamp` method, which is never fun. Then again the panic case with the old implementation wouldn't satisfy the clamp conditions anyway, so perhaps the panic is justified. I ended up using the `std` `clamp` here, but I'm open to adding a clippy ignore attribute instead if the panic is too much.

### DIFF
--- a/src/size.rs
+++ b/src/size.rs
@@ -69,6 +69,11 @@ impl Size {
 
     /// Returns a new size bounded by `min` and `max.`
     ///
+    /// # Panics
+    ///
+    /// Panics if `min.width` > `max.width` or `min.height` > `max.height`,
+    /// or if any of those four are `NaN`.
+    ///
     /// # Examples
     ///
     /// ```
@@ -80,8 +85,8 @@ impl Size {
     /// assert_eq!(this.clamp(min, max), Size::new(10., 50.))
     /// ```
     pub fn clamp(self, min: Size, max: Size) -> Self {
-        let width = self.width.max(min.width).min(max.width);
-        let height = self.height.max(min.height).min(max.height);
+        let width = self.width.clamp(min.width, max.width);
+        let height = self.height.clamp(min.height, max.height);
         Size { width, height }
     }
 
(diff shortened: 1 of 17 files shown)


src/size.rs
```diff
@@ -69,6 +69,11 @@ impl Size {
 
     /// Returns a new size bounded by `min` and `max.`
     ///
+    /// # Panics
```
cmyr, line 72: I think it's fine if we follow std semantics for panic, and as you say we can revisit if this ends up being a problem. cc @raphlinus in case he has any opinions.

cmyr (approved):
Thanks, I'll let @raphlinus chime in on the clamp suggestion but I would do the same as you have here.

cmyr:
@xStrom did a bunch of these lints get demoted before release? Just noticing this isn't merged, but I don't see these on 1.68

xStrom:
Not before release, no. `1.67.0-x86_64-pc-windows-msvc`, which was stable, still triggers these clippy warnings.

However `1.67.1-x86_64-pc-windows-msvc` doesn't. So it seems these were demoted in a bugfix release. 🤔

The [release notes for 1.67.1](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1671-2023-02-09) only mention [`uninlined_format_args`](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args). I also remember seeing some talk of demoting [`manual_clamp`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_clamp) due to behavior change in the clamp PR, but seems this was done silently. Unclear what happend to [`unnecessary_cast`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast).

I don't have strong feelings about any of these right now. We could merge as-is, sans the clamp, or not at all. Thoughts?

raphlinus:
I'm ok with the clamp one (and don't have any strong feelings about the others). This is technically a semver-breaking change, but we've always erred on the side of being too strict, so I think it's consistent with our philosophy.

PR #234, opened 2023-02-22
Add scaling traits to Point

samlich:


cmyr:
Thanks!

There was an intentional decision to not include these for `Point`, and only for `Vec2`. This may feel somewhat arbitrary, but the basic idea is that conceptually a point (unlike a vector) does not have a magnitude. If you wish to scale a point, you can always perform these operations by going through `Vec2`, e.g. `(my_point.to_vec2() * 2.0).to_point()`. 

I do think it could be reasonable to have an explicit `scale` method on `Point`, as a convenience, but I am not inclined to add the arithmetic operators.

PR #237, opened 2023-03-02
More readable SVG output

dfrg:
Changes the SVG writer to emit commas between coordinate components and spaces between coordinates and commands.

Requested by @rsheeter in this [comment](https://github.com/googlefonts/fontations/pull/237#discussion_r1110346081).

### DIFF
--- a/src/svg.rs
+++ b/src/svg.rs
@@ -548,7 +548,10 @@ mod tests {
         ];
         let path = BezPath::from_path_segments(segments.iter().cloned());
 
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
The extra space byte pains me slightly, but the previous state wasn't an optimized packer (it could do relative coords, special cases, limit precision, etc).

dfrg:
Might be worth adding a `to_svg_packed` method in the future that does all of these optimizations.

PR #238, opened 2023-03-02
Fix clippy

rsheeter:
Didn't want to fix unrelated clippy noise in #236


dfrg (approved):
Ran into the same for my PR.

PR #240, opened 2023-03-03
Add Affine::skew

dfrg:
Adds a `skew` constructor to `Affine` taking skew factors in the horizontal and vertical direction.

The need for this has come up in both Lottie animation and font rendering (faux italic/oblique).

Could be separated into `skew` and `skew_non_uniform` in the spirit of `scale` but uniform skews tend to be far less common in practice.

(creating a new PR because CI appears to be broken on the previous one)

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -77,6 +77,24 @@ impl Affine {
         Affine([1.0, 0.0, 0.0, 1.0, p.x, p.y])
     }
 
+    /// An affine transformation representing a skew.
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
Looks good, and thanks for being careful in the doc comments with the sign conventions!

dfrg:
Kurbo docs are generally very high quality and I didn’t want to be the one to break the trend :)

PR #242, opened 2023-03-04
feat: Add a feature that allows usage on no_std targets

notgull:
This PR adds features `std` (enabled by default) and `libm` that allow this crate to be used on `no_std` targets.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,8 @@ jobs:
           profile: minimal
           override: true
 
+      - run: rustup target add thumbv7m-none-eabi
+
       - name: cargo clippy
         uses: actions-rs/cargo@v1
         with:
@@ -58,6 +60,10 @@ jobs:
         with:
           command: test
           args: --all-features
+
+      - name: Build with no default features
+        # Use no-std target to ensure we don't link to std.
+        run: cargo build --no-default-features --features libm --target thumbv7m-none-eabi
 
   test-stable-wasm:
     runs-on: ${{ matrix.os }}
--- a/examples/circle.rs
+++ b/examples/circle.rs
@@ -1,21 +1,24 @@
 //! Example of circle
 
+#[cfg(feature = "std")]
 fn main() {
-    #[cfg(feature = "std")]
-    {
-        use kurbo::{Circle, Shape};
+    use kurbo::{Circle, Shape};
 
-        let circle = Circle::new((400.0, 400.0), 380.0);
-        println!("<!DOCTYPE html>");
-        println!("<html>");
-        println!("<body>");
-        println!("<svg height=\"800\" width=\"800\">");
-        let path = circle.to_path(1e-3).to_svg();
-        println!("  <path d=\"{}\" stroke=\"black\" fill=\"none\" />", path);
-        let path = circle.to_path(1.0).to_svg();
-        println!("  <path d=\"{}\" stroke=\"red\" fill=\"none\" />", path);
-        println!("</svg>");
-        println!("</body>");
-        println!("</html>");
-    }
+    let circle = Circle::new((400.0, 400.0), 380.0);
+    println!("<!DOCTYPE html>");
+    println!("<html>");
+    println!("<body>");
+    println!("<svg height=\"800\" width=\"800\">");
+    let path = circle.to_path(1e-3).to_svg();
+    println!("  <path d=\"{}\" stroke=\"black\" fill=\"none\" />", path);
+    let path = circle.to_path(1.0).to_svg();
+    println!("  <path d=\"{}\" stroke=\"red\" fill=\"none\" />", path);
+    println!("</svg>");
+    println!("</body>");
+    println!("</html>");
+}
+
+#[cfg(not(feature = "std"))]
+fn main() {
+    println!("This example requires the standard library");
 }
--- a/examples/circle.rs
+++ b/examples/circle.rs
@@ -1,18 +1,21 @@
 //! Example of circle
 
-use kurbo::{Circle, Shape};
-
 fn main() {
-    let circle = Circle::new((400.0, 400.0), 380.0);
-    println!("<!DOCTYPE html>");
-    println!("<html>");
-    println!("<body>");
-    println!("<svg height=\"800\" width=\"800\">");
-    let path = circle.to_path(1e-3).to_svg();
-    println!("  <path d=\"{path}\" stroke=\"black\" fill=\"none\" />");
-    let path = circle.to_path(1.0).to_svg();
-    println!("  <path d=\"{path}\" stroke=\"red\" fill=\"none\" />");
-    println!("</svg>");
-    println!("</body>");
-    println!("</html>");
+    #[cfg(feature = "std")]
+    {
+        use kurbo::{Circle, Shape};
+
+        let circle = Circle::new((400.0, 400.0), 380.0);
+        println!("<!DOCTYPE html>");
+        println!("<html>");
+        println!("<body>");
+        println!("<svg height=\"800\" width=\"800\">");
+        let path = circle.to_path(1e-3).to_svg();
+        println!("  <path d=\"{}\" stroke=\"black\" fill=\"none\" />", path);
+        let path = circle.to_path(1.0).to_svg();
+        println!("  <path d=\"{}\" stroke=\"red\" fill=\"none\" />", path);
+        println!("</svg>");
+        println!("</body>");
+        println!("</html>");
+    }
 }
--- a/examples/ellipse.rs
+++ b/examples/ellipse.rs
@@ -1,22 +1,25 @@
 //! Example of ellipse
 
-use kurbo::{Ellipse, Shape};
-use std::f64::consts::PI;
-
+#[cfg(feature = "std")]
 fn main() {
-    #[cfg(feature = "std")]
-    {
-        let ellipse = Ellipse::new((400.0, 400.0), (200.0, 100.0), 0.25 * PI);
-        println!("<!DOCTYPE html>");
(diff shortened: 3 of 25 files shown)


examples/circle.rs
```diff
@@ -1,18 +1,21 @@
...
-    println!("<body>");
-    println!("<svg height=\"800\" width=\"800\">");
-    let path = circle.to_path(1e-3).to_svg();
-    println!("  <path d=\"{}\" stroke=\"black\" fill=\"none\" />", path);
-    let path = circle.to_path(1.0).to_svg();
-    println!("  <path d=\"{}\" stroke=\"red\" fill=\"none\" />", path);
-    println!("</svg>");
-    println!("</body>");
-    println!("</html>");
+    #[cfg(feature = "std")]
```
cmyr, line 4: Can we do something like just adding `#![cfg(feature = "std")]` to the top of the file? (Where possible, I prefer solutions that minimize diff size)

examples/ellipse.rs
```diff
@@ -4,16 +4,19 @@ use kurbo::{Ellipse, Shape};
...
-    println!("<body>");
-    println!("<svg height=\"800\" width=\"800\" style=\"background-color: #999\">");
-    let path = ellipse.to_path(1e-3).to_svg();
-    println!("  <path d=\"{}\" stroke=\"black\" fill=\"none\" />", path);
-    let path = ellipse.to_path(1.0).to_svg();
-    println!("  <path d=\"{}\" stroke=\"red\" fill=\"none\" />", path);
-    println!("</svg>");
-    println!("</body>");
-    println!("</html>");
+    #[cfg(feature = "std")]
```
cmyr, line 7: ditto

.github/workflows/ci.yml
```diff
@@ -47,6 +47,8 @@ jobs:
           profile: minimal
           override: true
 
+      - run: rustup target add thumbv7m-none-eabi
```
cmyr, line 50: do we need to test against this target? I would be happy enough to have CI just check that no-std builds, against a default target.

src/bezpath.rs
```diff
@@ -15,6 +17,9 @@ use crate::{
     ParamCurveExtrema, ParamCurveNearest, Point, QuadBez, Rect, Shape, TranslateScale, Vec2,
 };
 
+#[allow(unused_imports)]
+use crate::common::FloatFuncs;
```
cmyr, line 21: If this is only needed for a feature, I would prefer it to just be `cfg`'d on that feature, instead of just always allowed.

src/common.rs
```diff
@@ -4,6 +4,91 @@
...
+        fn $name:ident(self $(,$arg:ident: $arg_ty:ty)*) -> $ret:ty
+        => $lname:ident/$lfname:ident;
+    )+) => {
+        pub(crate) trait FloatFuncs : Sized {
+            $(fn $name(self $(,$arg: $arg_ty)*) -> $ret;)+
+        }
+
+        impl FloatFuncs for f32 {
+            $(fn $name(self $(,$arg: $arg_ty)*) -> $ret {
+                #[cfg(feature = "std")]
```
cmyr, line 19: slightly pedantic, but if we *do* want to use this trait in both cases (which would require us to call the trait methods directly, or give them names that do not shadow the std names) I would do this the other way around: if `libm` is set, always use libm, if `libm` is not set and `std` is set use `std`. (Because perhaps for some reason you want to use musl even if std is available? Tell me if I'm wrong 💁)

src/common.rs
```diff
@@ -4,6 +4,91 @@
...
+    ($(
+        fn $name:ident(self $(,$arg:ident: $arg_ty:ty)*) -> $ret:ty
+        => $lname:ident/$lfname:ident;
+    )+) => {
+        pub(crate) trait FloatFuncs : Sized {
+            $(fn $name(self $(,$arg: $arg_ty)*) -> $ret;)+
+        }
+
+        impl FloatFuncs for f32 {
+            $(fn $name(self $(,$arg: $arg_ty)*) -> $ret {
```
cmyr, line 18: Do we want an `#[inline]` here? Not clear that this crosses crate boundaries (I guess not?) so maybe it doesn't matter..

src/common.rs
```diff
@@ -4,6 +4,91 @@
 
 use arrayvec::ArrayVec;
 
+/// Defines a trait that chooses between libstd or libm implementations of float methods.
```
cmyr, line 7: so... is this actually necessary in the case of std? If std is present then these methods will already exist on our types, and method resolution will work, so it feels like we could get away with only defining this trait if the `musl` feature is active?

(the fact that this trait does nothing if `std` is active is supported by the need to add `#[allow(unused_code)]` on the import)

src/cubicbez.rs
```diff
@@ -15,6 +17,9 @@ use crate::{
     ParamCurveDeriv, ParamCurveExtrema, ParamCurveNearest, PathEl, Point, QuadBez, Rect, Shape,
 };
 
+#[allow(unused_imports)]
```
cmyr, line 20:
```suggestion
#[cfg(feature = "libm")]
```

src/mindist.rs
```diff
@@ -21,6 +21,9 @@
 use crate::Vec2;
 use core::cmp::Ordering;
 
+#[allow(unused_imports)]
```
cmyr, line 24: #[cfg(feature = "libm")]

src/param_curve.rs
```diff
@@ -1,11 +1,14 @@
 //! A trait for curves parametrized by a scalar.
 
-use std::ops::Range;
+use core::ops::Range;
 
 use arrayvec::ArrayVec;
 
 use crate::{common, Point, Rect};
 
+#[allow(unused_imports)]
```
cmyr, line 9: ditto

src/quadbez.rs
```diff
@@ -12,6 +12,9 @@ use crate::{
     Rect, Shape,
 };
 
+#[allow(unused_imports)]
```
cmyr, line 15: ditto

cmyr (commented):
Okay, so: I definitely think we would like to work in no-std environments; this is something we've talked about, but never been motivated to actually approach.

That said, I'm not sure if we want to assume the solution is always musl? I really don't know what our options are, though, so I'd be curious if @dfrg or @raphlinus have any opinions.

I've reviewed this as-is; my main issue currently is that the trait is sort of half-baked, since we define it in both cases but it's only ever used if std is missing; if this is the approach we take I think we should simplify this trait to only exist if std is not active.

.github/workflows/ci.yml
notgull: `no_std` crates can accidentally link to std by accident, in some weird cases. :face_with_spiral_eyes: To catch these errors, you can build on a target that doesn't have a libstd.

notgull:
Thanks for the review!

> That said, I'm not sure if we want to assume the solution is always musl?

Just to clarify, the `libm` crate doesn't link to musl's `libm`. It's actually a from-scratch implementation of math routines in pure Rust. However, since it's pure Rust (read: not hardware-accelerated), I prefer using hardware intrinsics when available. That's why I prefer `std` over `libm` in this patch, since `libm` tends to be significantly slower.

> I've reviewed this as-is; my main issue currently is that the trait is sort of half-baked, since we define it in both cases but it's only ever used if std is missing

Good point! I'll go back and fix it like this when I have the chance.

.github/workflows/ci.yml
cmyr: okay cool, I hadn't heard about that trick before. :)

notgull:
@cmyr What is the status of reviewing this PR?

cmyr:
> Just to clarify, the `libm` crate doesn't link to musl's `libm`. It's actually a from-scratch implementation of math routines in pure Rust. However, since it's pure Rust (read: not hardware-accelerated), I prefer using hardware intrinsics when available. That's why I prefer `std` over `libm` in this patch, since `libm` tends to be significantly slower.

Okay cool, I hadn't understood that on my first pass.

This can get rebased to bring in the more recent clippy fixes, and then I'm going to solicit a second opinion; modulo that I'm happy to merge!

dfrg:
I’m so conflicted on this one. I’d really like to see no-std support in kurbo and this PR does what essentially amounts to the best that can be done given the constraints.

My primary concern is that pulling in a dependency that transitively references kurbo without std will still make use of the suboptimal math functions even if std exists in the full dependency graph. There is an issue in libm that suggests making all symbols [weak](https://github.com/rust-lang/libm/issues/257) which would solve this problem but that hasn’t been done yet. 

That said, since we can override this by adding an explicit kurbo dep with std, I’m also inclined to suggest merging.

+1

notgull:
I've rebased on the latest `master`.

cmyr (approved):
I think chad makes an interesting point, but there is a clear workaround available if this problem crops up, so I think there is clear value in landing this.

Thanks @notgull!

src/common.rs
```diff
@@ -4,6 +4,91 @@
...
+    fn round(self) -> Self => round/roundf;
+    fn sin_cos(self) -> (Self, Self) => sincos/sincosf;
+    fn sqrt(self) -> Self => sqrt/sqrtf;
+    fn tan(self) -> Self => tan/tanf;
+    fn trunc(self) -> Self => trunc/truncf;
+}
+
+/// Special implementation for signum, because libm doesn't have it.
+#[cfg(not(feature = "std"))]
+pub(crate) trait FloatFuncsExtra {
```
cmyr, line 64: oh one last note: could we remove this extra trait by just declaring the method in the `FloatFuncs` trait above, providing a default impl?

notgull: Easy enough, done!

Issue #178, opened 2021-04-21
Possibility of using the `Float` trait instead of `f64`?


ctrlcctrlv:
This relates to linebender/norad#108.

The `num_traits` crate declares a trait called `Float` which implements `Neg`, `Add`, `Mul`, `Rem`, `Div`, etc etc. It then implements this trait for `f32` and `f64`.

```rust
pub trait Float: Num + Copy + NumCast + PartialOrd + Neg<Output = Self> {...}
```

Many libraries which work on numbers use this crate to be generic over floating point types.

I wonder if Kurbo could too?

Edited to add: I think this issue is fundamentally different from, but similar to, #159, because it basically lets the API consumer decide whether or not they are willing to take the risks of `f32` lower precision.

raphlinus:
My response is basically the same as #159. Generics make the types messier and potentially compile times longer and code more bloated (if there are two versions of the same algorithm). I think when interoperating with other code that prefers f32 (as is the case for Direct2D for example), putting in explicit precision-losing conversions is acceptable.

I do see that f64 is more precision than needed for font coordinates. There are other cases, though (scrolling is one) where the lower precision causes problems.

If I saw evidence of a real performance difference (as in #159), or evidence that ergonomics would be substantially improved, I would be more inclined to consider it. But in the absence of solid evidence, I am not really in favor.

cmyr:
I remember we did some preliminary benchmarking a few years ago and found there was very little performance advantage to using  f64, at least on x86_64. There would of course be some binary size advantages.

From where I stand right now, I have a good sense of the downsides of this change, but I don't really have a good sense of the upsides: I mean this in an honest way, I just haven't personally encountered any situations where I've really wished we were using `f32`, although I don't doubt that they exist.

my main concern with this is really the messier types.

There's a related argument that kurbo types should include an explicit generic parameter to represent an associated coordinate space; in this way you might have `Point<PixelSpace>` and `Point<DesingSpace>`, and you would have to explicitly move between them;  this is a common thing to want in, say, a graphics editor. So far, though, this has also not felt worth the hassle.

Just to try and think this through, though, if we *were* going to do one or both of these, what I might imagine is having separate 'Raw' types that have the generic params, and then export simple types that have those params filled in, like:

```rust
pub struct RawPoint<T, R> {
    _space: PhantomData<R>,
    pub x: T,
    pub y: T,
}

pub trait CoordinateSpace: Copy + Eq {}

// everything is impl'd in terms of these params:
[...]

PR #248, opened 2023-03-23
Add epsilon API

ctrlcctrlv:
Didn't think there was much point in fixing the epsila until the API is even approved. (See #178.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,6 +69,7 @@
 //! [`Druid`]: https://docs.rs/druid
 //! [`libm`]: https://docs.rs/libm
 
+#![feature(const_fn_floating_point_arithmetic, const_trait_impl, const_ops)]
 #![forbid(unsafe_code)]
 #![deny(missing_docs, clippy::trivially_copy_pass_by_ref)]
 #![warn(rustdoc::broken_intra_doc_links)]
@@ -92,6 +93,7 @@ mod circle;
 pub mod common;
 mod cubicbez;
 mod ellipse;
+mod epsilon;
 mod insets;
 mod line;
 mod mindist;
@@ -115,6 +117,7 @@ pub use crate::bezpath::*;
 pub use crate::circle::*;
 pub use crate::cubicbez::*;
 pub use crate::ellipse::*;
+pub use crate::epsilon::*;
 pub use crate::insets::*;
 pub use crate::line::*;
 pub use crate::param_curve::*;
--- a/src/common.rs
+++ b/src/common.rs
@@ -5,6 +5,8 @@
 
 #![allow(missing_docs)]
 
+use crate::epsilon::Epsilon;
+
 use arrayvec::ArrayVec;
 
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -69,6 +69,7 @@
 //! [`Druid`]: https://docs.rs/druid
 //! [`libm`]: https://docs.rs/libm
 
+#![feature(const_fn_floating_point_arithmetic, const_trait_impl, const_ops)]
```
ctrlcctrlv, line 72: note: i can remove these and make the functions non-const but someone here may have better idea

raphlinus:
We discussed this in office hours today. Generally, seeing this and the back-and-forth in #178, I'm inclined to take this as confirmation that just making everything `f64` is the right way to go. Making kurbo generic over precision is *not* a simple task, and in particular some of the algorithms have very careful numerical robustness work that has to be re-thought. In particular, the work in #230 relies on quartic solving which is very delicate with respect to precision. Making that numerically robust is likely possible (though I'm not sure of that), but would probably be a major effort.

There are use cases where f32 would be very desirable. The biggest is of course if it's running on a GPU (rust-gpu is on my radar). Another is if we're starting to do serious SIMD optimization. However, the costs are nontrivial.

I'd like to understand better exactly what the end goals are. Some goals are probably best met by having all the core algorithms run in f64 but perhaps having conversions to f32 at the edges (this is more or less what we do for Vello scene encoding). Other goals may be met by having, say, an f32 variant of BezPath that implements the `Shape` trait (we considered this for Vello encoding but ultimately decided against).

Regarding epsilons, we've had discussion on this point before. There's no "one size fits all" approach, and the choice of epsilon depends on the use case. For interactive applications, it's probably best to use the largest value that doesn't cause visual problems, to get the fastest performance. For other applications, it's important to get the right answer. The choice of epsilon can also depend on the range of scaling transforms that are expected and the coordinate system for the final result. Thus, we try not to use "magic constant" epsilon values, and instead pass in an accuracy parameter. It's appropriate to use concrete values in tests (and in general those should be small), as the goal is to detect anything that might cause loss of precision.

Long story short, I am against making a 32 bit version of kurbo unless I see a clear motivation, and until then I am inclined not to accept any changes to facilitate this.

ctrlcctrlv:
> There are use cases where f32 would be very desirable. The biggest is of course if it's running on a GPU (rust-gpu is on my radar). Another is if we're starting to do serious SIMD optimization. However, the costs are nontrivial.

My code does run on GPU as Skia uses the GPU.

ctrlcctrlv:
> Thus, we try not to use "magic constant" epsilon values, and instead pass in an accuracy parameter. It's appropriate to use concrete values in tests (and in general those should be small), as the goal is to detect anything that might cause loss of precision.

Primarily that's what this API is for, the tests, as that's where most (all?) of the epsila in the codebase are defined.

raphlinus:
> My code does run on GPU as Skia uses the GPU.

I specifically meant running in shaders or CUDA.

PR #254, opened 2023-03-29
Expose and fix a cubic to quadratic error

rsheeter:
Note that `CubicBez::from_parameters` was computing the final point differently than the original FontTools implementation. Pre-fix kurbo couldn't solve the example cubic to quadratic at an accuracy < 1.03 while the original cu2qu merrily solved with accuracy 0.001.

Add more attribution to make it easier to track down any future issues.

Once released https://github.com/googlefonts/fontmake-rs/issues/217 fontmake-rs can be resolved by updating kurbo.

### DIFF
--- a/src/cubicbez.rs
+++ b/src/cubicbez.rs
@@ -256,13 +256,13 @@ impl CubicBez {
         (a, b, c, d)
     }
 
+    /// Rust port of cu2qu [calc_cubic_points](https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/cu2qu/cu2qu.py#L63-L68)
     fn from_parameters(a: Vec2, b: Vec2, c: Vec2, d: Vec2) -> Self {
-        CubicBez::new(
-            d.to_point(),
-            d.to_point() + (c / 3.0),
-            d.to_point() + (c / 3.0) + (b + c) / 3.0,
-            d.to_point() + (c / 3.0) + (b + c) / 3.0 + a,
-        )
+        let p0 = d.to_point();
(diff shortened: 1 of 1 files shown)


src/cubicbez.rs
```diff
@@ -256,13 +256,13 @@ impl CubicBez {
         (a, b, c, d)
     }
 
+    /// Rust port of cu2qu [calc_cubic_points](https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/cu2qu/cu2qu.py#L63-L68)
     fn from_parameters(a: Vec2, b: Vec2, c: Vec2, d: Vec2) -> Self {
-        CubicBez::new(
-            d.to_point(),
-            d.to_point() + (c / 3.0),
-            d.to_point() + (c / 3.0) + (b + c) / 3.0,
-            d.to_point() + (c / 3.0) + (b + c) / 3.0 + a,
```
anthrotype, line 264: good catch! it was only this line that computed the CubicBez end point `p3` that was at fault (the other three points were correct, you've simply rewritten to match fonttools' original code more literally, which is good). p3 should have been the sum of the four parameters a + b + c + d instead.

anthrotype (approved):

dfrg (approved):

src/cubicbez.rs
rsheeter: Correct, I attempted to highlight that in the PR description but perhaps I should have stuck a comment here too.

anthrotype: > "was computing the final point differently"

Oh I now reread that and makes sense, thank you

PR #256, opened 2023-03-30
This PR adds and makes public some helper methods on `Affine`

derekdreery:
I made this PR because `euclid` has these methods and I find them useful. But they are absolutely for discussion. If we decide we want some of them but not all, I'm happy to remove some.


raphlinus (approved):
These seem reasonable to me, and more convergence with `euclid` is also a good thing. Thanks!

PR #258, opened 2023-04-01
Cubic intersection experimental branch

raphlinus:
A prototype of a cubic/cubic intersection algorithm. The core of this produces better results than the "fat line" technique (see #199), but it's not made fully robust and is not wired up with a usable API. This draft PR just snapshots work in progress and hopefully is a step toward a real solution.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,7 @@ mod insets;
 mod line;
 mod mindist;
 mod param_curve;
+pub mod pathops;
 mod point;
 mod quadbez;
 mod quadspline;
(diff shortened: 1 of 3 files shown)


raphlinus:
I realize I had pushed the wrong branch. This is still quite messy (and should probably be replaced with a cleaner commit), but in any case, the interesting logic is in `examples/cubic_fun.rs`.

simoncozens:
I recently discovered there is actually an algebraic closed-form solution for cubic-cubic intersection! "All" you have to do is construct a Sturm sequence and solve a 9th-degree polynomial. Easy, really. There's an implementation [here](https://github.com/thelonious/kld-intersections/blob/0f0aecea6923c0042c72f514fa914cac95a72686/lib/Intersection.js#L621-L709).

PR #266, opened 2023-04-12
Add alloc feature to serde dependency

dfrg:
Restores serde support for BezPath.

Also bumps version so we can release.


dfrg:
For context, #242 turned off default_features for the serde dependency which disabled support for serializing collections. This means that kurbo now fails to compile with the serde feature enabled because it can't derive `Serialize`/`Deserialize` for `BezPath` which contains a `Vec`.

PR #267, opened 2023-04-12
[ci] Test serde feature in CI

cmyr:
Curious to see how much this impacts our CI runtime. Also just realized we're running CI on three platforms, which may not be necessary? I don't think we have any platform-specific code...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,9 @@ jobs:
         # Use no-std target to ensure we don't link to std.
         run: cargo build --no-default-features --features libm --target thumbv7m-none-eabi
 
+      - name: cargo test --features=serde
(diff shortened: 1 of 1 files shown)


dfrg (approved):
Looks good. Agreed that building on multiple platforms is probably unnecessary.

xStrom:
Kurbo has dependencies that might introduce platform specific code. Maybe they already have some, I haven't checked. For that reason I think that testing multiple platforms has value.

What are the downsides of doing such testing? I know of a few.
* Job runtime duration. However we run the jobs in parallel, so this should be very minor.
* Cache overflow. Can happen with a larger matrix of tests like Druid has, but this isn't an issue with Kurbo.

Are there are other downsides?

Issue #268, opened 2023-04-13
Robustness problems in path simplification


raphlinus:
This is essentially a continuation of #264 but with concrete failing test cases. Thanks to @rachael-wang for providing the test code and data. I've put the code in a [gist](https://gist.github.com/raphlinus/35d7e9acbb064ceac941313c681e4c4a); the easiest way to run it is to place it in the examples/ directory.

Running this command on the attached test data produces the following SVG (blue is input, red is output):

```
cargo run --example simplify_svg 219561.txt > simplify_result.svg
```

As discussed in #264, the problem is partly that `fit_to_bezpath_opt` is being called with path data containing corners, and it is intended to be used for reasonably smooth input paths. One solution is to implement an additional layer to segment the input by corners and optimize each smooth subrange separately.

Another issue, though, is that this test case triggers robustness issues. For example, changing the accuracy parameter from 0.18 to 0.3 in the code trips a [panic](https://github.com/linebender/kurbo/blob/615ed7ede2d80ae0f3509165c5a2dee6cef69251/src/common.rs#L456) corresponding to no real roots of the quartic equation. At the minimum, this code should be changed to report a lack of solution so it can be recovered, rather than panicking.

![simplify_result](https://user-images.githubusercontent.com/242367/231861300-04f7c44f-e8f3-4053-8475-8fc4f6b30c8b.svg)

[219561.txt](https://github.com/linebender/kurbo/files/11225752/219561.txt)

raphlinus:
In addition to the corner-handling robustness as mentioned above, there is another issue that deserves attention. In the looped Bézier segment toward the bottom of the figure, the error metric appears to be drastically underestimating the error between the source curve and approximation. I think I might know what's happening: in `fit_to_cubic` if a ray from the source curve fails to intersect the cubic at all, it returns an error larger than the provided accuracy parameter. In the `fit_to_bezpath` case that will always force a subdivision, but in the `_opt` variant it seems like there is at least one path where such a result is accepted without additional checking. That wouldn't happen if the error were monotonic in the subdivided range, but when the input curve is not smooth, that assumption is not valid.

raphlinus:
Above analysis is not correct, it is not the case that `fit_to_cubic` returns cubics that exceed the accuracy parameter provided to it. Here's an alternate explanation.

[...]

PR #269, opened 2023-04-14
Improve robustness of curve fitting

raphlinus:
This PR contains a number of improvements to the robustness and accuracy of curve fitting. A milestone is that seems to handle the test case of #268 fairly well. It also implements one aspect of #267 (handling the case where the quartic has no real roots), but does not implement the higher level driver proposed in that issue. Thus, it does its best to fit corners with smooth curves.

A list of the changes in somewhat more detail:

* The quartic solver simply reports when there are no real solutions, rather than panicking. A potential future expansion might be to compute complex coefficients, but that would require a dependency on complex math and does not seem to be needed here.
* The ITP solving for determining subdivision points chooses the conservative side of the bracket rather than the midpoint. This gives much better results when the function being solved has discontinuities. This also avoids problems with causing n to increase due to non-monotonicity of the curve fitting.
* The error metric for estimating the Fréchet distance between the source curve and the approximation has been revised substantially. Previously it was the ray projection method from Tiller and Hanson, but this can significantly underestimate error when the source curve has extreme variations of curvature. A particularly bad case (illustrated below) is when the source curve has a corner (or nearly so), in which case the cubic approximation might have a loop, and the error metric reports a low error because the loop is not sampled. Even so, the code is probably slower. There may be an error metric out there which is both faster than the current code and robust enough.
* The `fit_to_bezpath_opt` logic has been changed to make accuracy bounds tighter for the curve fitting steps; those are allowed to fail, and logic can deal with that. Previously it set a HUGE accuracy bound with the expectation that some solution would always be found. This should improve both robustness and performance (relative

The new method is based on arc length parametrized sampling of the approximation curve (the sample points are sampled evenly in parameter space in the source curve, which might possibly be improved, but otherwise would require solving inverse arc length problems there). While this error metric is more robust, avoiding the loops, it is also dramatically slower. Thus, this PR applies a hybrid approach, first estimating whether the source curve is "spicy" by determining whether the tangent of the delta between tangent vectors between samples exceeds 0.1 (the `SPICY_THRESH` constant in the code). If so, the arc length method is used (after first running the ray cast method).

Experimentation is encouraged, as the goal is truly robust curve fitting. The techniques deserve a proper writeup. One thing of note is that arc length is computed (as an integral) as *another* application of the `ParamCurveFit` trait; no changes were needed to the trait itself. It's also worth noting that errors in computing arc length make the error metric more conservative, so extreme care is not needed there.

Screenshot 2023-04-14 at 7 03 17 AM

### DIFF
--- a/src/fit.rs
+++ b/src/fit.rs
@@ -196,18 +196,36 @@ struct CurveDist {
     samples: ArrayVec<CurveFitSample, N_SAMPLE>,
     arcparams: ArrayVec<f64, N_SAMPLE>,
     range: Range<f64>,
+    /// A "spicy" curve is one with potentially extreme curvature variation,
+    /// so the more accurate measurement is.
+    spicy: bool,
 }
 
 impl CurveDist {
     fn from_curve(source: &impl ParamCurveFit, range: Range<f64>) -> Self {
         let step = (range.end - range.start) * (1.0 / (N_SAMPLE + 1) as f64);
+        let mut last_tan = None;
+        let mut spicy = false;
+        const SPICY_THRESH: f64 = 0.1;
         let samples = (0..N_SAMPLE)
-            .map(|i| source.sample_pt_tangent(range.start + (i + 1) as f64 * step, 1.0))
+            .map(|i| {
+                let sample = source.sample_pt_tangent(range.start + (i + 1) as f64 * step, 1.0);
+                if let Some(last_tan) = last_tan {
+                    let cross = sample.tangent.cross(last_tan);
+                    let dot = sample.tangent.dot(last_tan);
+                    if cross.abs() > SPICY_THRESH * dot.abs() {
+                        spicy = true;
+                    }
+                }
+                last_tan = Some(sample.tangent);
+                sample
+            })
             .collect();
         CurveDist {
             samples,
             arcparams: Default::default(),
             range,
+            spicy,
         }
     }
 
@@ -256,7 +274,7 @@ impl CurveDist {
     ///
     /// Returns the squre of the error, which is intended to be a good
     /// approximation of the Fréchet distance.
-    fn eval_dist(&self, c: CubicBez, acc2: f64) -> Option<f64> {
+    fn eval_ray(&self, c: CubicBez, acc2: f64) -> Option<f64> {
         let mut max_err2 = 0.0;
         for sample in &self.samples {
             let mut best = acc2 + 1.0;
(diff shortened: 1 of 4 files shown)


derekdreery:
I have a test case. The code below can be copy-pasted into src/offset.rs. I tested on HEAD of robust_fit and got a hang, I think during the cubic solver.

```rust
#[cfg(test)]
mod tests {
    use super::CubicOffset;
    use crate::{fit_to_bezpath_opt, CubicBez, PathEl};

    // This test tries combinations of parameters that have caused problems in the past.
    #[test]
    fn pathological_curves() {
        let curve = CubicBez {
            p0: (-1236.3746269978635, 152.17981429574826).into(),
            p1: (-1175.18662093517, 108.04721798590596).into(),
            p2: (-1152.142883879584, 105.76260301083356).into(),
            p3: (-1151.842639804639, 105.73040758939104).into(),
        };
        let offset = 3603.7267536453924;
        let accuracy = 0.1;
        let offset_path = CubicOffset::new(curve, offset);
        let path = fit_to_bezpath_opt(&offset_path, accuracy);
        assert!(matches!(path.iter().next(), Some(PathEl::MoveTo(_))));
    }
}
```

src/fit.rs
```diff
@@ -207,8 +322,12 @@ pub fn fit_to_cubic(
     let d = end.p - start.p;
     let th = d.atan2();
     let chord2 = d.hypot2();
-    let th0 = start.tangent.atan2() - th;
-    let th1 = th - end.tangent.atan2();
+    fn mod_2pi(th: f64) -> f64 {
```
derekdreery, line 325: could you use `f64::rem_euclid` for this?

src/fit.rs
```diff
@@ -238,46 +357,23 @@ pub fn fit_to_cubic(
...
-            }
-            let mut best = None;
-            for t in intersections {
-                let err = sample.p.distance_squared(c.eval(t));
-                if best.map(|best| err < best).unwrap_or(true) {
-                    best = Some(err);
-                }
+        if let Some(err2) = curve_dist.eval_dist(source, c, acc2) {
+            fn scale_f(d: f64) -> f64 {
+                1.0 + (d - 0.65).max(0.0) * 2.0
```
derekdreery, line 368: Is the `0.65` a magic number? Is it chosen because it is ~=2/3?

derekdreery (approved):
This looks awesome, I didn't realise that you also implemented a simplifier over BezPaths as well as improving the robustness of the original algorithm.

I didn't understand everything in the method, so I'm taking an 'empirical' approach to review - I've used the code in this branch and thrown some wierd curves at it, and it performs better than what's in `master` in all cases.

src/fit.rs
raphlinus: It wouldn't be a perfect match, because this is based on rounding, so giving the smallest absolute value, while rem_euclid gives the smallest nonnegative value. Also I checked on Godbolt and at least in some cases my code is faster, as fmod is a function call while rounding can be effected by vroundsd (when the avx level is supported).

PR #270, opened 2023-04-17
Replace panic with early return

derekdreery:
I need this because I'm using the curve offset algorithm and can't handle panics. I'm not sure if it's the best approach for kurbo.

### DIFF
--- a/src/fit.rs
+++ b/src/fit.rs
@@ -269,6 +269,7 @@ pub fn fit_to_cubic(
                     best = Some(err);
                 }
             }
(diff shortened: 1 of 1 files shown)


raphlinus:
Does this still trigger when #269 is applied? That addresses the panic in a way that I *think* is valid, but am not completely sure.

derekdreery:
I'll run with that patch tomorrow. If the problem doesn't arise, I'll close this PR :)

derekdreery:
I've experimented a bit and am not hitting the panic any more :smile: . I'll re-open if I manage to make it happen again.

PR #271, opened 2023-04-24
This patch adds some common aliases for vec2 methods.

### LINKED ISSUE #182: Some desired methods

derekdreery:
This should hopefully reduce confusion for new users. The suggestion was made in #182, and this PR will close #182.

### DIFF
--- a/src/point.rs
+++ b/src/point.rs
@@ -13,6 +13,15 @@ use crate::Vec2;
 use crate::common::FloatFuncs;
 
 /// A 2D point.
+///
+/// This type represents a point in 2D space. It has the same layout as [`Vec2`][crate::Vec2], but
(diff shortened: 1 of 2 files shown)


raphlinus (approved):
Thanks! Looks good to me.

PR #273, opened 2023-04-25
Add div_exact method to match cu2qu rounding behavior

### LINKED ISSUE #272: cubics_to_quadratic_splines yields different results from Python cu2qu
This code is described as a Rust port of fonttools cu2qu. However I noticed that it sometimes gives slightly different results with certain inputs.

```python
>>> from fontTools.cu2qu import curve_to_quadratic
>>> curve_to_quadratic([(408, 321), (408, 452), (342, 560), (260, 560)], 1.0)
[(408.0, 321.0),
 (408.0, 386.5),  # <== notice this 386.5
 (368.16666666666663, 495.0833333333333),
 (301.0, 560.0),
 (260.0, 560.0)]
```

```rust
>> use kurbo::{CubicBez, cubics_to_quadratic_splines};
>> let cubic = CubicBez::new(
    Point::new(408.0, 321.0),
    Point::new(408.0, 452.0),
    Point::new(342.0, 560.0),
    Point::new(260.0, 560.0)
);
>> cubic.approx_spline(1.0)
Some(QuadSpline(
    [(408.0, 321.0),
     (408.0, 386.49999999999994),  // <== this should have been 386.5!
     (368.16666666666663, 495.0833333333333),
     (301.0, 560.0),
     (260.0, 560.0)]))
```

I have tracked it down to numerical differences produced by the Vec2::div method being implemented in terms of multiplication by the reciprocal instead of straightforward division operator, which allegedly is less efficient (but more precise?): 

https://github.com/linebender/kurbo/blob/615ed7ede2d80ae0f3509165c5a2dee6cef69251/src/vec2.rs#L329-L340

When these x,y coordinates are rounded to integers, this very small numerical differences can add up and produce off-by-one differences in the rounded output.

Is there something kurbo can/should do about this? Can we change the cu2qu code in kurbo do use the division operator so that its output matches the python original?

dfrg:
This adds a (currently `pub(crate)`) `Vec2::div_exact` method that performs component-wise division and updates the cubic to quad spline code to make use of it. Allows us to match rounding behavior and output of fonttools cu2qu.

I'm unsure if there is a need to expose this method publicly.

Fixes #272

### DIFF
--- a/src/cubicbez.rs
+++ b/src/cubicbez.rs
@@ -259,8 +259,8 @@ impl CubicBez {
     /// Rust port of cu2qu [calc_cubic_points](https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/cu2qu/cu2qu.py#L63-L68)
     fn from_parameters(a: Vec2, b: Vec2, c: Vec2, d: Vec2) -> Self {
         let p0 = d.to_point();
-        let p1 = (c / 3.0).to_point() + d;
-        let p2 = ((b + c) / 3.0).to_point() + p1.to_vec2();
+        let p1 = c.div_exact(3.0).to_point() + d;
+        let p2 = (b + c).div_exact(3.0).to_point() + p1.to_vec2();
         let p3 = (a + d + c + b).to_point();
         CubicBez::new(p0, p1, p2, p3)
     }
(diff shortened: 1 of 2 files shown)


raphlinus (approved):
Looks good to me. My personal feeling is to expose it publicly if another use case emerges for it.

anthrotype:
Lgtm thanks! 🙏

PR #278, opened 2023-04-27
A couple of helpers in `bezpath.rs`.

derekdreery:
- Add a method to get mutable access to the elements slice.
 - Add a method to upcast a `PathSeg` to a `PathEl`, discarding the start point


raphlinus (approved):
These look useful. It's possible there's a place or two in the code that could be simpler using `as_path_el`, but that would be a low priority cleanup.

PR #280, opened 2023-05-05
Allow cloning of iterators over BezPath elements

derekdreery:
This makes it so you can clone iterators over `PathEl`s or `PathSeg`s.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -650,6 +650,7 @@ where
 /// An iterator that transforms path elements to path segments.
 ///
 /// This struct is created by the [`segments`] function.
+#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


src/bezpath.rs
```diff
@@ -734,6 +734,15 @@ impl<I: Iterator<Item = PathEl>> Segments<I> {
     }
 }
 
+impl<I: Clone + Iterator<Item = PathEl>> Clone for Segments<I> {
```
derekdreery, line 737: This might be able to be derived.

cmyr (approved):
Looks good, and agree that we should prefer derive if everything else is equal.

PR #286, opened 2023-05-20
First cut at stroke expansion

### LINKED ISSUE #285: Stroke expansion
This is a tracking issue for expanding strokes into fills, a common vector graphics operation. Among other things, it can be used to render strokes on top of a fill operation. That will be done as in intermediate step in Vello, before moving stroke expansion to a compute shader as described in https://github.com/linebender/vello/issues/303.

The essence of stroking is offset curves, and that infrastructure is now in reasonably good shape, though there is a tail of robustness issues (tracked in #279). The main substantive work is computing the line caps and joins. See [Nehab 2020] for a rather comprehensive discussion and many references. As discussed in the linked Vello issue, we will *not* target the most rigorously correct definition (which would require computation of evolutes) for performance reasons; perhaps that could be a future option.

The input is a path (discussion question: `impl Shape` is probably the friendliest approach, but it might be easier to have random access) and a stroke style. The latter is defined in peniko, in [style.rs](https://github.com/linebender/peniko/blob/cafdac9a211a0fb2fec5656bd663d1ac770bcc81/src/style.rs#L38). I don't think it's appropriate to take a dependency on that. Rather, in the grand scheme, it might be better to move the authoritative source for that type into kurbo, and have peniko re-export it as needed. There is also a tolerance parameter. The output is a `BezPath`.

There are many references for this algorithm, including:
* In zeno, [stroke.rs](https://github.com/dfrg/zeno/blob/master/src/stroke.rs)
* In Skia, [SkStroke.cpp](https://github.com/google/skia/blob/main/src/core/SkStroke.cpp)
* In Inkscape (livarot), [PathStroke.cpp](https://gitlab.com/inkscape/inkscape/blob/master/src/livarot/PathStroke.cpp)
* In FontForge, [fontforge#3991](https://github.com/fontforge/fontforge/pull/3991)

Of these, Skia should be considered a normative reference for compatibility and correctness. Perhaps some of the Skia tests can even be adapted.

See also the [Paper.js issue](https://github.com/paperjs/paper.js/issues/371) requesting this feature, which I believe has not landed.

Discussion question: what is the best type for the input path? Choices include:
* `impl Shape` - probably the friendliest)
* `impl IntoIterator<Item = PathEl>` - same as [simplify_bezpath](https://docs.rs/kurbo/latest/kurbo/simplify/fn.simplify_bezpath.html)
* `&[PathEl]` - random access might make life easier, especially for closed paths

A few more things. We may want to have some options guiding the optimization, analogous to [SimplifyOptions](https://docs.rs/kurbo/latest/kurbo/simplify/struct.SimplifyOptions.html). One of those may be whether each Bézier segment in the input path is offset independently, or whether curve fitting is applied to maximal G1-continuous ranges as is done for simplify. The latter may be fewer segments, but is also less editable.

[Nehab 2020]: http://dx.doi.org/10.1145/3386569.3392392

raphlinus:
This is a starting point; no dashes, only butt and miter, and closed paths are also not yet implemented.

Also not tested yet.

Will close #285

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,9 @@ features = ["mint", "schemars", "serde"]
 default = ["std"]
 std = []
 
+[dependencies]
+smallvec = "1.10"
+
 [dependencies.arrayvec]
 version = "0.7.1"
 default-features = false
--- a/src/stroke.rs
+++ b/src/stroke.rs
@@ -147,13 +147,17 @@ struct StrokeCtx {
     backward_path: BezPath,
     start_pt: Point,
     start_norm: Vec2,
+    start_tan: Vec2,
     last_pt: Point,
     last_tan: Vec2,
+    // if hypot < (hypot + dot) * bend_thresh, omit join altogether
+    join_thresh: f64,
 }
 
 /// Expand a stroke into a fill.
 pub fn stroke(path: impl IntoIterator<Item = PathEl>, style: &Stroke, tolerance: f64) -> BezPath {
     let mut ctx = StrokeCtx::default();
+    ctx.join_thresh = 2.0 * tolerance / style.width;
     for el in path {
         let p0 = ctx.last_pt;
         match el {
@@ -165,7 +169,8 @@ pub fn stroke(path: impl IntoIterator<Item = PathEl>, style: &Stroke, tolerance:
             PathEl::LineTo(p1) => {
                 if p1 != p0 {
                     let tangent = p1 - p0;
-                    ctx.do_tangents(style, tangent, tangent);
+                    ctx.do_join(style, tangent);
+                    ctx.last_tan = tangent;
                     ctx.do_line(style, tangent, p1);
                 }
             }
@@ -173,7 +178,8 @@ pub fn stroke(path: impl IntoIterator<Item = PathEl>, style: &Stroke, tolerance:
                 if p1 != p0 && p2 != p0 {
                     let q = QuadBez::new(p0, p1, p2);
                     let (tan0, tan1) = PathSeg::Quad(q).tangents();
-                    ctx.do_tangents(style, tan0, tan1);
+                    ctx.do_join(style, tan0);
+                    ctx.last_tan = tan1;
                     ctx.do_cubic(style, q.raise(), tolerance);
                 }
             }
@@ -181,11 +187,20 @@ pub fn stroke(path: impl IntoIterator<Item = PathEl>, style: &Stroke, tolerance:
                 if p1 != p0 && p2 != p0 && p3 != p0 {
                     let c = CubicBez::new(p0, p1, p2, p3);
                     let (tan0, tan1) = PathSeg::Cubic(c).tangents();
-                    ctx.do_tangents(style, tan0, tan1);
+                    ctx.do_join(style, tan0);
+                    ctx.last_tan = tan1;
                     ctx.do_cubic(style, c, tolerance);
                 }
             }
-            _ => todo!(),
+            PathEl::ClosePath => {
+                if p0 != ctx.start_pt {
+                    let tangent = ctx.start_pt - p0;
+                    ctx.do_join(style, tangent);
+                    ctx.last_tan = tangent;
+                    ctx.do_line(style, tangent, ctx.start_pt);
+                }
+                ctx.finish_closed(style);
+            }
         }
     }
     ctx.finish(style);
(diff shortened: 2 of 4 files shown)


src/stroke.rs
```diff
@@ -0,0 +1,254 @@
...
+                    ctx.do_cubic(style, c, tolerance);
+                }
+            }
+            _ => todo!(),
+        }
+    }
+    todo!()
+}
+
+fn get_end(el: &PathEl) -> Point {
```
derekdreery, line 187: I think there should be an `end_point` method on `PathEl` that returns `Option<Point>` and is implemented exactly like this.

Cargo.toml
```diff
@@ -17,6 +17,9 @@ features = ["mint", "schemars", "serde"]
 default = ["std"]
 std = []
 
+[dependencies]
+smallvec = "1.10"
```
notgull, line 21: I think `tinyvec` would be better if possible since it has no unsafe code.

src/stroke.rs
```diff
@@ -0,0 +1,630 @@
...
+    /// Builder method for setting whether or not the stroke should be affected
+    /// by the scale of any applied transform.
+    pub fn with_scale(mut self, yes: bool) -> Self {
+        self.scale = yes;
+        self
+    }
+}
+
+/// Collection of values representing lengths in a dash pattern.
+pub type Dashes = SmallVec<[f64; 4]>;
```
notgull, line 141: It may be better to do three dashes by default, since on 64 bit platforms there would be no wasted stack space compared to a `Vec`.

notgull (commented):
+1 to this. I need a stroking algorithm for my `piet` implementations that supports dashing. Both `tiny-skia` and `zeno` produce invalid output when I try to use them for all (although I may just be all thumbs in these cases).

notgull:
piet-glow-03-2 00

Looks like it's skipping some segments? For comparison, here's what `piet-samples` looks like:

sample

notgull:
In addition, I'm getting stack overflows in this function under certain dash patterns. Standby for more details

raphlinus:
Are you also patching #289 with the stack overflows? We need to get these landed, then consider more improvements on the remaining failures.

src/stroke.rs
```diff
@@ -0,0 +1,630 @@
...
+    start_norm: Vec2,
+    start_tan: Vec2,
+    last_pt: Point,
+    last_tan: Vec2,
+    // if hypot < (hypot + dot) * bend_thresh, omit join altogether
+    join_thresh: f64,
+}
+
+/// Expand a stroke into a fill.
+pub fn stroke(path: impl IntoIterator<Item = PathEl>, style: &Stroke, tolerance: f64) -> BezPath {
```
dfrg, line 160: Did we want to add an additional options parameter to this? I believe one use was to toggle optimal curve fitting.

src/stroke.rs
```diff
@@ -0,0 +1,630 @@
...
+                if p1 != p0 && p2 != p0 {
+                    let q = QuadBez::new(p0, p1, p2);
+                    let (tan0, tan1) = PathSeg::Quad(q).tangents();
+                    ctx.do_join(style, tan0);
+                    ctx.last_tan = tan1;
+                    ctx.do_cubic(style, q.raise(), tolerance);
+                }
+            }
+            PathEl::CurveTo(p1, p2, p3) => {
+                if p1 != p0 && p2 != p0 && p3 != p0 {
```
dfrg, line 205: Per f2f this appears to be too strict. For example, there are valid cases where `p1 == p0`. Same above for the quad case.

src/stroke.rs
```diff
@@ -0,0 +1,630 @@
...
+                    let c = CubicBez::new(p0, p1, p2, p3);
```
dfrg, line 206: friendly note: `.regularize()` :)

dfrg (approved):
This is a really solid foundation! Let’s land it, propagate the stroke type through peniko, and hook it up to vello so we have a nice playground for working through the remaining issues.

src/stroke.rs
```diff
@@ -0,0 +1,675 @@
...
+
+    fn do_cubic(&mut self, style: &Stroke, c: CubicBez, tolerance: f64, opts: &StrokeOpts) {
+        let co = CubicOffset::new(c, -0.5 * style.width);
+        let forward = fit_with_opts(&co, tolerance, opts);
+        self.forward_path.extend(forward.into_iter().skip(1));
+        // A tuning parameter for regularization. A value too large may distort the curve,
+        // while a value too small may fail to generate smooth curves. This is a somewhat
+        // arbitrary value, and should be revisited.
+        const DIM_TUNE: f64 = 0.25;
+        let co = CubicOffset::new_regularized(c, 0.5 * style.width, tolerance * DIM_TUNE);
```
dfrg, line 427: I’m confused as to why is this done here and only for the backward path? In my testing version, I called regularize on the initial cubic itself in stroke_undashed before computing the tangents for joins. My assumption was that regularize fixes zero length tangent vectors and we would want consistency there.

raphlinus: Simple brain fart error. I should have validated this more carefully before pushing. Thanks for the catch!

dfrg (approved):
Great! Let’s merge this and I’ll find some time soon to do serious stress testing and we’ll work out the remaining issues.

PR #287, opened 2023-05-27
Implement TranslateScale * Affine (both directions)

derekdreery:
As part of testing this, I also added `float_cmp` to dev dependencies and implemented it for Point/Affine. I propose that we start to standardize on this method for checking approximate equality.

Finally, I've added some MathML to the docs to make it so people can understand the code. In the TranslateScale * Affine (and rev) case, one probably doesn't need more docs, but having them proves the concept.


src/affine.rs
```diff
@@ -223,6 +223,28 @@ impl Affine {
...
+    ///
+    /// ```
+    /// # use kurbo::Affine;
+    /// # use float_cmp::assert_approx_eq;
+    /// let affine = Affine::new([1., 2., 3., 4., 5., 6.]);
+    /// assert_approx_eq!(f64, affine.norm_l_1_1(), affine.as_coeffs().into_iter().sum::<f64>() + 1.);
+    /// ```
+    ///
+    /// [matrix-norm#entry-wise]: https://en.wikipedia.org/wiki/Matrix_norm#%22Entry-wise%22_matrix_norms
+    pub fn norm_l_1_1(self) -> f64 {
```
derekdreery, line 243: I added this function for error checking, before deciding that max(element-wise error) was a better way. I've left it in to stimulate conversation about matrix norms, but fully expect to need to remove it before merging.

derekdreery: If we did keep it I almost certainly think we should remove the `1`, otherwise |a - a| != 0.

PR #288, opened 2023-06-02
Find tangents from point to cubic Bézier

raphlinus:


src/cubicbez.rs
```diff
@@ -348,6 +348,26 @@ impl CubicBez {
             .filter(|t| *t >= 0.0 && *t <= 1.0)
             .collect()
     }
+
+    /// Find lines from the point to the curve tangent to the curve.
+    ///
+    /// Result is array of t values such that the line from the given point
+    /// to the curve evaluated at that value is tangent to the curve.
+    pub fn point_tangents(&self, p: Point) -> ArrayVec<f64, 4> {
```
Keavon, line 356: Would `tangents_though_point` be a clearer name?

src/cubicbez.rs
```diff
@@ -348,6 +348,26 @@ impl CubicBez {
             .filter(|t| *t >= 0.0 && *t <= 1.0)
             .collect()
     }
+
+    /// Find lines from the point to the curve tangent to the curve.
```
Keavon, line 352: On my initial few read-throughs of this sentence I had a hard time groking it. I can also take a shot at wording it, which might also be flawed, but maybe we can meet in the middle with something that's clearer to unbiased readers:

> Find points on the curve where each one's tangent line passes through the given point in space.

Keavon (commented):
🎉

src/cubicbez.rs
raphlinus: I made it a little shorter (it should be one line), but agree this is clearer. I think in my mind I was thinking about the line from the point to the curve, but thinking about it the other way is simpler because it's more apparent that you're getting both the tangent and the incidence.

waywardmonkeys:
@raphlinus i have heard that @Keavon  might be interested in this being in a kurbo release… thoughts?

raphlinus:
I've fixed the merge conflict. @Keavon can you confirm this is still useful, and that review comments have been addressed to your satisfaction?

Keavon:
@raphlinus This is still useful and the code/comments indeed look good!

It would also be useful to have a `normal_to_point` function as well, and add support for quadratic and linear segments in addition to cubic. (For linear, only `normal` is applicable and it will only return a single t value.) But I wouldn't suggest delaying this until those can be included, because this is useful in its own right.

For some context surrounding Graphite and Bezier-rs:

In Graphite, we're preparing to begin replacing some Bezier-rs functionality with Kurbo because it has better robustness, so this will be helpful to have. Our motivation for making Bezier-rs was twofold:
- Needing more features/algorithms (normals/tangents to point is an example of one). But our scope also includes more sophisticated ones like [Poisson-disk point sampling](https://graphite.rs/libraries/bezier-rs/#subpath/poisson-disk-points) that would be out of scope for Kurbo. Rapid implementation mattered more than robustness/correctness in the short term for us, but now we're getting towards desiring more robustness. Kurbo's performance is likely also considerably better and that's starting to matter to us a lot more too.
- Needing to represent the data structure as "manipulator groups" (anchor + that anchor's 0/1/2 handles) rather than SVG `d` path command style segments (0/1 outward handle, 0/1/ inward handle, and anchor) so our tools didn't need to pay the performance penalty for translating between Kurbo SVG-style and our tools' manipulator group style; but this requirement has changed and is no longer a factor, which means Kurbo would be just as efficient for us, leaving only point 1 above regarding features. Hence, the desire to have these tangent/normal to point algorithms implemented.

Maintenance and the future of Bezier-rs is still up in the air but I'm envisioning it as a possible wrapper for Kurbo that could use those algorithms that exist and augment it with our additional features that Kurbo currently lacks (or always will), such as the Poisson-disk point sampling linked above. Downside: Bezier-rs would no longer be a zero-dependency crate. Upside: we can reduce our maintenance burden through deleting a lot of our code by wrapping Kurbo's algorithms, provide better robustness and performance, and probably continue to use Bezier-rs as a testbed or incubator for more complex algorithms instead of needing to officially abandon our maintenance for the crate if we fully switched to Kurbo in Graphite.

Philipp-M (approved):
(Approving this, as @Keavon doesn't have the rights to do this, but has confirmed usefulness)

waywardmonkeys:
Unless I hear otherwise or someone else merges this, I will merge this during or after office hours today.

raphlinus:
I'm merging it now. Sounds like there's a good discussion to be had about future coevolution of bezier-rs and kurbo. The poisson disk feature looks quite interesting, but as you say it's probably not in scope for kurbo.

waywardmonkeys:
I'm very excited about bezier-rs moving some things into Kurbo!

Keavon:
I should go through and create a comparison chart about which algorithms each of our libraries implement in order to focus attention towards those which may be useful in Kurbo. But I need to emphasize that our algorithms are somewhat janky and probably aren't fast or robust enough for the standards of Kurbo. Though it would still be a useful way to evaluate what may be worth reimplementing in Kurbo with better attention towards those fine details.

Issue #279, opened 2023-05-04
Tracking issue for curve fitting robustness


derekdreery:
This is a new tracking issue for fitting robustness. There shouldn't be too many problems once #269 lands, but if there are any, please post them here.

This issue also tracks curve offset, as that is a special case of curve fitting.

raphlinus:
Achieving perfection here will take a lot of work, but there are two things I plan to do in the coming days which I believe will help a lot.

First, make the `sample_pt_tangent()` implementation in `CurveOffset` handle zero and near-zero derivatives in the source curve, as well as cusps caused by offsets. There's some logic for that in `simplify_bezpath()`, but that's at a higher level, before it gets into curve fitting.

Second, degenerate curves are manifesting as infinite recursion, it keeps subdividing (apparently sometimes to zero length) and you get either a hang or a stack overflow depending on the details. I'll add special case logic to handle the endpoints within the given accuracy of each other. You *usually* don't want to get there, but if you do, it should just draw a line, which is the simplest answer that meets the Fréchet bound.

raphlinus:
There are a number of test cases floating around, but I'll start with [this one](https://github.com/linebender/kurbo/pull/269#issuecomment-1517937843) from #269, as it's clearly documented. It hangs with the _opt variant, but also generates a stack overflow (indicating infinitely recursing subdivision) in the non-opt case. Just looking at the input, it's not obvious to me where the problem is, as the source curve has perfectly fine derivatives everywhere. I'll focus on that one first, then bring in more cases as needed.

valadaptive:
Not sure if this falls under the category of "robustness" or simply subpar results, but I'm trying to use kurbo's curve-simplification algorithm and there are some pretty large "bumps" in the resulting paths, of the sort [described in your blog post](https://raphlinus.github.io/curves/2023/04/18/bezpath-simplify.html):

![image](https://github.com/user-attachments/assets/718f204c-c15f-4743-a994-74dcae410a4d)
![image](https://github.com/user-attachments/assets/c015a0c4-ff02-4931-9a43-711002423879)

Tweaking the currently-hardcoded `D_PENALTY_ELBOW` and `D_PENALTY_SLOPE` values helps with this, but they need to be set to some pretty aggressive values.

[...]

PR #289, opened 2023-06-08
Improve robustness of curve fitting

raphlinus:
The main improvement is to try to fit a line when the chord is very short. Very short segments can occur when the cusp finding reports imprecise results, and also the main cubic fit logic is not robust when the chord is short.

I believe the simple subdivision case is now fairly robust in that it won't keep recursing. This is in spite of not having an explicit bound on recursion depth; the theory is that each subdivision reduces the parameter space in half, and at some point you'll reach the point where the chord is shorter than the given tolerance. This is true even for an adversarial break_cusp, as it's bypassed in the short chord case.

The logic for the optimized case is not as careful, in particular break_cusp is not bypassed. I'm curious whether there are still failure cases.

This is likely not the end of the numerical robustness work. In particular, break_cusp can be better tuned to not over-report cusps, and the tangent detection can be improved for cusp and near-cusp source curves. But hopefully this commit gets us to where we at least return a valid fit.

Also adds a test lightly adapted from #269.

Progress toward #269 and #279

### DIFF
--- a/src/cubicbez.rs
+++ b/src/cubicbez.rs
@@ -44,6 +44,15 @@ struct ToQuads {
     n: usize,
 }
 
+/// Classification result for cusp detection.
+#[derive(Debug)]
+pub enum CuspType {
+    /// Cusp is a loop.
+    Loop,
+    /// Cusp has two closely spaced inflection points.
+    DoubleInflection,
+}
+
 impl CubicBez {
     /// Create a new cubic Bézier segment.
     #[inline]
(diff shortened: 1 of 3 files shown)


dfrg:
This is a substantial improvement. We’re still hitting a stack overflow with tiger that is caused when `hypot == 0` in `CubicOffset::eval_offset`. If I hack in a fix for that, this succeeds. 

Further, if I remove all the point equality comparisons guarding cubics in the stroke converter, I get almost perfect strokes on tiger! There are a few very minor artifacts remaining that I haven’t investigated yet.

raphlinus:
Good to hear. I haven't run through tiger myself, but it sounds like that's a good set of test data, and we should try to nail down all the remaining artifacts.

dfrg (approved):
I think we still have more work to do but this improves the current state of things dramatically and I’m happy to merge as is and address further issues in a future PR.

PR #290, opened 2023-06-08
Implement reflection over a line -> affine transformation

derekdreery:
Implementation using Householder matrix, which is numerically robust IIUC. Sometimes other operations can be made more robust by using reflections.


raphlinus (approved):
Sure, looks good. Thanks!

PR #297, opened 2023-07-13
Rename master branch to main in CI.

xStrom:
Now that `master` has been renamed to `main` we also need to update the CI to match.


raphlinus (approved):
Yay!

PR #300, opened 2023-08-08
Cubic offset that triggers infinite recursion

dfrg:
Adds a simple test for a curve that triggers infinite recursion during fitting for an offset. This was pulled from a run of the mmark example. Notably, the failing cubics all seem to have exactly the same y coordinates.

At the point of failure, the `start..end` parameters are nearly identical (within 1e-12!) but the sampled points never converge within reasonable tolerance so line fitting is not attempted.

@raphlinus let me know if I can do any additional investigation on this!

### DIFF
--- a/src/fit.rs
+++ b/src/fit.rs
@@ -697,3 +697,25 @@ fn fit_opt_err_delta(
     let err = measure_one_seg(source, t0..t1, limit).unwrap_or(accuracy * 2.0);
     Ok(accuracy - err)
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


raphlinus:
Thanks! This is very clearly an instance of the "colinear" case described on the [cubic robustness](https://linebender.org/wiki/cubic-robustness/) page, where the cusp is to the left of both endpoints. So now there are two things to do: first, figure out why there's still infinite recursion even though that's supposed to be bounded even in pathological situations, and second, make the detection more robust to include colinear cases.

PR #301, opened 2023-08-28
Mul and Div for two Vec2s

tz629:
Currently there is only mul and div on vec2s by an f64 number. i added the ability to multiply/divide a vector2 by another vector2. you might want to use * .recip() instead of /.


tz629:
I don't have a specific use case, It just seemed useful. Considering you left this out for a reason, I'll close this

PR #304, opened 2023-09-28
Better numerical robustness for stroking

raphlinus:
A number of fixes to improve stroke robustness.

Infinite recursion on curve fit is avoided by checking whether subdivision has hit the floating point limit. That said, the quality of curve fit is poor when the input is discontinuous (which does happen for degenerate cubics with cusps, where all points are co-linear).

Thus, in stroking, the degenerate cases are detected and drawn with special-case logic (do_linear). This uses existing drawing capabilities to draw the degenerate cubic with lines and round joins.

Finally, do_join mistakenly detected a half-turn as zero turn, as it was only considering the cross product (which vanishes).

Fixes #300

### DIFF
--- a/src/fit.rs
+++ b/src/fit.rs
@@ -195,6 +195,16 @@ fn fit_to_bezpath_rec(
         // A smarter approach is possible than midpoint subdivision, but would be
         // a significant increase in complexity.
         let t = 0.5 * (start + end);
+        if t == start || t == end {
+            // infinite recursion, just draw a line
+            let p1 = start_p.lerp(end_p, 1.0 / 3.0);
+            let p2 = end_p.lerp(start_p, 1.0 / 3.0);
+            if path.is_empty() {
+                path.move_to(start_p);
+            }
+            path.curve_to(p1, p2, end_p);
+            return;
+        }
         fit_to_bezpath_rec(source, start..t, accuracy, path);
         fit_to_bezpath_rec(source, t..end, accuracy, path);
     }
--- a/src/offset.rs
+++ b/src/offset.rs
@@ -194,4 +194,19 @@ mod tests {
         let path_opt = fit_to_bezpath(&offset_path, accuracy);
         assert!(matches!(path_opt.iter().next(), Some(PathEl::MoveTo(_))));
     }
+
+    /// Cubic offset that used to trigger infinite recursion.
+    #[test]
+    fn infinite_recursion() {
+        const DIM_TUNE: f64 = 0.25;
+        const TOLERANCE: f64 = 0.1;
+        let c = CubicBez::new(
+            (1096.2962962962963, 593.90243902439033),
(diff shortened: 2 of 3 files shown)


raphlinus:
Hmm, these are obviously false positives, but I've desk-checked the code and tried to write a repro (I originally suspected a case where one control point is coincident with an endpoint) and am not finding anything. I'll dig more deeply. Could you perhaps upload your vello branch? (doesn't have to be cleaned up in any way)

raphlinus:
I am only able to repro when I set unreasonably high tolerance values, so I am wondering if we can rule that out. The break between detection and not is happening exactly where I'd expect when I vary the tolerance. That said, I am seeing some distorted rendering when the chord and the tangent don't match, so will probably refine that (and, ideally, write a test).

raphlinus:
I'd appreciate isolated failing examples. Clearly there's more work to be done to validate this, and perhaps I should invest time to make a test suite (I can see that coming in handy when moving to GPU). But right now I just want to get something "good enough" landed.

dfrg:
I'll try to extract a failing case or two from tiger if I can identify them.

raphlinus:
Got it! For reference, here is the repro case:

Screenshot 2023-09-29 at 10 21 56 AM

```rust
    let curve = CubicBez::new((83.083, 120.279), (83.004, 119.572), (83.004, 119.572), (83.711, 119.886));
    let path = curve.into_path(0.1);
    let stroke_style = Stroke::new(0.039);
    let stroked = stroke(path, &stroke_style, &Default::default(), 1.0);
```

And I can see what's going on. Because the control arm is longer than the chord, that's being chosen as the reference chord, but the p01 cross-product is zero because p0 = p1 here.

So I was right about the coincident endpoint, it's just I didn't have the creativity to try a short chord when making test cases.

raphlinus:
A followup on the "distorted rendering" comment above. What I had seen was when the tolerance was set to a very high value, so was within spec. When I create a test case with tangents not matching the chord (short control handles), it renders as expected. In particular, stroking (10., 10.), (9.99, 9.99), (19.99, 9.99), (20., 10.) with a width of 1 and a tolerance of 0.1 gives the following, which is as expected:

Screenshot 2023-10-02 at 9 43 10 AM

This is a challenging case and definitely gets into the territory (as in the Nehab paper) where different renderers will give different results. The correct rendering (tolerance 0.001) is below:

Screenshot 2023-10-02 at 9 50 09 AM

And, for reference, here's what Skia does for a similar example:

Screenshot 2023-10-02 at 9 50 52 AM

So as far as I know, this code is now correct. Of course there may be edge cases still remaining, but I believe to uncover them requires

dfrg (approved):
Thanks for getting this done! Haven’t gone over the math in detail but this looks solid at a glance and produces correct results in all scenes now. Let’s get it merged. 

🚢

PR #306, opened 2023-10-03
Add control_box method to BezPath

dfrg:
This adds a method to `BezPath` that computes a conservative bounding box only taking control points into account.


raphlinus (approved):
I think I'd slightly prefer returning `Option<Rect>`, but Chad points out that this is more consistent with the existing `bounding_box` method.

PR #307, opened 2023-10-03
Add reverse_subpaths() method to BezPath

dfrg:
Returns a new path with winding direction of all subpaths reversed.

This is mostly based on the code here: https://github.com/googlefonts/fontations/blob/d4e07eb3b45836f3241e5a5343fdb697c40d978b/write-fonts/src/pens.rs#L164 which it is intended to replace. Tests are taken from that code as well.

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -456,24 +456,16 @@ impl BezPath {
             match el {
                 PathEl::MoveTo(pt) => {
                     if start_ix < ix {
-                        reverse_subpath(
-                            start_pt,
-                            false,
-                            elements.get(start_ix..ix)?,
-                            &mut reversed,
-                        );
+                        reverse_subpath(start_pt, elements.get(start_ix..ix)?, &mut reversed);
                     }
                     start_pt = *pt;
                     start_ix = ix + 1;
                 }
                 PathEl::ClosePath => {
                     if start_ix < ix {
-                        start_pt = reverse_subpath(
-                            start_pt,
-                            true,
-                            elements.get(start_ix..ix)?,
-                            &mut reversed,
-                        );
+                        start_pt =
+                            reverse_subpath(start_pt, elements.get(start_ix..ix)?, &mut reversed);
+                        reversed.push(PathEl::ClosePath);
                     }
                     start_ix = ix + 1;
                 }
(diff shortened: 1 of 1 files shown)


src/bezpath.rs
```diff
@@ -440,6 +440,63 @@ impl BezPath {
         }
         cbox.unwrap_or_default()
     }
+
+    /// Returns a new path with the winding direction of all subpaths reversed.
+    ///
+    /// If the path does not start with a move, returns `None`.
```
raphlinus, line 446: The behavior doesn't quite match the comment - on empty path input, the result is also empty (which is valid), but it's not a path that starts with a move. Given that a path starting with an element other than move is an invalid state, I'm not sure it should complicate the return type.

src/bezpath.rs
```diff
@@ -440,6 +440,63 @@ impl BezPath {
...
+    pub fn reverse_subpaths(&self) -> Option<BezPath> {
```
cmyr, line 447: If it doesn't start with a move we could also return the path, unchanged? I'm not sure how annoying the option is, here.

cmyr (approved):
good idea, it makes way more sense to have this live here.

src/bezpath.rs
dfrg: My initial cut didn't have the Option but I was trying to match behavior of the original code. On second thought, we should be doing validation outside of contour reversal, so I'll go ahead and remove this.

src/bezpath.rs
```diff
@@ -440,6 +440,63 @@ impl BezPath {
...
+        let elements = self.elements();
+        if !elements.is_empty() && !matches!(elements.first(), Some(PathEl::MoveTo(_))) {
+            return None;
```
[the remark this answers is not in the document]
dfrg: Yeah, this didn't make me too happy either. See my comment to Raph above about the return type.

src/bezpath.rs
```diff
@@ -1535,4 +1592,65 @@ mod tests {
             .collect::<Vec<_>>();
         assert_eq!(segments, get_segs);
     }
+
+    #[test]
+    fn test_control_box() {
```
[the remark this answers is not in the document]
dfrg: Too much attention to detail :) I added control_box earlier and missed the test. Since I was copying tests over for contour reversal, I thought I might sneak this in.

Caught!

src/bezpath.rs
dfrg: As above, I think we should do some sort of validation step in fontc/write-fonts instead. I was trying to match behavior of `ReverseContourPen` but now I'm realizing that we're only checking validity when reversing contours which seems broken to me.

PR #309, opened 2023-10-04
Additional tests and fixes for BezPath::reverse_subpaths

dfrg:
The initial PR for this was missing a large batch of test cases. Adding these revealed that subpath reversal did not accurately capture degenerate contours (move only or move + close) which we'd like to retain in glyphs.

This adds the tests and updates the code to pass them (ref: https://github.com/googlefonts/fontations/blob/145a2ae269f30a3fbc217ef9cddb18a8a80a8f5b/write-fonts/src/pens.rs#L562C1-L564C104)

### DIFF
--- a/src/bezpath.rs
+++ b/src/bezpath.rs
@@ -447,27 +447,39 @@ impl BezPath {
         let mut start_ix = 1;
         let mut start_pt = Point::default();
         let mut reversed = BezPath(Vec::with_capacity(elements.len()));
+        // Pending move is used to capture degenerate subpaths that should
+        // remain in the reversed output.
+        let mut pending_move = false;
         for (ix, el) in elements.iter().enumerate() {
             match el {
                 PathEl::MoveTo(pt) => {
+                    if pending_move {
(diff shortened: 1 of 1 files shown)


src/bezpath.rs
```diff
@@ -1648,4 +1660,360 @@ mod tests {
...
+    #[test]
+    fn test_reverse_single_point() {
+        reverse_test_helper(
+            vec![PathEl::MoveTo((0.0, 0.0).into())],
+            vec![PathEl::MoveTo((0.0, 0.0).into())],
+        )
+    }
+
+    #[test]
+    fn test_reverse_single_point_closed() {
```
anthrotype, line 1904: this test and the one immediately preceding are exactly the same as test_reverse_move_only_path_offset_validity and test_reverse_move_close_path_offset_validity, are they not?

anthrotype (approved):

src/bezpath.rs
dfrg: They appear to be, but I just copied them all directly from our tests in write-fonts :) I'll remove the redundant ones.

PR #318, opened 2023-10-18
Improve cusp detection for strokes

raphlinus:
The cusp detection was failing because a comparison wasn't inclusive enough. In particular, it was trying to compute whether curvature at the point exceeded 1/dimension, without doing division, but both values in the comparison become zero when the derivative vanishes.

Also avoids infinite recursion when break_cusp reports a cusp within ulp of an endpoint. This is supposed to be a violation of the contract, but it's not hard to see how it can happen.

Includes test cases adapted from vello#388


armansito (approved):
Looks good to me, and I confirmed that the tests in vello#388 render correctly with this change.

PR #319, opened 2023-10-19
Prepare 0.10.4 release; DashIterator doc

raphlinus:
Improve documentation of DashIterator and bump version in preparation for release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kurbo"
-version = "0.10.3"
+version = "0.10.4"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2021"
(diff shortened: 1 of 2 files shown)


dfrg (approved):
Just bikeshedding a bit: I wasn’t aware this was already public. The API incongruity (with `segments` and maybe `flatten`) is a bit unfortunate. I would have preferred `fn dash(…) -> impl Iterator` (or `-> Dash` if following Rust naming guidelines and matching Segments if we want the concrete type exposed).

raphlinus:
I'm ok with changing it to look like a function. That should probably be a bump to 0.11, as it would certainly be semver breaking, unless we left it in as an alias with doc(hidden) maybe.

If it were changed to a function, the main comment should be attached to that, not the type, yes?

dfrg:
I was thinking the same thing wrt doc(hidden) :)

And yes, the lovely new doc comment should be on the function in this case.

PR #321, opened 2023-10-29
Add serde/schemars derive to Stroke, Join and Cap.

ratmice:
This derives some traits for Stroke, friends.  It enables the features `schemars/smallvec`, and `serde/smallvec` so that `Dashes` within Stroke derives `JsonSchema`/`Serialize`/`Deserialize` respectively.  The impetus for this was that I was trying to add the same traits to peniko's `Style`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,8 @@ features = ["mint", "schemars", "serde"]
 [features]
 default = ["std"]
 std = []
+serde=["smallvec/serde", "dep:serde"]
+schemars=["schemars/smallvec", "dep:schemars"]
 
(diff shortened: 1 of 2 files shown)


jneem (approved):
I'm not actually sure if I still have write access here, but this seems pretty uncontroversial so let's try.

ratmice:
I think that your approval worked, at least it appears to be approved in the main PR queue.
However I definitely do not have write access, so hopefully no-one is waiting for me to merge it.
I'm not in any hurry though. I just wanted to note that in case the responsibility to merge is mine.

PR #323, opened 2023-11-13
Make TranslateScale::from_scale_about to take `Into<Point>`.

ratmice:
bring it into alignment with `Rect` methods that take `Into<Point>`.

Randomly noticed this, this patch doesn't do a full audit of the methods taking `Point`, should (we, or I) do such?
I went ahead and had a look through, here are the functions that take Points as parameters that we could consider
changing.

1. The `nearest` functions of `ParamCurve`, `CubicBez`, `QuadBez`, `PathSeg`, and `Line`.
2. `Shape::winding`, and `Shape::contains` trait methods.
3. The `other` variable, of various Point functions `lerp` `midpoint`, `distance`, `distance_squared`
4. `Rect::union_pt`
5. `Ellipse::with_center`
6. `Affine::rotate_about`, `pre_rotate_about`, `then_rotate_about`.

Here are some things I don't think should change, and reasons
4. While the `new` fn of `QuadSpline` takes a `Vec<Point>`, it takes ownership, turning it into an `IntoIterator` or something
Would force an allocation in collect.

### DIFF
--- a/src/translate_scale.rs
+++ b/src/translate_scale.rs
@@ -91,11 +91,11 @@ impl TranslateScale {
     /// assert_near(ts * Point::new(2., 2.), Point::new(3., 3.));
     /// ```
     #[inline]
-    pub fn from_scale_about(scale: f64, focus: Point) -> Self {
+    pub fn from_scale_about(scale: f64, focus: impl Into<Point>) -> Self {
         // We need to create a transform that is equivalent to translating `focus`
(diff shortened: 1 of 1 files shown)


derekdreery (approved):
LGTM. Out of scope for this PR, but we need more tests - it would be nice to have every method tested with at least 'normal' input.

PR #325, opened 2023-11-17
Handle special case of fitting a line

raphlinus:
When the algorithm is asked to curve-fit a line, area and moment are both zero, which provokes numerical instablily. Just return a straight line in that case.


derekdreery (approved):
I've been using this method in production for a while (check for straight line) and it seems to solve the problem. 🙂

derekdreery:
Hi would you mind if I merged this and cut a release?

raphlinus:
Feel free to do a release, sure.

PR #326, opened 2023-11-27
More fixes for strokes

raphlinus:
Fixes a couple of non-working cases: stroke of an empty path, and a tricky co-linear case where the numerical robustness of the quadratic solve becomes poor.

### DIFF
--- a/src/stroke.rs
+++ b/src/stroke.rs
@@ -338,6 +338,9 @@ impl StrokeCtx {
 
     /// Finish a closed path
     fn finish_closed(&mut self, style: &Stroke) {
+        if self.forward_path.is_empty() {
+            return;
+        }
         self.do_join(style, self.start_tan);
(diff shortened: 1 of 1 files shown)


src/stroke.rs
```diff
@@ -502,7 +505,7 @@ impl StrokeCtx {
         let c2 = p[3] - 3.0 * p[2] + 3.0 * p[1] - p[0];
         let roots = solve_quadratic(c0, c1, c2);
         // discard cusps right at endpoints
-        const EPSILON: f64 = 1e-12;
+        const EPSILON: f64 = 1e-6;
```
derekdreery, line 508: I wonder if it's worth spending some time rationalizing all the accuracy parameters sometime (saying why they are the value they are, or if they are arbitrary).

derekdreery (approved):

PR #327, opened 2023-12-05
Support for no-alloc

jneem:
A surprising amount of useful kurbo code doesn't require allocation (which is really only used for `BezPath` and its dependents). This adds a feature for turning off allocation support altogether.

Note that this change might break anyone who uses kurbo with `default-features = false`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,11 @@ features = ["mint", "schemars", "serde"]
 [features]
 default = ["std", "alloc"]
 std = ["alloc"]
-alloc = ["serde/alloc"]
+alloc = ["serde/alloc", "smallvec"]
 
-[dependencies]
-smallvec = "1.10"
+[dependencies.smallvec]
+version = "1.10"
+optional = true
 
 [dependencies.arrayvec]
 version = "0.7.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,9 @@ categories = ["graphics"]
 features = ["mint", "schemars", "serde"]
 
 [features]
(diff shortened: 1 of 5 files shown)


xStrom:
There was a tiny bit of discussion about this PR during [office hours](https://xi.zulipchat.com/#narrow/stream/359642-office-hours/topic/2024-01-04) and a question that came up was the use case. I'm relaying that here to help move this forward.

Could you describe your use case a bit where no-alloc is beneficial? No need for a massive rationale, but because this does add a bit of complexity to the code and testing, it would be useful to have a bit more specifics about the benefits.

jneem:
Ah, sorry, I replied on the zulip before seeing this comment. I'm using kurbo on an esp32c3 device, using [esp-hal](https://github.com/esp-rs/esp-hal) which doesn't support alloc out-of-the-box. It's quite possible that this is too niche to be worth supporting upstream, in which case I'm quite happy to maintain my own patchset.

Cargo.toml
```diff
@@ -15,11 +15,13 @@ categories = ["graphics"]
 features = ["mint", "schemars", "serde"]
 
 [features]
-default = ["std"]
-std = []
+default = ["std", "alloc"]
```
waywardmonkeys, line 18: My brain is fried, but can't this just be `default = ["std"]` since the line below handles enabling `"alloc"`?

jneem:
I've done some more experimentation, and I think I'm better off switching to euclid and lyon_geom for the embedded part. They already have no-alloc (and f32) support, and the conversions aren't too painful.

PR #329, opened 2024-01-16
Replace panic with error return in `BezPath::from_svg`

platlas:
As pointed out in https://github.com/linebender/kurbo/issues/303#issuecomment-1856382911 `BezPath::from_svg` does not handle case when SVG path does not start with initial point. This PR makes the `BezPath::from_svg` return an error for this case.


platlas:
It is failing on clippy lints: `unused_enumerate_index` (that was added to latest version) and `items_after_test_module`. But I need to find out why clippy have not warned me on my local machine even if I use latest version.

raphlinus (approved):
Agree, panicking is not a good thing to do for simply invalid input. Thanks!

PR #332, opened 2024-02-06
Make the FloatFuncs trait a sealed public trait, add sin/sinf.

ratmice:
This trait, with the addition of sin/sinf is needed to build peniko with the `--no-default-features --features libm` combination (added in the pr below).

The approach taken here is that rather than duplicating this common module in peniko, to make it pub, and add the sin/sinf. for peniko.  This trait is only conditionally defined so use of it in peniko would look something like:

```
+#[cfg(not(feature = "std"))]
+#[allow(unused_imports)]
+use kurbo::common::FloatFuncs as _
```

This patch is in regards to fixing the testsuite failures for default features added in
https://github.com/linebender/peniko/pull/23

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -5,6 +5,16 @@
 
 #![allow(missing_docs)]
 
+#[cfg(not(feature = "std"))]
+mod sealed {
+    /// A [sealed trait](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/)
+    /// which stops [`super::FloatFuncs`] from being implemented outside kurbo. This could
+    /// be relaxed in the future if there is are good reasons to allow external impls.
+    /// The benefit from being sealed is that we can add methods without breaking downstream
+    /// implementations.
+    pub trait FloatFuncsSealed {}
+}
+
 use arrayvec::ArrayVec;
 
 /// Defines a trait that chooses between libstd or libm implementations of float methods.
@@ -13,13 +23,22 @@ macro_rules! define_float_funcs {
         fn $name:ident(self $(,$arg:ident: $arg_ty:ty)*) -> $ret:ty
         => $lname:ident/$lfname:ident;
     )+) => {
+
+        /// Since core doesn't depend upon libm, this provides libm implementations
+        /// of float functions which are typically provided by the std library, when
+        /// the `std` feature is not enabled.
+        ///
+        /// For documentation see the respective functions in the std library.
         #[cfg(not(feature = "std"))]
-        pub(crate) trait FloatFuncs : Sized {
+        pub trait FloatFuncs : Sized + sealed::FloatFuncsSealed {
             /// Special implementation for signum, because libm doesn't have it.
             fn signum(self) -> Self;
 
             $(fn $name(self $(,$arg: $arg_ty)*) -> $ret;)+
         }
+
+        #[cfg(not(feature = "std"))]
+        impl sealed::FloatFuncsSealed for f32{}
 
         #[cfg(not(feature = "std"))]
         impl FloatFuncs for f32 {
@@ -41,6 +60,8 @@ macro_rules! define_float_funcs {
             })+
         }
 
+        #[cfg(not(feature = "std"))]
+        impl sealed::FloatFuncsSealed for f64{}
         #[cfg(not(feature = "std"))]
         impl FloatFuncs for f64 {
             #[inline]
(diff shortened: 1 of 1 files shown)


src/common.rs
```diff
@@ -5,6 +5,10 @@
 
 #![allow(missing_docs)]
 
+mod common_sealed {
```
DJMcNab, line 8: I'm not sure that `common_sealed` is a great name for this - I think I'd prefer something like:
```rust
mod sealed {
    /// Seals [super::FloatFunc]
    pub trait FloatFuncsSealed {}
}
```

I don't think the Rust project documents sealed traits itself, so I think we should also probably have a link which explains it.
I don't know what link that would be, however

src/common.rs
```diff
@@ -14,13 +18,16 @@ macro_rules! define_float_funcs {
         => $lname:ident/$lfname:ident;
     )+) => {
         #[cfg(not(feature = "std"))]
-        pub(crate) trait FloatFuncs : Sized {
+        pub trait FloatFuncs : Sized + common_sealed::Sealed {
```
DJMcNab, line 21: This should have some docs now that it's public

DJMcNab (commented):
I'm not a Kurbo maintainer, so I'm not certain that they would want this

src/common.rs
```diff
@@ -5,6 +5,10 @@
 
 #![allow(missing_docs)]
```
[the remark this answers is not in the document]
ratmice: It looks like this is here to avoid documenting the existing `pub const GAUSS_LEGENDRE_COEFFS`* values.
I added basically a stub documentation explaining the trait, and pointing to the std implementation for further docs.

src/common.rs
ratmice: fixed in [cb3c29d](https://github.com/linebender/kurbo/pull/332/commits/cb3c29d41a0163f3118954e4c0ef44a0b53023f2)

src/common.rs
ratmice: Should be fixed in [85ad5eb](https://github.com/linebender/kurbo/pull/332/commits/85ad5eb179ebcf2c6c28f775ab482a8143f223f8)

ratmice:
One thing this patch *doesn't* try to fix is `cargo test --no-default-features --features libm` I manually checked that the respective `build` target works, don't know see that this combination is actually being tested by ci though.

It is perhaps worth adding that.

src/common.rs
```diff
@@ -6,7 +6,9 @@
...
 
 mod sealed {
-    pub trait Sealed {}
+    /// A sealed trait which [sealed trait](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/) which stops this trait
```
DJMcNab, line 9: I think this reads:
A sealed trait which sealed trait which..., which doesn't seem right

ratmice: whoops.

derekdreery:
Hey @ratmice sorry if this is a silly question, but why do we need to seal the trait?

ratmice:
I just did it in the spirit of it being the more conservative change, i.e. it is easy to unseal it later, but sealing it would be a breaking change.

So honestly no *good* reason.

derekdreery:
> I just did it in the spirit of it being the more conservative change, i.e. it is easy to unseal it later, but sealing it would be a breaking change.

@ratmice Would you mind putting a comment on the sealing trait along these lines? Then future me knows we can remove the sealing if we need to :). Other than that LGTM.

src/common.rs
```diff
@@ -5,6 +5,13 @@
 
 #![allow(missing_docs)]
 
+mod sealed {
+    /// A [sealed trait](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/)
+    /// which stops this trait from being derived outside the library. This could be relaxed
+    /// in the future if there is are good reasons to allow outside impls.
```
DJMcNab, line 11:
```suggestion
    /// A [sealed trait](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/)
    /// which stops [super::FloatFuncs] from being implemented outside kurbo. This could be relaxed
    /// in the future if there are good reasons to allow outside impls.
```
I'd also probably change 'outside' to 'external' in the final line

ratmice: Should be fixed in [4684dac](https://github.com/linebender/kurbo/pull/332/commits/4684dacad8fd2fb5929a510364c9196ae5c7ec32)

ratmice: I think this link fails with clippy because of the `#[cfg(not(features="std")))]` conditional compilation?

ratmice:
If it's alright i'll squash/force push this all down to one commit with the initial commit message.

DJMcNab:
That would be alright, although that will also happen as part of the merge process

derekdreery (approved):
LGTM

ratmice:
I'm happy with it as is, and ran the peniko patch on top of this revision through it's CI.

I don't believe anything in a peniko review would require changes here, but feel like I
should wait for at least a cursory a review there until landing this here.

Issue #218, opened 2022-03-20
mindist doesn't actually work for lines


simoncozens:
I'm sorry. I screwed up. It seemed like the Chen algorithm I used for minimum distance between two Bezier segments worked for any parametric curve, including lines. It turns out it does not. Here's a test case:

```rust
    #[test]
    fn test_line_curve() {
        let line = PathSeg::Line(Line::new((929.0, 335.0), (911.0, 340.0)));

        let line_as_bez = PathSeg::Cubic(CubicBez::new(
            line.eval(0.0),
            line.eval(1.0 / 3.0),
            line.eval(2.0 / 3.0),
            line.eval(1.0),
        ));

        let bez2 = PathSeg::Cubic(CubicBez::new(
            (1052.0, 401.0),
            (1048.0, 305.0),
            (1046.0, 216.0),
            (1054.0, 146.0),
        ));
        let mindist_as_bez = line_as_bez.min_dist(bez2, 0.5);
        let mindist_as_line = line.min_dist(bez2, 0.5);
        println!(
            "Distance if we pretend it's a bezier: {:?}",
            mindist_as_bez.distance
        );
        println!(
            "Distance if we use the line: {:?}",
            mindist_as_line.distance
        );
        assert!((mindist_as_line.distance - mindist_as_bez.distance).abs() < 0.5);
    }
```

I don't have a fix for this; you could either raise the degree of the line, or use a separate algorithm for the line/curve distance case.

simoncozens:
This is still broken and the test case still fails.

derekdreery:
So are you saying that the above test case should be added to kurbo and that it should pass?

simoncozens:
Yes, exactly. The test case is demonstrating a current bug; we should make it pass.

platlas:
The paper mentions in Property 2:  

> Similarly, if there exists an integer k = 0 or k = 2m, such that for all i ∈ {0, 1, . . . , 2n} and j ∈ {0, 1, . . . , 2m}, Di,j ≥ Di,k, then S(u, v) must reach the minimum value at the boundary v = 0 or v = 1, which means the end-point Q0 or Qm is the nearest point on C2(v).

If I change `n` to `m` here:
https://github.com/linebender/kurbo/blob/6ea9f53498749703606f6f8d019fe0bc9080ea30/src/mindist.rs#L88  

That test will pass.

PR #334, opened 2024-02-12
Fix edge case for `mindist`

platlas:
This is fix for  https://github.com/linebender/kurbo/issues/218.  

Depends on orientation of curves, function fails if curves have different order.  Check, detecting end-points as closest points, returns false positives here. This is caused by small typo in implementation where check takes order of first curve instead of order from second curve.  

PR consist of test case from https://github.com/linebender/kurbo/issues/218 and replacement of  `n` order to `m` in that specific check.

### DIFF
--- a/src/mindist.rs
+++ b/src/mindist.rs
@@ -85,7 +85,7 @@ pub(crate) fn min_dist_param(
             if dij < dkj {
                 at_boundary0on_bez2 = false
             }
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
I didn't analyze the logic deeply, but I agree this is a typo, and the test case seems to cover it. Thanks!

(I don't remember if you have commit access, but if not, let me know. We have a policy of author-commits)

PR #340, opened 2024-03-07
Move `Self: Sized` bound from `Shape` to methods.

waywardmonkeys:
The `Shape` trait overall doesn't need the `Self: Sized` bound only 2 methods do, so move the bound there.


waywardmonkeys:
I'd found this when I was looking at why `Shape` isn't object-safe, which was causing me problems when looking at FFI for Vello. This alone won't solve that, but seemed like a change in a positive direction.

PR #341, opened 2024-03-07
clippy: Fix semicolon_if_nothing_returned lints, enable it.

waywardmonkeys:


raphlinus (approved):
Yes, good style. There were a lot more of these than I would have guessed. Thanks!

PR #343, opened 2024-03-13
Allow lines in simplify input

### LINKED ISSUE #342: `simplify_bezpath` panics on `LineTo`
I'm attempting to use `simplify_bezpath` to approximate a large number of tiny line segments (generated by a tool I can't modify) with a small number of beziers for easier laser-cutting. My input consists of the concatenation of some number of sequences of `PathEl` where the first is a `MoveTo` and the remainder are all `LineTo`. As soon as `simplify_bezpath` encounters the first `LineTo`, it panics in `SimplifyState::add_seg`:

https://github.com/linebender/kurbo/blob/3ea8f6bf4c18688ce82bb7d105e962b8f204a484/src/simplify.rs#L250-L259

It's not clear why this is believed to be unreachable. Is my input malformed somehow?

raphlinus:
This patch allows lines in paths to be simplified, which previously caused a panic due to an unhandled case. It also passes them through unmodified if the angle exceeds the threshold.

The docstring also adds a little discussion about expectations with noisy input.

Fixes #342

### DIFF
--- a/src/simplify.rs
+++ b/src/simplify.rs
@@ -252,9 +252,9 @@ impl SimplifyState {
             self.queue.move_to(seg.start());
         }
         match seg {
+            PathSeg::Line(l) => self.queue.line_to(l.p1),
             PathSeg::Quad(q) => self.queue.quad_to(q.p1, q.p2),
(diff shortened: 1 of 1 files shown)


platlas (approved):
Looks good. I also checked the one segment scenario.

PR #346, opened 2024-05-20
lints: Enable `warn(unused_qualifications)`

waywardmonkeys:
This helps point out where the code can be more concise with an identifier already being in scope and not needing qualifications.


DJMcNab (approved):
I don't really maintain Kurbo, but I can't see any way this causes a problem - presuming our CI is checking MSRV properly, as finding out the MSRV of a lint requires more research than a single search.

waywardmonkeys:
For reference, it existed earlier than December, 2021, so I know it is fine here. :)

PR #347, opened 2024-05-25
Add the `Rect::overlaps`  and `Rect::contains_rect` methods

nils-mathieu:
I stumbled on this while toying with the `vello` rendering engine (which is great btw, you guys are doing god's work). I'm trying to create a simple audio sequencer with it and while trying to determine whether shapes were visible on screen, I found that there was no way of checking for the intersection of two `Rect`s.

Right now, the only way to do that is to compute their intersection, then verify whether it's empty or not.

```Rust
let rect1 = Rect::new(0.0, 0.0, 10.0, 10.0);
let rect2 = Rect::new(5.0, 5.0, 15.0, 15.0);
assert!(!rect1.intersect(rect2).is_empty());
```

This is not ideal for two reasons:

1. It's not very ergonomic. Checking whether two rectangles overlap should probably not involve that intermediary step.
2. This is doing a bunch of wasted work to find out what the intersection is when we only care about whether it's empty or not.

This pull request adds `Rect::overlaps` and `Rect::contains_rect`, which do exactly that.

```Rust
let rect1 = Rect::new(0.0, 0.0, 10.0, 10.0);
let rect2 = Rect::new(5.0, 5.0, 15.0, 15.0);
assert!(rect1.overlaps(rect2));
assert!(!rect1.contains_rect(rect2));
```


src/rect.rs
```diff
@@ -768,4 +824,54 @@ mod tests {
         let test = Rect::new(0.0, 0.0, 1.0, 1.0);
         assert!((test.aspect_ratio() - 1.0).abs() < 1e-6);
     }
+
+    #[test]
```
derekdreery, line 828: Maybe consider adding a test for (0, 0, 10, 10), (0, 10, 10, 20) (to demonstrate that rects overlap if they share a boundary.

nils-mathieu:
My mistake, I compltely forgot about this. I just commited the suggested modifications.

waywardmonkeys:
@nils-mathieu  Could you please rebase this on top of current `main` to pick up the CI changes / updates? Thanks!

nils-mathieu:
Github's autosync merged rather than rebasing without asking, I hope it's okay.
Seems green though x)

waywardmonkeys:
One last thing (from me): Could you add an entry to `CHANGELOG.md`? Just use the existing format, add the PR number sorted numerically and your GitHub user name sorted alphabetically and an entry for this under 'Added'.

raphlinus (approved):
I'm going to approve this, as the methods seem useful, but I will point out that `overlaps` is a weaker predicate than the question of whether there exists a point that is inside both rectangles, as the latter is defined as x0 <= x < x1. The documentation is correct, and, while I considered another name like "overlaps_inclusive," I think that's clunky and not worth it.

Thanks for the PR and sorry for the delay.

nils-mathieu:
done!

waywardmonkeys:
Thanks everyone!

PR #350, opened 2024-05-29
add triangle shape

juliapaci:
Implements (at least starts to) a triangle shape, im not sure if this is useful enough to merge though.

Also while looking at the source i noticed that that places like [here](https://github.com/linebender/kurbo/blob/main/src/rect.rs#L43) use `Rect` instead of `Self` unlike other parts of the codebase. And some parts of the documentation uses "\[\`Type\`\]" and other parts just use "Type". Wasnt sure to make an issue or not but should there be a cleanup of the codebase?

also there is alot of conversion between `Vec2` and `Point` right now. is there a better way to handle these situations where i want to use Vec2 methods on a Point but will end up converting right back to a Point afterwards?

triangles have alot of opportunities for more specific methods so its worth noting that this can be heavily expanded upon which could come later. Aswell as this, i have made some decisions like using centroids as the origin of the triangle (https://web.archive.org/web/20131104015950/http://www.jimloy.com/geometry/centers.htm) which might not be the best? let me know.

ill continue working on this, specifically the TODOs (see inlined todos)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,6 +110,7 @@ mod size;
 mod stroke;
 mod svg;
 mod translate_scale;
+mod triangle;
 mod vec2;
 
 pub use crate::affine::Affine;
@@ -143,4 +144,5 @@ pub use crate::stroke::{
 };
 pub use crate::svg::{SvgArc, SvgParseError};
 pub use crate::translate_scale::TranslateScale;
+pub use crate::triangle::{Triangle, TrianglePathIter};
 pub use crate::vec2::Vec2;
--- a/src/point.rs
+++ b/src/point.rs
@@ -3,6 +3,7 @@
 
 //! A 2D point.
 
+use core::cmp::Ordering;
 use core::fmt;
 use core::ops::{Add, AddAssign, Sub, SubAssign};
 
@@ -201,6 +202,30 @@ impl Point {
     #[inline]
     pub fn is_nan(self) -> bool {
         self.x.is_nan() || self.y.is_nan()
+    }
+
+    /// The topmost [`Point`]
+    #[inline]
+    pub fn topmost(&self, other: &Self) -> Ordering {
+        self.y.total_cmp(&other.y)
+    }
+
+    /// The bottommost [`Point`]
+    #[inline]
+    pub fn bottommost(&self, other: &Self) -> Ordering {
+        other.y.total_cmp(&self.y)
+    }
+
+    /// The rightmost [`Point`]
+    #[inline]
+    pub fn rightmost(&self, other: &Self) -> Ordering {
+        self.x.total_cmp(&other.x)
+    }
+
+    /// The leftmost [`Point`]
+    #[inline]
+    pub fn leftmost(&self, other: &Self) -> Ordering {
+        other.x.total_cmp(&self.x)
     }
 }
 
--- a/src/point.rs
+++ b/src/point.rs
@@ -204,29 +204,6 @@ impl Point {
         self.x.is_nan() || self.y.is_nan()
     }
 
-    /// The topmost [`Point`]
-    #[inline]
-    pub fn topmost(&self, other: &Self) -> Ordering {
-        self.y.total_cmp(&other.y)
-    }
-
-    /// The bottommost [`Point`]
-    #[inline]
-    pub fn bottommost(&self, other: &Self) -> Ordering {
-        other.y.total_cmp(&self.y)
-    }
-
-    /// The rightmost [`Point`]
-    #[inline]
-    pub fn rightmost(&self, other: &Self) -> Ordering {
-        self.x.total_cmp(&other.x)
-    }
-
-    /// The leftmost [`Point`]
-    #[inline]
-    pub fn leftmost(&self, other: &Self) -> Ordering {
-        other.x.total_cmp(&self.x)
-    }
 }
 
 impl From<(f32, f32)> for Point {
--- a/src/point.rs
+++ b/src/point.rs
@@ -3,7 +3,6 @@
 
 //! A 2D point.
 
-use core::cmp::Ordering;
 use core::fmt;
 use core::ops::{Add, AddAssign, Sub, SubAssign};
 
--- a/src/point.rs
+++ b/src/point.rs
@@ -202,7 +202,6 @@ impl Point {
     pub fn is_nan(self) -> bool {
         self.x.is_nan() || self.y.is_nan()
     }
-
 }
 
 impl From<(f32, f32)> for Point {
--- a/src/triangle.rs
+++ b/src/triangle.rs
@@ -119,7 +119,9 @@ impl Triangle {
         self.area() == 0.0
     }
 
-    /// Inradius of [`Triangle`] (the greatest radius of a circle that is within the [`Triangle`])
+    /// Inradius of [`Triangle`] 
+    /// 
+    /// This is defined as the greatest radius of a circle that is within the [`Triangle`] 
     /// with center [`Triangle::circumcenter`]
     #[inline]
     pub fn inradius(&self) -> f64 {
@@ -283,7 +285,7 @@ mod tests {
     #[test]
     fn centroid() {
         let test = Triangle::from_coords((-90.02, 3.5), (7.2, -9.3), (8.0, 9.1)).centroid();
-        let expected = Point::new(-24.939999999999998, 1.0999999999999996);
+        let expected = Point::new(-24.94, 1.1);
 
         assert_eq!(test, expected);
     }
@@ -292,9 +294,9 @@ mod tests {
     fn offsets() {
         let test = Triangle::from_coords((-20.0, 180.2), (1.2, 0.0), (290.0, 100.0)).offsets();
         let expected = [
-            Vec2::new(-110.39999999999999, 86.8),
-            Vec2::new(-89.19999999999999, -93.39999999999999),
-            Vec2::new(199.60000000000002, 6.6000000000000085),
+            Vec2::new(-110.4, 86.8),
+            Vec2::new(-89.2, -93.4),
+            Vec2::new(199.6, 6.6),
         ];
 
         assert_eq!(test, expected);
--- a/src/triangle.rs
+++ b/src/triangle.rs
@@ -2,10 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 //! Triangle shape
-use crate::{Ellipse, PathEl, Point, Rect, Shape, Vec2};
+use crate::{Ellipse, PathEl, Point, Rect, Shape, Size, Vec2};
 
 use core::f64::consts::FRAC_PI_3;
 use core::ops::{Add, Sub};
+use std::cmp::*;
 
 #[cfg(not(feature = "std"))]
 use crate::common::FloatFuncs;
@@ -17,7 +18,7 @@ use crate::common::FloatFuncs;
 //   /   \
 //  *-----*
 //  B     C
-#[derive(Clone, Copy, PartialEq)]
+#[derive(Clone, Copy, PartialEq, Debug)]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Triangle {
@@ -62,29 +63,45 @@ impl Triangle {
         }
     }
 
-    /// A new `Triangle` from centroid and sizes
+    /// A new [`Triangle`] from centroid and sizes ([`Size`]) - offset of each vertex from the centroid
     #[inline]
-    pub fn from_centroid_sizes(centroid: impl Into<Point>, sizes: [f64; 3]) -> Self {
+    pub fn from_centroid_sizes(centroid: impl Into<Point>, sizes: [impl Into<Size>; 3]) -> Self {
         let centroid = centroid.into();
+
+        let mut points: [Point; 3] = Default::default();
+        for (i, size) in sizes.into_iter().enumerate() {
+            points[i] = centroid + size.into().to_vec2();
+        }
+
+        Self::new(points[0], points[1], points[2])
+    }
+
+    /// A new [`Triangle`] with vertex distances of `distances` from the `centroid`
+    ///
+    /// NOTE: the new [`Triangle`] does not keep `centroid`
+    #[inline]
+    pub fn from_centroid_distances(centroid: impl Into<Point>, distances: [f64; 3]) -> Self {
+        let centroid = centroid.into();
+
         Self::new(
-            (centroid.x, centroid.y + sizes[0]),
-            centroid + Vec2::IDENTITY * sizes[1],
-            centroid + Vec2::IDENTITY * sizes[2],
+            centroid + (0.0, distances[0]),
+            centroid - Vec2::ONE * distances[1] / 2.0,
+            centroid + Vec2::ONE * distances[2] / 2.0,
         )
     }
 
     /// A new equilateral [`Triangle`]
     /// takes the center and a point equidistant to the midpoints of the vertices (radius)
     #[inline]
-    pub fn from_center_size(center: impl Into<Point>, radius: f64) -> Self {
+    pub fn from_center_radius(center: impl Into<Point>, radius: f64) -> Self {
         const THETA: f64 = FRAC_PI_3; // equilateral triangle guarantee
         let center = center.into().to_vec2();
 
         let h = (radius / ((THETA / 2.0).tan()) * radius / ((THETA / 2.0).tan()) + radius * radius)
             .sqrt();
         let a = center.y + radius;
-        let b = center + Vec2::IDENTITY * h;
-        let c = center - Vec2::IDENTITY * h;
+        let b = center + Vec2::ONE * h;
+        let c = center - Vec2::ONE * h;
 
         Self::new((center.x, a), b.to_point(), c.to_point())
     }
@@ -95,10 +112,16 @@ impl Triangle {
         Self::from_centroid_sizes(centroid, self.sizes())
     }
 
-    /// Creates a new [`Triangle`] with new `sizes`
+    /// Creates a new [`Triangle`] with `sizes`
     #[inline]
-    pub fn with_sizes(self, sizes: [f64; 3]) -> Self {
+    pub fn with_sizes(self, sizes: [impl Into<Size>; 3]) -> Self {
         Self::from_centroid_sizes(self.centroid(), sizes)
+    }
+
+    /// Creates a new [`Triangle`] with each vertice's `distances` offset from centroid
+    #[inline]
+    pub fn with_distances(self, distances: [f64; 3]) -> Self {
+        Self::from_centroid_distances(self.centroid(), distances)
     }
 
     /// The centroid of the [`Triangle`]
@@ -110,15 +133,27 @@ impl Triangle {
         )
     }
 
-    /// The distance of each vertex from the centroid
+    /// The euclidean distance of each vertex from the centroid
     #[inline]
-    pub fn sizes(&self) -> [f64; 3] {
+    pub fn distance(&self) -> [f64; 3] {
         let centroid = self.centroid();
 
         [
             self.a.distance(centroid),
             self.b.distance(centroid),
             self.c.distance(centroid),
+        ]
+    }
+
+    /// The offset of each vertex from the centroid
+    #[inline]
+    pub fn sizes(&self) -> [Size; 3] {
+        let centroid = self.centroid().to_vec2();
+
+        [
+            (centroid - self.a.to_vec2()).to_size(),
+            (centroid - self.b.to_vec2()).to_size(),
+            (centroid - self.c.to_vec2()).to_size(),
         ]
     }
 
@@ -143,6 +178,65 @@ impl Triangle {
     pub fn right_angled_area(&self) -> f64 {
         // A = 1/2*b*h
         (self.b.distance(self.c) * self.b.midpoint(self.c).distance(self.a)) / 2.0
+    }
+
+    /// Updates [`Triangle`]'s vertice positions
+    /// such that [`Triangle::a`] is topmost, [`Triangle::b`] is leftmost, and [`Triangle::c`] is rightmost
+    #[inline]
+    pub fn organise(self) -> Self {
+        Self::new(self.topmost(), self.leftmost(), self.rightmost())
+    }
+
+    /// Vertex coordinate ([`Point`]) of [`Triangle`] with `x` ordinate
+    #[inline]
+    pub fn vertex_from_x(&self, x: f64) -> Option<Point> {
+        self.as_array().iter().find(|&v| v.x == x).copied()
+    }
+
+    /// Vertex coordinate ([`Point`]) of [`Triangle`] with `y` ordinate
+    #[inline]
+    pub fn vertex_from_y(&self, y: f64) -> Option<Point> {
+        self.as_array().iter().find(|&v| v.y == y).copied()
+    }
+
+    /// The topmost vertex of `self` as a [`Point`]
+    #[inline]
+    pub fn topmost(&self) -> Point {
+        *self
+            .as_array()
+            .iter()
+            .max_by(|x, y| x.y.total_cmp(&y.y))
+            .unwrap()
+    }
+
+    /// The bottommost vertex of `self` as a [`Point`]
+    #[inline]
+    pub fn bottommost(&self) -> Point {
+        *self
+            .as_array()
+            .iter()
+            .max_by(|x, y| y.y.total_cmp(&x.y))
+            .unwrap()
+    }
+
+    /// The rightmost vertex of `self` as a [`Point`]
+    #[inline]
+    pub fn rightmost(&self) -> Point {
+        *self
+            .as_array()
+            .iter()
+            .max_by(|x, y| x.x.total_cmp(&y.x))
+            .unwrap()
+    }
+
+    /// The leftmost vertex of `self` as a [`Point`]
+    #[inline]
+    pub fn leftmost(&self) -> Point {
+        *self
+            .as_array()
+            .iter()
+            .max_by(|x, y| y.x.total_cmp(&x.x))
+            .unwrap()
     }
 
     /// Maximum x-coordinate of the [`Triangle`]'s vertices
@@ -208,9 +302,9 @@ impl Triangle {
     /// Expand the triangle by a constant amount (`sizes`) in all directions
     pub fn inflate(&self, sizes: [f64; 3]) -> Self {
         Self::new(
-            (self.a.to_vec2() + Vec2::IDENTITY * sizes[0]).to_point(),
-            (self.b.to_vec2() + Vec2::IDENTITY * sizes[1]).to_point(),
-            (self.c.to_vec2() + Vec2::IDENTITY * sizes[2]).to_point(),
+            (self.a.to_vec2() + Vec2::ONE * sizes[0]).to_point(),
+            (self.b.to_vec2() + Vec2::ONE * sizes[1]).to_point(),
+            (self.c.to_vec2() + Vec2::ONE * sizes[2]).to_point(),
         )
     }
 
@@ -279,6 +373,12 @@ impl Triangle {
     pub fn is_nan(&self) -> bool {
         self.a.is_nan() || self.b.is_nan() || self.c.is_nan()
     }
+
+    /// [`Triangle`]'s vertices as an array of Points
+    #[inline]
+    pub fn as_array(&self) -> [Point; 3] {
+        [self.a, self.b, self.c]
+    }
 }
 
 impl From<(Point, Point, Point)> for Triangle {
@@ -289,7 +389,7 @@ impl From<(Point, Point, Point)> for Triangle {
 
 impl From<(Point, f64)> for Triangle {
     fn from(params: (Point, f64)) -> Triangle {
-        Triangle::from_center_size(params.0, params.1)
+        Triangle::from_center_radius(params.0, params.1)
     }
 }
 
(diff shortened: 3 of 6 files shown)


aymey:
Should we add an `Insets` for triangle?

src/triangle.rs
```diff
@@ -0,0 +1,401 @@
...
+        Self::from_centroid_sizes(self.centroid(), sizes)
+    }
+
+    /// The centroid of the [`Triangle`]
+    #[inline]
+    pub fn centroid(&self) -> Point {
+        Point::new(
+            (self.a.x + self.b.x + self.c.x) / 3.0,
+            (self.a.y + self.b.y + self.c.y) / 3.0,
+        )
```
aymey, line 110: One of those situations where we could do this or:

```rust
((self.a.to_vec2() + self.b.to_vec2() + self.c.to_vec2()) / 3.0).to_point()
```

aymey:
i realise that changes to other source files like [vec2.rs](https://github.com/linebender/kurbo/blob/main/src/vec2.rs) should be another pr but right no this pr depends on them so ill keep it here for no. if this doesn't get merged then ill split my fork and make a pr just for those other source files.

src/triangle.rs
```diff
@@ -0,0 +1,537 @@
...
+    #[inline]
+    pub fn right_angled_area(&self) -> f64 {
+        // A = 1/2*b*h
+        (self.b.distance(self.c) * self.b.midpoint(self.c).distance(self.a)) / 2.0
+    }
+
+    /// Updates [`Triangle`]'s vertice positions
+    /// such that [`Triangle::a`] is topmost, [`Triangle::b`] is leftmost, and [`Triangle::c`] is rightmost
+    #[inline]
+    pub fn organise(self) -> Self {
```
aymey, line 186: should British spelling be changed? (probably)

src/point.rs
```diff
@@ -185,6 +186,30 @@ impl Point {
...
+    #[inline]
+    pub fn rightmost(&self, other: &Self) -> Ordering {
+        self.x.total_cmp(&other.x)
+    }
+
+    /// The leftmost [`Point`]
+    #[inline]
+    pub fn leftmost(&self, other: &Self) -> Ordering {
+        other.x.total_cmp(&self.x)
+    }
```
aymey, line 212: Is there a more standard way for this type of comparison other than just implementing a function? Maybe we should introduce a trait that size and vec2 can also benefit from?

src/triangle.rs
```diff
@@ -174,17 +189,21 @@ impl Triangle {
...
 
     /// Updates [`Triangle`]'s vertice positions
     /// such that [`Triangle::a`] is topmost, [`Triangle::b`] is leftmost, and [`Triangle::c`] is rightmost
     #[inline]
     pub fn organise(self) -> Self {
-        Self::new(self.topmost(), self.leftmost(), self.rightmost())
+        let t = Self::new(self.topmost(), self.leftmost(), self.rightmost());
+
+        t
     }
```
aymey: (ignoring the bug which will be fixed, im just trying to find a clean way) there are a few issues / design choices that can be made here:
- should organise() mutate `self`?
- a lot of other functions assume an organised position so should we either let the user call organised before these functions OR call organise in these functions

src/triangle.rs
```diff
@@ -151,12 +160,18 @@ impl Triangle {
...
+            (self.c.to_vec2() - centroid).to_size(),
         ]
     }
 
+    /// The perimeter of the [`Triangle`]
+    #[inline]
+    fn perimeter(&self) -> f64 {
+        self.a.distance(self.b) + self.b.distance(self.c) + self.c.distance(self.a)
+    }
+
```
aymey, line 185: we have general `perimeter` and `area` functions but we also have specific ones for right angled triangles. should we add more specific ones for equilateral triangles and stuff or is it too much clutter?

also, i dont see the need to implement perimeter and area here when its in the `Shape` trait? this is also done for `Rect`.

src/triangle.rs
```diff
@@ -0,0 +1,647 @@
...
+    #[inline]
+    pub fn from_centroid_distances(centroid: impl Into<Point>, distances: [f64; 3]) -> Self {
+        let centroid = centroid.into();
+
+        Self::new(
+            centroid + (0.0, distances[0]),
+            centroid + distances[1] * Vec2::from_angle(5.0 * FRAC_PI_4),
+            centroid + distances[2] * Vec2::from_angle(7.0 * FRAC_PI_4),
+        )
+    }
```
aymey, line 100: maybe a transition from `from*` to `with*` like in the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) is better. although im hesitant to do this here because this would require a change for the entire codebase so maybe should be another pr?

aymey: we also wouldnt have to include the `centroid` param this way since users could chain with `with_centroid()`

src/triangle.rs
```diff
@@ -0,0 +1,651 @@
...
+    #[inline]
+    pub fn organise(self) -> Self {
+        let mut points = self.as_array();
+        points.sort_by(|x, y| x.rightmost(y));
+
+        let a = self.topmost();
+        let b = *points.iter().find(|&&p| p != a).unwrap();
+        let c = *points.iter().rev().find(|&&p| p != a).unwrap();
+        Self::new(a, b, c)
+    }
```
aymey, line 222: we probably can do a better way, we cant just do [this](https://github.com/linebender/kurbo/pull/350/commits/be6ed8d8792555e55d8baf1f9c15b9fcdee31e25#diff-660e9939fa0efec5379f87f73cb3c2aa5f9d70191d8a50ec720441c817a07a7cR204) though because of double ups.

aymey:
and with 8b03193f6f932807b723c9d693a4010ba5e5076d this pr should be ready for a review!

src/triangle.rs
```diff
@@ -0,0 +1,651 @@
...
+
+    #[test]
+    fn area() {
+        let test = Triangle::new(
+            (12123.423, 2382.7834),
+            (7892.729, 238.459),
+            (7820.2, 712.23),
+        )
+        .area();
+        let expected = 1079952.91574081;
```
aymey, line 613: dumb floating point nonsense. tests dont pass because of this, how should we handle it? other places use `assert_approx_eq()`

aymey:
on a side note i also think it would be useful to have `from/to_triangle/circle/rect()`, cause we do it for `Ellipse` already we may aswell do it for all other shapes which would also be super easy to implement. prob should be another pr tho

src/triangle.rs
```diff
@@ -0,0 +1,655 @@
...
+        self.as_array().iter().find(|&v| v.x == x).copied()
+    }
+
+    /// Vertex coordinate ([`Point`]) of [`Triangle`] with `y` ordinate
+    #[inline]
+    pub fn vertex_from_y(&self, y: f64) -> Option<Point> {
+        self.as_array().iter().find(|&v| v.y == y).copied()
+    }
+
+    /// The topmost vertex of `self` as a [`Point`]
```
raphlinus, line 236: I don't think these are justified, I think it's fairly specialized and users can write it themselves. In addition, there's a naming problem, as "topmost" implies a y-down coordinate system, and one of the conceits of kurbo is that it's agnostic to whether the coordinates are y-up or y-down.

src/triangle.rs
```diff
@@ -0,0 +1,655 @@
...
+    #[inline]
+    pub fn scale(self, factor: f64) -> Self {
+        Self::new(
+            (self.a.to_vec2() * factor).to_point(),
+            (self.b.to_vec2() * factor).to_point(),
+            (self.c.to_vec2() * factor).to_point(),
+        )
+    }
+
+    /// Returns the [`Ellipse`] that is bounded by this [`Triangle`].
```
raphlinus, line 380: As mentioned elsewhere, I think incircle and circumcircle are the methods that should be included.

src/triangle.rs
```diff
@@ -0,0 +1,655 @@
...
+    }
+
+    /// [`Triangle`]'s vertices as an array of Points
+    #[inline]
+    pub fn as_array(&self) -> [Point; 3] {
+        [self.a, self.b, self.c]
+    }
+}
+
+impl From<(Point, Point, Point)> for Triangle {
```
raphlinus, line 405: There is no corresponding conversion for `Line`, but maybe this is something that should be revisited.

src/point.rs
```diff
@@ -185,6 +186,30 @@ impl Point {
     pub fn is_nan(self) -> bool {
         self.x.is_nan() || self.y.is_nan()
     }
+
```
raphlinus, line 189: I don't think these belong in the public API. There's also a naming problem, the names suggest that the return value is a `Point` (and the incorrect docstring agrees).

raphlinus (commented):
Thanks for the PR, and apologies for taking so long to review. Triangle is a beautiful shape with useful special properties, and this PR is a good start. Some general thoughts.

A very large number of items are marked as "not justified." Basically, we want methods that are used by potentially a broad range of applications. If it's only going to be used once, or for a test, it doesn't belong in the public API. It's very easy to add methods, but unfortunately extremely difficult to remove them. In addition, a theme in kurbo is mathematical correctness, which means that some methods which might be convenient (like `Point + Point` or, recently `Affine * Vec2`) aren't present. So most methods that aren't present for other shapes are out. That said, when there's something really special to triangles, it makes sense. I like centroid, and I'm also open to incircle (and circumcircle). For other methods, I'm open to an argument why it should be included.

There should be an `impl Mul<Triangle> for Affine`. Fortunately, it's simple, you can pretty much cut'n'paste the one for line.

Looking forward to the revisions, then I'll do another review.

src/triangle.rs
aymey: Oh right, their useless now anyway because the `organise` is useless

src/triangle.rs
```diff
@@ -0,0 +1,655 @@
...
+        // semi perimeter
+        let s = (ab + bc + ac) / 2.0;
+
+        // Heron's formula
+        let area = (s * (s - ab) * (s - bc) * (s - ac)).sqrt();
+
+        area / s
+    }
+
+    /// Expand the triangle by a constant amount (`sizes`) in all directions
```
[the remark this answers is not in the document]
aymey: Oops i forgot to update this. is [4d8f906](https://github.com/linebender/kurbo/pull/350/commits/4d8f906af02ebe9bdff84ffe75b35161da5e736f) `inflate` fine?

src/triangle.rs
```diff
@@ -0,0 +1,655 @@
...
+        let size = size.into();
+
+        Self::new(
+            (self.a.to_vec2() + size.to_vec2()).to_point(),
+            (self.b.to_vec2() + size.to_vec2()).to_point(),
+            (self.c.to_vec2() + size.to_vec2()).to_point(),
+        )
+    }
+
+    /// A new [`Triangle`] with each vertecie's ordinates rounded to the nearest integer
```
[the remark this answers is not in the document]
aymey: me neither, i just took alot of these helpers from rect

aymey (commented):
[4d8f906](https://github.com/linebender/kurbo/pull/350/commits/4d8f906af02ebe9bdff84ffe75b35161da5e736f)

ill update the tests once you check it out

aymey:
Ive double checked and it looks like my `incircle` and `circumcircle` implementations are correct yet when i check them in desmos.com the incircle radius is too small and the circumcircle radius is too large. Have i done something wrong?

also can the `circumcenter` function be more efficient?

src/triangle.rs
aymey: I was gonna do something like this for `Line`:

```Rust
impl From<(Line, Line)> for Triangle {
    fn from(lines: (Line, Line)) -> Triangle {
        Triangle::new(lines.0.p0, lines.0.p0, lines.1.p0)
    }
}
```

but this just ignores the annoying cases like if the two lines dont meet at a point

DJMcNab: This is a different question. I believe that Raph is referring to the api:
```rust
impl From<(Point, Point)> for Line {...}
```

aymey: Oh right that makes a bit more sense, sorry. This is already implemented for rect which is why i did it here for triangle, I feel like it should therefore be implemented for `Line` to keep some continuity

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+    #[inline]
+    pub const fn from_coords(a: (f64, f64), b: (f64, f64), c: (f64, f64)) -> Self {
+        Self {
+            a: Point::new(a.0, a.1),
+            b: Point::new(b.0, b.1),
+            c: Point::new(c.0, c.1),
+        }
+    }
+
+    /// A new [`Triangle`] from centroid and offsets ([`Vec2`]) - offset of each vertex from the centroid
```
raphlinus, line 70: I think this function is no longer used.

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+            (self.a.to_vec2() - centroid),
+            (self.b.to_vec2() - centroid),
+            (self.c.to_vec2() - centroid),
+        ]
+    }
+
+    /// The area of the [`Triangle`]
+    #[inline]
+    pub fn area(&self) -> f64 {
+        0.5 * (self.b - self.a).cross(self.c - self.a).abs()
```
raphlinus, line 125: I disagree that there should be an `abs` here. We default to signed area in kurbo, and by taking the absolute value we'll have situations where the return value of this function doesn't match the area of converting into shape first.

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Whether this [`Triangle`] has no (zero) area
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.area() == 0.0
+    }
+
+    /// The greatest radius of a circle that is within the [`Triangle`]
+    /// with center [`Triangle::circumcenter`]
+    #[inline]
+    pub fn incircle(&self) -> f64 {
```
raphlinus, line 137: This function isn't named well. I think incircle which returns a circle would be useful (and would subsume this function), but if this function is to be kept, it should be named "inradius".

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+        let ab = self.a.distance(self.b);
+        let bc = self.b.distance(self.c);
+        let ac = self.a.distance(self.c);
+
+        2.0 * self.area() / (ab + bc + ac)
+    }
+
+    /// Circumcircle of the [`Triangle`]
+    /// with center [`Triangle::circumcenter`]
+    #[inline]
```
raphlinus, line 147: Similar naming concern as with incircle.

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub fn circumcircle(&self) -> f64 {
+        let ab = self.a.distance(self.b);
+        let bc = self.b.distance(self.c);
+        let ac = self.a.distance(self.c);
+
+        (ab * bc * ac) / (4.0 * self.area())
+        // ab * bc * ac / (((ab + bc + ac) * (bc + ac - ab) * (ab + ac - bc) * (ab + bc - ac)).sqrt())
```
raphlinus, line 154: The comment should probably be removed.

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+impl Add<Vec2> for Triangle {
+    type Output = Triangle;
+
+    #[inline]
+    fn add(self, v: Vec2) -> Triangle {
+        Triangle::new(
+            (self.a.to_vec2() + v).to_point(),
```
raphlinus, line 199: You can just say `self.a + v` here, you don't need the conversions to vec2.

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+impl Sub<Vec2> for Triangle {
+    type Output = Triangle;
+
+    #[inline]
+    fn sub(self, v: Vec2) -> Triangle {
+        Triangle::new(
+            (self.a.to_vec2() - v).to_point(),
```
raphlinus, line 212: Same as for add.

src/triangle.rs
```diff
@@ -0,0 +1,356 @@
...
+        .area();
+        let expected = 1079952.91574081;
+
+        assert_approx_eq(test, expected);
+    }
+
+    #[test]
+    fn circumcenter() {
+        let test = Triangle::EQUILATERAL.circumcenter();
+        let expected = Point::new(0.5, 0.2886751345948128);
```
raphlinus, line 336: This should be approx equal, otherwise the test is fragile. I see it's fine for incircle.

raphlinus (commented):
This is looking pretty good, thanks for all the progress. And apologies again for the slow turnaround on review, I'm juggling too many things. I think not much more is needed to land it.

src/triangle.rs
aymey: not exactly sure the use of this cause i dont know of any benefits for signed area but ill just do it if its the norm.

should the area test be updated to be negative now?

raphlinus: Yes, or permute the order of vertices. Converting to shape and evaluating area gives a negative result, so that's confirmation that the sign conventions are consistent. And in fact, you can make a good argument that this should be tested explicitly.

aymey: > Converting to shape and evaluating area gives a negative result, so that's confirmation that the sign conventions are consistent

but doesn't the shape implementation default to triangles pre-existing signed area function? 

no problem though, ive changed to test the original and one with a vertex permutation

aymey:
do you think there should be an `as_triangle` in "shape.rs"?

raphlinus:
> do you think there should be an `as_triangle` in "shape.rs"?

Probably not needed. The main purpose of those methods is fast paths for renderers, and triangle is not (unlike circle, rect, etc) not a common specialized shape - for example, it's not in SVG or the Canvas API.

src/triangle.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// The area of the [`Triangle`]
+    #[inline]
+    pub fn area(&self) -> f64 {
+        0.5 * (self.b - self.a).cross(self.c - self.a)
+    }
+
+    /// Whether this [`Triangle`] has no (zero) area
+    #[inline]
+    pub fn is_empty(&self) -> bool {
```
waywardmonkeys, line 117: In #370, I made it so that we'll use `is_zero_area` for this going forward. Since there hasn't been a release with this function in it for triangles, I think you should rename this to `is_zero_area` and add a doc alias like the other `is_zero_area` functions have.

We don't need a deprecated `is_empty` here.

src/triangle.rs
```diff
@@ -0,0 +1,344 @@
...
+        let centroid = self.centroid();
+
+        Self::new(
+            centroid + (0.0, scalar),
+            centroid + scalar * Vec2::from_angle(5.0 * FRAC_PI_4),
+            centroid + scalar * Vec2::from_angle(7.0 * FRAC_PI_4),
+        )
+    }
+
+    /// Is this [`Triangle`] finite?
```
waywardmonkeys, line 154: Please update this and the `is_nan` doc comments to match the newer style on the other types.

src/triangle.rs
```diff
@@ -0,0 +1,344 @@
...
+        )
+    }
+
+    #[inline]
+    fn as_rect(&self) -> Option<Rect> {
+        Some(self.bounding_box())
+    }
+}
+
+// NOTE: vertices a, b and c are not garunteed to be in order as described in the struct comments
```
waywardmonkeys, line 258: "guaranteed"

aymey: thanks, fixed in [8da1a8c](https://github.com/linebender/kurbo/pull/350/commits/8da1a8c506fa99d3ca76ed36b39986d157143734)!

a bit unrelated but do you think this justifies my `organise` proposal ([8b03193](https://github.com/linebender/kurbo/pull/350/commits/8b03193f6f932807b723c9d693a4010ba5e5076d))?

src/triangle.rs
aymey: just saw #371 should i add it here too?

waywardmonkeys: Yes please!

src/triangle.rs
DJMcNab: Has this been submitted to [typos](https://github.com/crate-ci/typos)?

aymey: yep! https://github.com/crate-ci/typos/issues/1024#issuecomment-2168557914

src/triangle.rs
waywardmonkeys: I've added a `From<(Point, Point)> for Line` in #376.

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+        Rect::new(
+            self.a.x.min(self.b.x.min(self.c.x)),
+            self.a.y.min(self.b.y.min(self.c.y)),
+            self.a.x.max(self.b.x.max(self.c.x)),
+            self.a.y.max(self.b.y.max(self.c.y)),
+        )
+    }
+
+    #[inline]
+    fn as_rect(&self) -> Option<Rect> {
```
derekdreery, line 252: This should be `None` IMO.

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub fn circumradius(&self) -> f64 {
+        let ab = self.a.distance(self.b);
+        let bc = self.b.distance(self.c);
+        let ac = self.a.distance(self.c);
+
+        (ab * bc * ac) / (4.0 * self.area())
+    }
+
+    /// Expand the triangle by a constant amount (`scalar`) in all directions
+    pub fn inflate(&self, scalar: f64) -> Self {
```
derekdreery, line 145: I'd consider adding a `#[doc(alias = "offset")]` here because we use the term `offset` generally to describe offsetting paths like this. The `Rect` type does also use this language IIRC.

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+// TODO: better and more tests
+#[cfg(test)]
+mod tests {
+    use crate::{Point, Triangle, Vec2};
+
+    fn assert_approx_eq(x: f64, y: f64) {
+        assert!((x - y).abs() < 1e-7);
```
derekdreery, line 280: This is quite a loose bound for `f64`.  I'd use something closer to machine epsilon e.g. `1e-13`. Out of scope for this PR, but we really need to standardize this (I like [`float_cmp`](https://docs.rs/float-cmp/latest/float_cmp/) because it allows big epsilons as long as the expected and actually are <n representable numbers apart in f64).

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    #[test]
+    fn centroid() {
+        let test = Triangle::from_coords((-90.02, 3.5), (7.2, -9.3), (8.0, 9.1)).centroid();
+        let expected = Point::new(-24.939999999999998, 1.0999999999999996);
```
derekdreery, line 286:
```suggestion
        let expected = Point::new(-24.94, 1.1);
```

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+        assert_eq!(test, expected);
+    }
+
+    #[test]
+    fn offsets() {
+        let test = Triangle::from_coords((-20.0, 180.2), (1.2, 0.0), (290.0, 100.0)).offsets();
+        let expected = [
+            Vec2::new(-110.39999999999999, 86.8),
+            Vec2::new(-89.19999999999999, -93.39999999999999),
+            Vec2::new(199.60000000000002, 6.6000000000000085),
```
derekdreery, line 297:
```suggestion
            Vec2::new(-110.4, 86.8),
            Vec2::new(-89.2, -93.4),
            Vec2::new(199.6, 6.6),
```

derekdreery (changes requested):
I like this PR and think it would make a good addition to `kurbo`, but I think it needs some changes first. The biggest suggestion in my review is to add functions `inscribed_circle` and `circumscribed_circle` and remove the functions that calculate radii/centers for these circles.

Feel free to push back if you disagree with any of my suggestions, and we can have a discussion.

src/point.rs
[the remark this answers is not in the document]
DJMcNab: As far as I can see, these functions aren't used in this PR. If we want to have them, we can just move them into a different PR to allow room for a proper discussion.
```suggestion
```

src/point.rs
```diff
@@ -185,6 +186,30 @@ impl Point {
...
+    /// The topmost [`Point`]
+    #[inline]
+    pub fn topmost(&self, other: &Self) -> Ordering {
```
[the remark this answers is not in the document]
juliapaci: yes i think this was pointed out by @raphlinus but i ended up keeping the names while `organise()` ([8b03193](https://github.com/linebender/kurbo/pull/350/commits/8b03193f6f932807b723c9d693a4010ba5e5076d)) was still in use. i just forgot to change them since `organise()` was removed, sorry.

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+        Self {
+            a: a.into(),
+            b: b.into(),
+            c: c.into(),
+        }
+    }
+
+    /// A new [`Triangle`] from three float vertex coordinates
+    #[inline]
+    pub const fn from_coords(a: (f64, f64), b: (f64, f64), c: (f64, f64)) -> Self {
```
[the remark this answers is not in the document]
juliapaci: Ive used it to keep the constants `ZERO` and `EQUILATERAL`. 

Maybe there is a way to use the non constant `new()` since we can assume the arguments are already of `Point`. If there isnt another way then should i remove them or maybe i can make `from_coords()` private?

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    /// The centroid of the [`Triangle`]
+    #[inline]
+    pub fn centroid(&self) -> Point {
+        (1.0 / 3.0 * (self.a.to_vec2() + self.b.to_vec2() + self.c.to_vec2())).to_point()
+    }
+
+    /// The circumcenter of the [`Triangle`]
+    #[inline]
+    pub fn circumcenter(&self) -> Point {
```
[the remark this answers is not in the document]
juliapaci: i looked a bit through the compiler explorer for some small optimizations and cleanups but i fear that i am missing a faster algorithm in general. Also any thoughts on [this](https://github.com/linebender/kurbo/pull/350#issuecomment-2290855599) (specifically the first paragraph)?

src/triangle.rs
```diff
@@ -0,0 +1,343 @@
...
+        let bc = self.b.distance(self.c);
+        let ac = self.a.distance(self.c);
+
+        2.0 * self.area() / (ab + bc + ac)
+    }
+
+    /// Circumradius of [`Triangle`] (the smallest radius of a circle such that it intercepts each vertex)
+    /// with center [`Triangle::circumcenter`]
+    #[inline]
+    pub fn circumradius(&self) -> f64 {
```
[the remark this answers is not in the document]
juliapaci: Yeah i have no problem either way. 

> The only argument against is that we might be computing extra information...

I think its more likely that the user would want the center _and_ the radius rather than just one of them

src/triangle.rs
juliapaci: sorry, did you mean to comment this under `Triangle::offsets()`?

src/triangle.rs
juliapaci: Yes, i should've left the defaults.

src/triangle.rs
juliapaci: I copied the bound from "rect.rs" i understood that it was loose but knew that it was only a temporary solution. 

i could change it to be more strict but even with `1e-13`, `Triangle::area()` tests are already failing from  precision. i still could arbitrarily restrict the bound or could just wait for it to be standardized as you said, which would you prefer?

juliapaci (commented):
Thanks for the review and sorry for the late response.

All suggestions have been implemented ([4f17103](https://github.com/linebender/kurbo/pull/350/commits/4f171037e4f8bf59adc8dab622447465abf41b4c), [3da32b7](https://github.com/linebender/kurbo/pull/350/commits/3da32b722da480bde816b64e295f314a59711e5b), 4d0afb9) except for those which i further commented on, please see those!

also why is github repeating my comments twice from a code review (once for a reply and another for a standalone comment it seems)? its really cluttering up this page, am i not using the code review functionality correctly?

src/point.rs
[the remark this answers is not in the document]
juliapaci: see https://github.com/linebender/kurbo/pull/350#discussion_r1770856959

src/triangle.rs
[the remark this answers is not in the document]
DJMcNab: The easiest way is probably to just inline this function at the definitions. I.e. `const ZERO: Triangle = Triangle {a: Point::new(...), ...}`.

I do agree that there should be some consistency here, but I'm happy just using the already-exposed "fieldwise" constructor as the `const` constructor for now to unblock progress.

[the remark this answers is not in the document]
juliapaci: Yeah im happy to do that. should we quickly do it now or wait for @derekdreery's input?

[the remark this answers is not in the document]
derekdreery: Hmm I hadn't realised that this function exists for `const`. In that case I think it's good to have 2 functions,  but perhaps put a note in the docs explaining that this function works like `new`, but can be used in `const` contexts.

src/triangle.rs
derekdreery: Just leave it as-is for now - what you have is fine.

juliapaci:
tests are failing from https://github.com/linebender/kurbo/pull/350#discussion_r1770606024 and https://github.com/linebender/kurbo/pull/350#discussion_r1770605939

src/triangle.rs
derekdreery: I mean on the `inflate` function. This is also an offset with mitered joins.

juliapaci: oh right, i see.

src/lib.rs
```diff
@@ -136,4 +137,5 @@ pub use crate::stroke::*;
 #[cfg(feature = "std")]
 pub use crate::svg::*;
 pub use crate::translate_scale::*;
+pub use crate::triangle::*;
```
waywardmonkeys, line 140: This part needs to rebase forward as it conflicts with `main`.

src/triangle.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// vertex a
+    pub a: Point,
+    /// vertex b
+    pub b: Point,
+    /// vertex c
+    pub c: Point,
+}
+
+impl Triangle {
+    /// The empty [`Triangle`] at the origin
```
waywardmonkeys, line 34: This and many (most) other comments in this file need to end in a `.`

waywardmonkeys (approved):
Apart from 2 things noted here, I think I'm happy with the current state of this, recognizing there are probably some good follow-up items that we can tackle in smaller pieces going forward.

waywardmonkeys:
@derekdreery You'll need to approve this (so long as you're okay with everything of course) so that we'll be able to merge it once the last couple of things are in place.

DJMcNab:
Fwiw, you should be able to dismiss the review, if we think everything has been addressed.

waywardmonkeys:
@juliapaci If you could fix the tests (maybe look at `assert_approx_eq`), then I think we can land this. Hopefully @derekdreery can approve from his review comments being addressed.

waywardmonkeys:
I will merge this in about 26-28 hours.

waywardmonkeys (approved):

waywardmonkeys:
I think there are probably things that we could follow up on ... but let's get this landed and deal with those separately.

PR #352, opened 2024-06-06
Fix license discrepancy

raphlinus:
All files should be Apache + MIT dual. As author of this file I don't believe any further approvals are needed.

Thanks to @dominikh for flagging.


xStrom (approved):
Looks good.

Discrepancies likes this will eventually also be caught by the CI, when I get around to updating the Kurbo CI to the freshest Linebender standard.

PR #353, opened 2024-06-10
Don't skip first dash in dash pattern

dominikh:
Before this change, the dash iterator's initial state (before accounting
for the dash offset) was "processing first dash, with remaining length
of zero", which would cause the iterator to immediately jump to the next
component in the dash pattern, effectively skipping the first dash and
starting with a gap.

### DIFF
--- a/src/stroke.rs
+++ b/src/stroke.rs
@@ -649,16 +649,12 @@ fn dash_impl<T: Iterator<Item = PathEl>>(
     dashes: &[f64],
 ) -> DashIterator<T> {
     let mut dash_ix = 0;
-    let mut dash_remaining = dash_offset;
+    let mut dash_remaining = dashes[dash_ix] - dash_offset;
     let mut is_active = true;
     // Find place in dashes array for initial offset.
-    while dash_remaining > 0.0 {
-        let dash_len = dashes[dash_ix];
-        if dash_remaining < dash_len {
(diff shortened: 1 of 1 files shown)


dominikh:
This change currently breaks for non-zero dash offsets

waywardmonkeys:
@dominikh Could you please rebase this on top of current `main` to pick up the CI changes / updates? Thanks!

raphlinus (approved):
Yes, the logic was wrong and this seems to be a valid fix. Thanks!

Related to this is the report in [Dashed Line misbehaves when used with dash-offset](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Dashed.20Line.20misbehaves.20when.20used.20with.20dash-offset). I think it's worth re-checking the behavior.

Reading https://github.com/GraphiteEditor/Graphite/pull/1875, which is the tracking issue where the Graphite behavior was reported, I *think* this takes care of the sign negation, but it's possible there's another issue. Please file another bug against either kurbo or Vello if so.

Keavon:
Thanks for getting to this in your review queue!

Once this lands and it's pulled in by the latest Vello master commit, we can try it out in Graphite to confirm the behavior matches the SVG spec. I'll report back in that Zulip thread if it's working, or file an issue here again if it's not.

PR #355, opened 2024-07-22
ci: Add typos check.

waywardmonkeys:


DJMcNab (approved):
Given that there are no failures, this seems low-risk.

Do we want to also add this as required? I don't really work on Kurbo, so don't want to barge in and add that.

PR #356, opened 2024-08-02
Allow some of the SVG support in `no_std`

waywardmonkeys:
Some of the SVG functionality builds with `no_std` and can reasonably be enabled.


waywardmonkeys:
The CI failure is due to the CI here not being Linebender standard and having a fixed stable toolchain and we haven't yet fixed the issues here with `unexpected_cfgs` and friends.

Will look into that as a separate PR.

waywardmonkeys:
I'll rebase this on top of #358 once that lands so that I can add an appropriate changelog entry here without having it generate a conflict.

Philipp-M (approved):
Looks good, tested both std and no_std

PR #357, opened 2024-08-02
ci: Use Linebender standard CI. Add `Cargo.lock`.

waywardmonkeys:
This is copied from Peniko with minor adjustments:
* `RUST_STABLE_VER` is 1.79 as Kurbo needs fixes for 1.80.
* `RUST_MIN_VER` is 1.65.
* `RUST_MIN_VER_PKGS` checks `kurbo`.
* `FEATURES_DEPENDING_ON_STD` added `schemars`.
* Change the authors check in the copyright step.
 * Changes to handle not saving the cache when doing a merge_group.
   These come from Vello.

The dependencies are all updated to their current versions and a `Cargo.lock` is now used in accordance with other Linebender projects.


waywardmonkeys:
@xStrom @DJMcNab This will need administrative assistance because the CI checks are changing.

.github/workflows/ci.yml
```diff
@@ -7,122 +50,197 @@ on:
   pull_request:
```
DJMcNab, line 50: Do we want to add support for merge queues here?
See also https://github.com/linebender/vello/pull/647 in that case.

DJMcNab (approved):
If there are no other Kurbo maintainers we need to involve with this, happy for it to land.

Once you're sure this is ready to go in, I'll make the config changes.

.github/workflows/ci.yml
waywardmonkeys: Updated for that.

waywardmonkeys:
@DJMcNab I think this is good to go now and if you want to squash and then go, that's great.

PR #359, opened 2024-08-02
Enable `feature(doc_auto_cfg)` for docs.rs

waywardmonkeys:
Docs built using this (like for docs.rs) will show what feature must be enabled for things that are `cfg(feature = ...)` gated.

This is a nightly only thing. If you want to run it locally, you can use:

    RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,6 +263,8 @@ jobs:
       # We test documentation using nightly to match docs.rs. This prevents potential breakages
       - name: cargo doc
         run: cargo doc --workspace --locked --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples
+        env:
+          RUSTDOCFLAGS: '--cfg docsrs'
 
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Seems reasonably low-risk, as we can blend in with the crowd to stop docs.rs breaking this (e.g. [iced](https://github.com/iced-rs/iced/commit/cf434236e7e15e0fa05e5915b8d4d78dcaf1b7e8), [polars](https://github.com/pola-rs/polars/pull/6201))

I'm slightly concerned about the failure mode of this feature becoming stabilised; I will have at one point looked into what happens then, but can't remember off the top of my head.

waywardmonkeys:
I'm kind of unhappy that the feature doesn't handle the case of a `#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]` ... but at least this is better than nothing.

I will (eventually) have PRs coming for the other repos as appropriate ...

PR #361, opened 2024-08-03
Deprecate `BezPath::flatten`, prefer `flatten`.

waywardmonkeys:
The free-standing function is more useful as it works on a wider range of types. The `BezPath` method has the documentation though and is just a wrapper around the free-standing function.

Most people will want to use the free-standing function, so we can encourage that and make that the place where the docs live.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This release has an [MSRV][] of 1.65.
 
 - Move `Self: Sized` bound from `Shape` to methods. ([#340] by [@waywardmonkeys])
 - Enable partial SVG path support in `no_std` builds. ([#356] by [@waywardmonkeys])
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -50,7 +51,11 @@ Note: A changelog was not kept for or before this release
 [#343]: https://github.com/linebender/kurbo/pull/343
 [#347]: https://github.com/linebender/kurbo/pull/347
 [#354]: https://github.com/linebender/kurbo/pull/354
+<<<<<<< HEAD
 [#356]: https://github.com/linebender/kurbo/pull/356
+=======
+[#361]: https://github.com/linebender/kurbo/pull/361
+>>>>>>> d7f955a (Deprecate `BezPath::flatten`, prefer `flatten`.)
```
DJMcNab, line 58: Merge artifact

raphlinus (approved):
Yup, makes sense, thanks!

PR #363, opened 2024-08-04
docs: Improve docs for `Point` and `Vec2`

waywardmonkeys:
This adds some cross-linking as well as a couple of additional doc tests.

### DIFF
--- a/src/point.rs
+++ b/src/point.rs
@@ -64,19 +64,22 @@ impl Point {
     }
 
     /// Euclidean distance.
+    ///
+    /// See [`Vec2::hypot`] for the same operation on [`Vec2`].
     #[inline]
     pub fn distance(self, other: Point) -> f64 {
         (self - other).hypot()
     }
 
     /// Squared Euclidean distance.
+    ///
+    /// See [`Vec2::hypot2`] for the same operation on [`Vec2`].
     #[inline]
     pub fn distance_squared(self, other: Point) -> f64 {
         (self - other).hypot2()
     }
 
-    /// Returns a new `Point`,
-    /// with `x` and `y` rounded to the nearest integer.
+    /// Returns a new `Point`, with `x` and `y` [rounded] to the nearest integer.
     ///
     /// # Examples
     ///
@@ -89,13 +92,15 @@ impl Point {
     /// assert_eq!(b.x, 3.0);
     /// assert_eq!(b.y, -3.0);
     /// ```
+    ///
+    /// [rounded]: f64::round
     #[inline]
     pub fn round(self) -> Point {
         Point::new(self.x.round(), self.y.round())
     }
 
     /// Returns a new `Point`,
-    /// with `x` and `y` rounded up to the nearest integer,
+    /// with `x` and `y` [rounded up] to the nearest integer,
     /// unless they are already an integer.
     ///
     /// # Examples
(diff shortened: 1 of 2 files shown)


src/vec2.rs
```diff
@@ -86,6 +111,23 @@ impl Vec2 {
...
+    /// # See also
+    ///
+    /// * [`Point::distance_squared`]
+    /// * [`Vec2::dot`]
+    /// * [`Vec2::hypot`]
+    /// * [`Vec2::length`]
+    /// * [`Vec2::length_squared`]
+    /// * [`Vec2::normalize`]
     #[inline]
     pub fn hypot2(self) -> f64 {
```
DJMcNab, line 132: I love that this is `hypot2` whereas the point method is `distance_squared` :)

src/point.rs
```diff
@@ -89,6 +99,14 @@ impl Point {
...
     /// assert_eq!(b.y, -3.0);
     /// ```
+    ///
+    /// # See also
+    ///
+    /// * [`f64::round`]
+    /// * [`Point::ceil`]
+    /// * [`Point::expand`]
+    /// * [`Point::floor`]
+    /// * [`Point::trunc`]
```
DJMcNab, line 109: I'm not sure these are that helpful; `std` doesn't have the equivalent links.

src/point.rs
```diff
@@ -89,6 +99,14 @@ impl Point {
     /// assert_eq!(b.x, 3.0);
     /// assert_eq!(b.y, -3.0);
     /// ```
+    ///
+    /// # See also
+    ///
+    /// * [`f64::round`]
```
DJMcNab, line 105: This is very useful though, as it gives more details

src/vec2.rs
```diff
@@ -145,10 +187,12 @@ impl Vec2 {
         self + t * (other - self)
     }
 
-    /// Returns a vector of magnitude 1.0 with the same angle as `self`; i.e.
+    /// Returns a vector of [magnitude] 1.0 with the same angle as `self`; i.e.
```
DJMcNab, line 190: 🔥
I like this!

src/point.rs
```diff
@@ -64,12 +64,22 @@ impl Point {
     }
 
     /// Euclidean distance.
+    ///
+    /// # See also
+    ///
+    /// * [`Point::distance_squared`]
```
DJMcNab, line 70: This is a useful cross-link. It probably should be explained in prose instead of a generic see also, but this is an improvement.

src/vec2.rs
```diff
@@ -72,6 +80,23 @@ impl Vec2 {
     ///
     /// This is similar to `self.hypot2().sqrt()` but defers to the platform `hypot` method, which
```
DJMcNab, line 81: Can we make this be a link to `f64::hypot`? I for one didn't know this method existed previously

DJMcNab (approved):
Good to improve docs here.

I think that almost all of the links to items which are on the same page are quite noisy and don't provide much value, so would prefer them to be removed.
But I'm not going to block on that if you think it's better.

src/vec2.rs
waywardmonkeys: Yes. I find a lot of the naming in `kurbo` to be confusing.

waywardmonkeys:
Updated, I think I've addressed the feedback.

I disagree about some of it, but I'll go with the flow for now!

src/vec2.rs
```diff
@@ -248,24 +279,30 @@ impl Vec2 {
...
     /// ```
+    ///
+    /// [rounded towards]: f64::trunc
     #[inline]
     pub fn trunc(self) -> Vec2 {
         Vec2::new(self.x.trunc(), self.y.trunc())
     }
 
-    /// Is this Vec2 finite?
+    /// Is this `Vec2` finite?
```
DJMcNab, line 289:
```suggestion
    /// Is this `Vec2` [finite]?
```

DJMcNab (approved):
Great! This looks really good!

PR #364, opened 2024-08-06
Explicitly export rather than `pub use mod::*`.

waywardmonkeys:
This makes it easier to see what the actual API is and what's coming from where.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,25 +114,35 @@ mod svg;
 mod translate_scale;
 mod vec2;
 
-pub use crate::affine::*;
-pub use crate::arc::*;
-pub use crate::bezpath::*;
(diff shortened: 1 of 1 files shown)


DJMcNab (commented):
The style I used in Xilem Core was:
```
mod whatever;
use whatever::{...};

mod something_else;
use something_else::{...};
```

But I'm generally in favour of this. We should probably have `unreachable_pub` enabled

raphlinus (approved):
Agreed, this is better. I can see I'm the one guilty of the original pattern, which dates back to the very first commit.

waywardmonkeys:
@DJMcNab That style felt too verbose here ... not sure why the difference in feel.

PR #365, opened 2024-08-14
fix: parsing scientific notation in SVG path

GabrielDertoni:
## Before this change

The SVG parsing code failed to parse scientific notation in `BezPath`. For example `M 0 0 L 1e-123 -4E+5` would fail to parse with `SvgParseError::Wrong` even though it is a valid SVG path. This error was introduced by 8d10d4e which [forgot to negate the conditional](https://github.com/linebender/kurbo/commit/8d10d4e00981238f1ba555fe44af000b81aa4ec1#diff-a0ea1370686e222812265cfdba79ff5c9458cfc267ebd7c8b2221dc735909483L317), causing the parser to fail when it should succeed.

## After this change

The change fixes the error and adds a test for it.

### DIFF
--- a/src/svg.rs
+++ b/src/svg.rs
@@ -352,11 +352,11 @@ impl<'a> SvgLexer<'a> {
                 if c == b'-' || c == b'+' {
                     c = self.get_byte().ok_or(SvgParseError::Wrong)?;
                 }
-                if c.is_ascii_digit() {
+                if !c.is_ascii_digit() {
                     return Err(SvgParseError::Wrong);
                 }
                 while let Some(c) = self.get_byte() {
-                    if c.is_ascii_digit() {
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
Thanks! I actually have a local branch called "fix_svg_exp" with the exact change and a commit comment saying "WIP, really needs a test." So thanks for getting this in.

We have a policy of "author commits" for substantive changes. This is a smaller one (but one that fixes a real bug) so it's up to you whether you'd like us to merge or whether you'd like write access.

GabrielDertoni:
No need to give me write access yet (if I end up wanting to make larger contributions we can revisit). I am happy to just get this in whichever way is faster.

raphlinus:
Thanks! I just added it to the merge queue.

PR #366, opened 2024-08-19
Add more docs.

waywardmonkeys:


src/affine.rs
```diff
@@ -156,81 +156,99 @@ impl Affine {
...
     #[inline]
     #[must_use]
     pub fn pre_rotate(self, th: f64) -> Self {
         self * Affine::rotate(th)
     }
 
-    /// A rotation by `th` about `center` followed by `self`.
+    /// A [rotation] by `th` about `center` followed by `self`.
     ///
     /// Equivalent to `self * Affine::rotate_about(th)`
```
DJMcNab, line 172:
```suggestion
    /// Equivalent to `self * Affine::rotate_about(th, center)`
```

src/affine.rs
```diff
@@ -156,81 +156,99 @@ impl Affine {
...
+    ///
+    /// [rotation]: Affine::rotate_about
     #[inline]
     #[must_use]
     pub fn pre_rotate_about(self, th: f64, center: Point) -> Self {
         Affine::rotate_about(th, center) * self
```
DJMcNab, line 178: I think this needs to be:

```suggestion
        self * Affine::rotate_about(th, center)
```
Although that might be for a different PR

src/arc.rs
```diff
@@ -80,9 +80,9 @@ impl Arc {
         }
     }
 
-    /// Converts an Arc into a series of cubic bezier segments.
+    /// Converts an `Arc` into a series of cubic bezier segments.
     ///
-    /// Closure will be invoked for each segment.
+    /// The closure `p` will be invoked for each segment.
```
DJMcNab, line 85:
```suggestion
    /// The closure `p` will be invoked with the control points for each segment.
```
(Guessing what the arguments are, but users will also need to do so)

src/size.rs
```diff
@@ -172,13 +178,15 @@ impl Size {
...
+    /// [rounded away]: FloatExt::expand
     #[inline]
     pub fn expand(self) -> Size {
         Size::new(self.width.expand(), self.height.expand())
     }
 
     /// Returns a new `Size`,
-    /// with `width` and `height` rounded down towards zero the nearest integer,
+    /// with `width` and `height` [rounded towards] zero to the nearest integer,
     /// unless they are already an integer.
```
DJMcNab, line 190: An optional suggestion - I'd slightly prefer this to be something like:
```suggestion
    /// with `width` and `height` [truncated][rounded towards] to their integer part.
    /// This effectively rounds them to the nearest integer towards zero.
```
I don't think "rounds towards zero to the nearest integer" flows very well, because in my mind "towards zero" is a modifier on the entire "rounds to the nearest integer".

DJMcNab (approved):
Lovely, thanks!

There is one nascent bug which this has made obvious but doesn't need to fix, but which does need addressing.

src/size.rs
waywardmonkeys: I'd rather this one be addressed separately as this sort of text is in more than one place.

src/affine.rs
waywardmonkeys: This needs to be a separate PR. I think the core is wrong here, not the doc comment.

src/affine.rs
waywardmonkeys: Applied.

src/arc.rs
waywardmonkeys: Applied.

src/affine.rs
waywardmonkeys: What I meant was that I too think the code is wrong and this is a docs patch, so I'd rather handle it separately. :)

PR #367, opened 2024-08-19
Add `Arc::flipped`

waywardmonkeys:


DJMcNab (commented):
The implementation looks reasonable; what are the expected use-cases for this?

src/arc.rs
```diff
@@ -206,3 +221,24 @@ impl Mul<Arc> for Affine {
...
+        // Most fields should be unchanged:
+        assert_eq!(a.center, f.center);
+        assert_eq!(a.radii, f.radii);
+        assert_eq!(a.x_rotation, f.x_rotation);
+
+        // Sweep angle should be in reverse
+        assert_eq!(a.sweep_angle, -f.sweep_angle);
+
+        // Flipping it again should result in the original arc
+        assert_eq!(a, f.flipped());
```
derekdreery, line 242: No fuzzy testing needed, since there is only addition/subtraction meaning perfect accuracy, at least for normal input.

derekdreery (approved):
LGTM.

@DJMcNab these kinds of operations can be useful sometimes, for applications other than drawing (when it really doesn't matter what direction the line goes in).

src/arc.rs
waywardmonkeys: You are confirming what I have here is fine, not suggesting a change, right?

DJMcNab:
What are the next steps here?

waywardmonkeys:
I am going to land it before the meeting.

PR #368, opened 2024-08-22
`CircleSegment`: Add `inner_arc` and `outer_arc`

waywardmonkeys:


waywardmonkeys:
Draft because:

* Depends on #367, `Arc::flipped` for a doc comment link.
* Am I right to have the `inner_arc` flipped from the `outer_arc`?
* Should we add similar things for the two radial lines connecting the arcs?

derekdreery (approved):
LGTM

 - I think it's right that the direction matches the way the arc is drawn
 - we can always add radial lines later (and personally I doubt they're useful)

PR #370, opened 2024-08-28
Add `Rect::is_zero_area`, `Size::is_zero_area`

waywardmonkeys:
These replace `Rect::is_empty` and `Size::is_empty` (which are now deprecated) as the treatment of negative areas as not being empty can be slightly confusing, so using a more specific name makes things more clear.


xorgy (approved):

derekdreery (approved):
Yep the names are better.

Can I suggest adding `is_empty` as a `#[doc(alias = "is_empty")]` because `is_empty` is something a user might guess as the name (and it will disappear once the deprecated methods are removed).

waywardmonkeys:
@derekdreery Done!

PR #373, opened 2024-08-30
Update changelog

waywardmonkeys:


DJMcNab (approved):
Looks good, thanks. I've scanned through the changes, and this looks like everything which should be mentioned in the changelog.
The only change I'd like to see is:

```md
## Unreleased
```
(line 14)
being changed to:
```md
## [Unreleased](https://github.com/linebender/kurbo/compare/v0.11.0...HEAD)
```

I'm also not completely sure about how we want our "fixed" section to read; whether there's an implied "fixed ...". The current format saying "fix" in each item is more compatible with something like "Bug Fixes". (similarly "New Features" instead of "Added")

That is not to be dealt with in this PR.

waywardmonkeys:
It should've just been `[Unreleased]` since there was already a link for it below. Fixed.

PR #372, opened 2024-08-30
deps: Update for release.

waywardmonkeys:

PR #374, opened 2024-08-30
Release 0.11.1

waywardmonkeys:


waywardmonkeys:
Tentatively thinking about Monday, September 2. This is blocked on #372 and #373. Not sure that we'll get #350 in or not, but we can always cut a new release sooner rather than later.

DJMcNab:
Going through the commits in this release, whilst reviewing #373. 

Could #340 be breaking? I think probably technically yes, but practically it requires writing adversarial code to do so.
Are #361/#370 breaking? I think the worst case is it breaks end-user CI when they are using `-D warnings`, and not using a Cargo.lock, so it's probably fine.

waywardmonkeys:
Re: #340, there was a consensus at the time that it wasn't. I also think I checked with the author of semver-checks and he agreed not breaking.

Re: #361 / #370: By convention, this is not breaking.

DJMcNab (commented):
I'd like to see this landed, although I can't formally approve due to lacking publish rights on Kurbo.

The actual code changes in this PR are correct.

xorgy (approved):

PR #375, opened 2024-09-01
Add `Line::reversed`, `Line::midpoint`, rename `Arc::flipped`

waywardmonkeys:


src/line.rs
```diff
@@ -34,12 +34,32 @@ impl Line {
         }
     }
 
+    /// Return a copy of this `Line` in the opposite direction.
+    #[must_use]
+    #[inline]
+    pub fn flipped(&self) -> Line {
```
dfrg, line 40: nit: I feel like `reversed` is a slightly better name for this.

Also, it might be helpful if the comment describes what it does: “Returns a copy of this Line with the end points swapped so that it points in the opposite direction.”

dfrg (approved):
Some minor bikeshedding, otherwise LGTM.

src/line.rs
waywardmonkeys: I'll revise the comment.

The name matches the other `flipped` that I've added in a previous PR, `Arc::flipped`.

I think I like `reversed` better though as well, so will update both.

waywardmonkeys:
@dfrg  Updated to also rename `Arc::flipped` ... what do you think?

dfrg:
All looks good, thanks!

derekdreery:
If there are 2 plausible names, feel free to add the less good one as a doc alias :)

PR #376, opened 2024-09-09
`Line::from` for `(Point, Point)`, `(Point, Vec2)`

waywardmonkeys:


Philipp-M (approved):
LGTM

src/line.rs
```diff
@@ -90,6 +90,18 @@ impl Line {
     }
 }
 
+impl From<(Point, Point)> for Line {
+    fn from(points: (Point, Point)) -> Self {
```
DJMcNab, line 94:
```suggestion
    fn from((a, b): (Point, Point)) -> Self {
```

src/line.rs
```diff
@@ -90,6 +90,18 @@ impl Line {
...
+        Line::new(points.0, points.1)
+    }
+}
+
+impl From<(Point, Vec2)> for Line {
+    fn from(pv: (Point, Vec2)) -> Self {
```
DJMcNab, line 100:
```suggestion
    fn from((origin, direction): (Point, Vec2)) -> Self {
```

DJMcNab (commented):
I'd suggest using destructuring here

src/line.rs
waywardmonkeys: Done, using `(from, to)`

src/line.rs
waywardmonkeys: Done, using `(origin, displacement)`

DJMcNab (approved):
Code looks correct once the push is actually done - if this is definitely an API we want, then I'd be happy to land this.

waywardmonkeys:
Will merge today if there's no objections registered.

PR #378, opened 2024-09-14
Impl `ParamCurve`, `ParamCurveArclen` for `Arc`

waywardmonkeys:
This uses an approximation of the arc length using beziers as the analytic solution is quite involved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ You can find its changes [documented below](#0111-2024-09-12).
 
 This release has an [MSRV][] of 1.65.
 
+### `Arc` now implements `ParamCurve` and `ParamCurveArclen`. ([#378] by [@waywardmonkeys])
+
 ## [0.11.1][] (2024-09-12)
 
 This release has an [MSRV][] of 1.65.
@@ -75,6 +77,7 @@ Note: A changelog was not kept for or before this release
 [#370]: https://github.com/linebender/kurbo/pull/370
 [#375]: https://github.com/linebender/kurbo/pull/375
 [#376]: https://github.com/linebender/kurbo/pull/376
+[#378]: https://github.com/linebender/kurbo/pull/378
 
 [Unreleased]: https://github.com/linebender/kurbo/compare/v0.11.1...HEAD
 [0.11.0]: https://github.com/linebender/kurbo/releases/tag/v0.11.0
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -3,11 +3,11 @@
 
 //! An ellipse arc.
 
-use crate::{Affine, Ellipse, PathEl, Point, Rect, Shape, Vec2};
+use crate::{Affine, Ellipse, ParamCurve, ParamCurveArclen, PathEl, Point, Rect, Shape, Vec2};
 use core::{
     f64::consts::{FRAC_PI_2, PI},
     iter,
-    ops::Mul,
+    ops::{Mul, Range},
 };
 
 #[cfg(not(feature = "std"))]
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ You can find its changes [documented below](#0111-2024-09-12).
 
 This release has an [MSRV][] of 1.65.
 
+### `Arc` now implements `ParamCurve` and `ParamCurveArclen`. ([#378] by [@waywardmonkeys])
```
DJMcNab, line 18:
```suggestion
### Changes

- `Arc` now implements `ParamCurve` and `ParamCurveArclen`. ([#378] by [@waywardmonkeys])
```

DJMcNab (commented):
Can't comment on the maths

src/arc.rs
```diff
@@ -171,6 +171,42 @@ fn rotate_pt(pt: Vec2, angle: f64) -> Vec2 {
...
+        let angle = self.start_angle + (self.sweep_angle * t);
+        sample_ellipse(self.radii, self.x_rotation, angle).to_point()
+    }
+
+    fn subsegment(&self, range: Range<f64>) -> Self {
+        Self {
+            center: self.center,
+            radii: self.radii,
+            start_angle: self.start_angle + (self.sweep_angle * range.start),
+            sweep_angle: self.sweep_angle - (self.sweep_angle * (1.0 - range.end)),
```
raphlinus, line 185: This feels wrong, though I haven't done tests to validate it. My intuition says it should be `self.sweep_angle * (range.end - range.start)`. In any case, the code that's there reduces algebraically to `self.sweep_angle * range.end`. It is of course possible I'm missing something.

waywardmonkeys: I think what I have is wrong. I'll get back into this and see.

src/arc.rs
```diff
@@ -171,6 +171,42 @@ fn rotate_pt(pt: Vec2, angle: f64) -> Vec2 {
...
+            self.x_rotation,
+            self.start_angle + self.sweep_angle,
+        )
+        .to_point()
+    }
+}
+
+impl ParamCurveArclen for Arc {
+    fn arclen(&self, accuracy: f64) -> f64 {
+        self.path_segments(0.1).perimeter(accuracy)
```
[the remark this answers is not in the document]
waywardmonkeys: I was using the same math being used for the `perimeter` function ... I do think we should improve upon this and also either have this call the `perimeter` function or have that one call this one.

src/arc.rs
waywardmonkeys: I think I've done what you asked here. I think I should add a test.

src/arc.rs
```diff
@@ -171,6 +171,42 @@ fn rotate_pt(pt: Vec2, angle: f64) -> Vec2 {
...
+        let angle = self.start_angle + (self.sweep_angle * t);
+        sample_ellipse(self.radii, self.x_rotation, angle).to_point()
+    }
+
+    fn subsegment(&self, range: Range<f64>) -> Self {
+        Self {
+            center: self.center,
+            radii: self.radii,
+            start_angle: self.start_angle + (self.sweep_angle * range.start),
+            sweep_angle: self.sweep_angle - (self.sweep_angle * (range.end - range.start)),
```
tomcur, line 185: I believe this is not right, I think it should be 

```rust
sweep_angle: self.sweep_angle * (range.end - range.start),
```

For example, in the case where the range is `0.0..1.0` the subsegment should return the original arc.

src/arc.rs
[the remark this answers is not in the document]
xorgy: `accuracy.sqrt()`  :+ )

PR #379, opened 2024-09-20
Derive `PartialEq` for `Stroke`, common derives for `StrokeOpts`.

waywardmonkeys:
As discussed in https://xi.zulipchat.com/#narrow/stream/260979-kurbo/topic/PartialEq.20for.20Stroke.3F


raphlinus (approved):
Yup, I think these make sense. Thanks!

PR #380, opened 2024-09-26
test: 💍 add failing offset test

richard-uk1:
This patch doesn't have a fix, but it records the problem.

Changing to `fit_to_bezpath` (without the `_opt`) results in no error.

Run `cargo test -- --ignored` to see the failure.

### DIFF
--- a/src/offset.rs
+++ b/src/offset.rs
@@ -199,6 +199,22 @@ mod tests {
         fit_to_bezpath(&co, TOLERANCE);
     }
 
+    /// Another cubic offset that triggers infinite recursion
+    // TODO make this test pass and then remove the `#[ignore]`
+    #[test]
+    #[ignore]
+    fn infinite_recursion_2() {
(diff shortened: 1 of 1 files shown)


mlwilkerson:
I think the problem identified by this is the same as the one I reported in #602. Claude helped investigate and found that the curve in this PR no longer manifests the bug, since `CubicOffset` was removed in #489 and offsetting rewritten.

But I think #602 has a repro that still does reproduce the same underlying defect that this PR's test case previously reproduced and it might be that both #588 and #600 fix that remaining defect.

So this PR is probably stale and could be closed. Or, it might be worth keeping its test case as a regression test against what it previously failed on. However, Claude noted that the test's name suggests that infinite recursion is the root cause, but actually:

>   The stack is finite; the ITP bracket simply stops moving once scaled_epsilon wraps, and the uncapped while b - a >
  2.0 * epsilon loop spins forever.

PR #381, opened 2024-09-27
Numerically approximate arc perimeter

tomcur:
This is on top of PR https://github.com/linebender/kurbo/pull/378.

This implementation uses Carlson symmetric forms (https://arxiv.org/abs/math/9409227v1) of incomplete elliptic integrals of the second kind to numerically approximate elliptical arc lengths. These forms have nice computational properties allowing quick convergence of the approximations (each iteration reduces the error bound by 4⁶). Boost and Gnu Scientific Library are some other projects using them.

For the complete elliptic parts of the arc, the existing complete elliptic integral approximation (#383) is used.

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -385,7 +385,6 @@ fn carlson_rf(accuracy: f64, x: f64, y: f64, z: f64) -> f64 {
     let mut e = a - x.min(y).min(z);
     let mut r = accuracy * 4. * e.powi(-6);
 
-    // let mut q = 1. / (3. * f64::EPSILON).cbrt().sqrt() * (a - x.min(y).min(z));
     loop {
         if 1. <= r * a.powi(6) * a.sqrt() * (1. - e / a) {
             break;
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -327,6 +327,9 @@ impl Mul<Arc> for Affine {
 ///
 /// RF = 1/2 ∫ 1 / ( sqrt(t+x) sqrt(t+y) sqrt(t+z) ) dt from 0 to inf
 fn carlson_rf(relative_error: f64, x: f64, y: f64, z: f64) -> f64 {
+    // At most one of (x, y, z) may be 0.
+    debug_assert!((x == 0.) as u8 + (y == 0.) as u8 + (z == 0.) as u8 <= 1);
+
     let mut x = x;
     let mut y = y;
     let mut z = z;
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
I've no idea about the math here, but what Raph and I had discussed (somewhat) for #378 is that I would fix the one part that is broken and it could land and then a follow up (like this one?) would replace the current math with better math for the `arclen` operation.

tomcur:
This is based on top of your PR, changing `arclen` only. I'll rebase on top of your future fix , and rebase on main once your PR is merged.

The `arclen` implementation here needs more work, but that can proceed independently from fixes in #378.

tomcur:
Simplified a bit and hopefully all edge cases are now correct.

`arclen`'s `accuracy` param is not handled yet, but the Carlson approximations can be iterated to arbitrary precision. The easiest is to guarantee a relative `accuracy`, but an absolute `accuracy` might also be possible with some careful calculations.

tomcur:
I've optimized a bit, the implementation now requires three calls to the elliptic integrals, one of which is actually a complete integral and could be optimized with a quadratically converging approximation (which I suggest we leave for a later PR, that's also relevant for the `Ellipse` shape).

tomcur:
I need to go over the math once more when I'm fresh, but the relatively stringent tests pass, while the number of iterations per incomplete elliptic integral evaluation is around 3 or 4.

src/arc.rs
```diff
@@ -171,6 +177,118 @@ fn rotate_pt(pt: Vec2, angle: f64) -> Vec2 {
...
+        // See also:
+        // https://en.wikipedia.org/w/index.php?title=Ellipse&oldid=1248023575#Arc_length
+        //
+        // The implementation here allows calculating the incomplete elliptic integral in the range
+        // 0 <= phi <= 1/2 pi (the first elliptic quadrant), so split the arc into segments in
+        // that range.
+        let mut arclen = 0.;
+
+        // The available accuracy (tolerance) is distributed over the calculation of the two
+        // incomplete and one complete elliptic integrals.
```
tomcur, line 251: The distribution of the accuracy tolerance is not very clever, but I'd rather not spend too many clock cycles on it, when those cycles could be spent iterating the convergence loops.

The worst case is when the number of complete quarter turns, evaluated as complete elliptic integrals, is very high, as the tolerance available to it decreases linearly in the number of quarter turns.

src/arc.rs
```diff
@@ -223,6 +341,206 @@ impl Mul<Arc> for Affine {
...
+
+    // These are partial terms of the inequality derived above. The multiply by (powers of) 4 are
+    // performed per iteration for computational efficiency.
+    let mut e = a - x.min(y).min(z);
+    let mut r = (accuracy / 3.).powf(2. / 3.) * e.powi(-4);
+
+    loop {
+        if 1. <= r * a.powi(5) * (1. - e / a) {
+            break;
+        }
```
tomcur, line 467: This bound checking is a non-trivial amount of the operations per iteration, and while I think this is fine, there's an interesting alternative. Iterating to a relative error is cheaper (the formula is in "Numerical computation of real or complex elliptic integrals"), perhaps calculating a conservative upper bound before the loop starts and basing a relative error bound on that is also a possibility.

dominikh:
> This does still require #378 to be merged first.

Good news!

PR #383, opened 2024-10-01
Numerically approximate ellipse perimeter

tomcur:
This implements the (truncated) infinite series for the ellipse perimeter as described by Kummer (1837) and rediscovered by Linderholm and Segal (1995), and the iterative arithmetic-geometric mean method.

In the common case of ellipses that are only moderately eccentric and for "normal" accuracy (say, 0.001), the infinite series converges quickly. The series is known at compile-time, truncated to the 6th power. For a given ellipse and accuracy, a quick check is performed at runtime, to determine whether the truncated series' approximation is within the desired accuracy. If so, the truncated series is evaluated.

If the check determines the approximation is not good enough, the problem is handed to the iterative arithmetic-geometric mean method. This method converges quadratically for all cases.

### DIFF
--- a/src/ellipse.rs
+++ b/src/ellipse.rs
@@ -304,29 +304,44 @@ fn kummer_elliptic_perimeter(accuracy: f64, radii: Vec2) -> f64 {
 
     let accuracy = accuracy / (PI * (x + y));
 
-    let mut h = (x - y).powi(2) / (x + y).powi(2);
-    let mut sum = 1.;
-    let mut m = 1;
+    let h = ((x - y) / (x + y)).powi(2);
+    let mut h_m = 1.;
+
+    let mut sum = 0.;
 
     // The binomial coefficient binom(1/2, n), calculated as
     // ∏ (1.5 - i) / i for i = 1 to n
-    let mut binom = 1.;
+    let mut binom = 1f64;
 
-    loop {
+    // An upper bound on the sum of the series remainder (see the note below)
+    let mut binom_sq_sum_remainder = 4. / PI;
+
+    for m in 1.. {
+        // Calculate the `m-1`th term
+        let binom_sq = binom.powi(2);
+        binom_sq_sum_remainder -= binom_sq;
+        sum += binom_sq * h_m;
+
+        h_m *= h;
         binom *= (1.5 - m as f64) / m as f64;
-        let binom2 = binom.powi(2);
-        let term = binom2 * h;
-        sum += term;
 
-        m += 1;
-        h *= h;
-
-        // Stop iterating when the terms become small enough.
+        // Stop iterating when the error becomes small enough.
         //
-        // Perhaps there's a somewhat stricter bound we could use, but I believe
-        // binom(1/2, x)^2 ≥ ∑ binom(1/2, n)^2 for n = x+1 to inf,
-        // and the following terms are multiplied by `h` or less.
-        if binom2 * h <= accuracy {
+        // We do not know what the remainder of the infinite series sums to, but we can calculate
+        // an upper bound:
+        //
+        // ∑ binom(1/2, n)^2 for n = 0 to inf
+        //   = 1 + (1 / 2!!)^2 + (1!! / 4!!)^2 + (3!! / 6!!)^2 + (5!! / 8!!)^2 + ..
+        //   = 4 / pi
+        //   with !! the double factorial
+        // (equation 274 in "Summation of Series", L. B. W. Jolley, 1961).
+        //
+        // This means the remainder of the infinite series for C, assuming h = 1, sums to
+        // 4 / pi - ∑ binom(1/2, n)^2 for n = 0 to m-1
+        //
+        // As 0 ≤ h ≤ 1, this is an upper bound.
+        if binom_sq_sum_remainder * h_m <= accuracy {
+            sum += binom_sq_sum_remainder * h_m;
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


tomcur:
This solves the general case up to arbitrary precision.

For ellipses that are only moderately eccentric, the series converges very quickly, such that only a few terms are needed for normal accuracy. Even for practically infinite accuracy (i.e., for Kurbo's case: within machine precision of an `f64`), only a few terms are needed when ellipses are just slightly eccentric.

Wikipedia states terms up to and including the 4th power are enough for eccentricities of 0.5 or less to reach the limits of `f64` precision. With a back-of-the-envelope calculation that seems about right: the next term would be `49 * 0.005^5 / 65536`, log2 of that is `-48`, so 48 bits away from the first term of 1.

We could expand the first few terms explicitly, only starting iteration when necessary. Though we'd have to benchmark whether that really is faster.

tomcur:
The method I first added (`kummer_elliptic_perimeter`, based on a series with binomial coefficients in n=1/2) only converges quadratically for ellipses that are not too eccentric. I've added an iterative method based on the arithmetic-geometric mean that actually does converge quadratically for all cases, requiring 7 iterations even in the case of an ellipse with radii `(0.000_000_1, 1.)` and desired accuracy `0.000_001`. The binomial series required many more terms, with the difference increasing rapidly with higher eccentricity or better accuracy.

The method based on binomial coefficients is still nice, in the sense that the coefficients can be known at compile-time. We can expand that series up to some number of terms, evaluating that in the case the approximation is good enough for the given eccentricity and accuracy. That allows a loopless approximation for the most common cases, only requiring the iterative arithmetic-geometric mean method for very high eccentricity or very precise accuracy.

tomcur:
> That allows a loopless approximation for the most common cases, using the quadratically converging iterative arithmetic-geometric mean method for very high eccentricity or very precise accuracy.

Done, and PR text updated. Perhaps we should benchmark whether this is actually enough of a performance improvement for common ellipses, to justify the additional lines of code.

src/ellipse.rs
```diff
@@ -229,15 +229,34 @@ impl Shape for Ellipse {
...
-        // TODO rather than delegate to the bezier path, it is possible to use various series
-        // expansions to compute the perimeter to any accuracy. I believe Ramanujan authored the
-        // quickest to converge. See
-        // https://www.mathematica-journal.com/2009/11/23/on-the-perimeter-of-an-ellipse/
-        // and https://en.wikipedia.org/wiki/Ellipse#Circumference
-        //
-        self.path_segments(0.1).perimeter(accuracy)
+        let radii = self.radii();
+
+        if radii.is_nan() {
```
raphlinus, line 243: Maybe change this to `!is_finite()`?

src/ellipse.rs
```diff
@@ -293,6 +452,58 @@ mod tests {
...
+    #[test]
+    fn circular_perimeter() {
+        for radius in [1.0, 0., 1.5, PI, 10.0, -1.0, 1_234_567_890.1] {
+            let circle = Circle::new((0., 0.), radius);
+            let ellipse = Ellipse::new((0., 0.), (radius, radius), 0.);
+
+            let circle_p = circle.perimeter(0.);
+            let ellipse_p = ellipse.perimeter(0.1);
+
+            assert_eq!(
```
raphlinus, line 464: Asserting equality seems fragile, shouldn't this be an approximate equality test?

src/ellipse.rs
```diff
@@ -282,9 +301,149 @@ impl Shape for Ellipse {
...
+    let h3 = h2 * h;
+    let h4 = h3 * h;
+    let h5 = h4 * h;
+    let h6 = h5 * h;
+
+    let lower = 1.
+        + h / 4.
+        + h2 / 64.
+        + h3 / 256.
+        + 25. * h4 / 16384.
```
raphlinus, line 336: These would be slightly more efficient if written as `h4 * (25. / 16384.)` etc.

raphlinus (approved):
Very cool, from what I can see the math checks out.

I think the AGM algorithm is probably very good, but for one other possibility that seems to work well in the high eccentricity case, see "Cayley" in [The Perimeter of an Ellipse](https://web.tecnico.ulisboa.pt/~mcasquilho/compute/com/,ellips/PerimeterOfEllipse.pdf). That formula requires logarithms, though, so it's not clear it will be better than AGM. Also, I suspect the accuracy of table 1, as perimeter is not monotonic.

src/ellipse.rs
```diff
@@ -282,9 +301,149 @@ impl Shape for Ellipse {
...
+    let h5 = h4 * h;
+    let h6 = h5 * h;
+    let h7 = h6 * h;
+
+    const BINOM_SQUARED_REMAINDER: f64 = 0.00101416479131503;
+    // = 4. / PI - (1. + 1. / 4. + 1. / 64. + 1. / 256. + 25. / 16384. + 49. / 65536. + 441. / 1048576.)
+
+    let remainder = BINOM_SQUARED_REMAINDER * h7;
+
+    PI * (x + y) * remainder
```
raphlinus, line 377: This can be simplified to `(PI * BINOM_SQUARED_REMAINDER) * h.powi(7)` and eliminate the intermediate h's.

tomcur:
> one other possibility that seems to work well in the high eccentricity case, see "Cayley" in [The Perimeter of an Ellipse](https://web.tecnico.ulisboa.pt/~mcasquilho/compute/com/,ellips/PerimeterOfEllipse.pdf). That formula requires logarithms, though, so it's not clear it will be better than AGM.

Very nice, that does seem to work well. It only requires evaluating `ln(4/k)`, so the cost is constant. At the very least it's a good candidate to handle the same was as `kummer_elliptic_perimeter`, as the coefficients are known at compile-time, requiring just a few additions and multiplications per term.

I wonder if the overlap between Kummer and Cayley is enough to guarantee machine precision for all eccentricities, without too many terms.

> Also, I suspect the accuracy of table 1, as perimeter is not monotonic.

Well-spotted. Having checked a few, it seems from b=0.002 onwards the reported perimeter is accidentally multiplied by 10. The decimals are otherwise correct.

src/ellipse.rs
tomcur: Ah, true. I originally developed on nightly Rust and marked those functions `const fn` (`f64::powi` isn't available on nightly yet).

src/ellipse.rs
tomcur: Good point. I've lifted `PI` into that as well, plus a similar change for `fn kummer_elliptic_perimeter_range`. Only the purely variable operations are now done at runtime: 4e5e03f4165c9ffb9ffce56b435efa40de72eeba

src/ellipse.rs
tomcur: Hmm... my intention was to return `f64::INFINITY` in case of infinite radii. I'm not sure if that's desirable though, and I think it broke somewhere along the way. Didn't add a test for it early on.

src/ellipse.rs
tomcur: Maybe. In the Kummer series expansion for the circular case, `x == y`, meaning `h == 0.`. This means the calculation becomes `(x + y) * PI` (`lower == PI` exactly).

`Circle::perimeter` calculates `2.0 * PI * self.radius`. Ellipse effectively calculates `2.0 * x * PI`.

Is this guaranteed to round the same way on all platforms? I'm not sure.

tomcur: I think it is guaranteed to round the same way, because the initial multiply-by-2 should simply increment the float's exponent. I like that the results are identical, but indeed, perhaps it's too fragile.

tomcur: I made the test approximate. Equality indeed is too fragile: even if the calculation itself would be fine, the radii are calculated through the inner affine map, so there are too many moving parts.

src/ellipse.rs
tomcur: I remember now: because ellipses are defined by an `Affine`, constructing an ellipse with infinite radii causes the ellipse radii to be calculated as NaN. I have changed the check to `!is_finite()` as you suggested, to make it robust if the inner definition of an ellipse were ever to change.

tomcur:
I think the Cayley formula is worth exploring, but will leave that for another PR. I haven't looked into it yet. In the meantime this is an improvement and is useful also for #381.

PR #387, opened 2024-10-07
Correctly calculate incenter in `Triangle::inscribed_circle`

tomcur:
The method erroneously calculated the circumcenter instead.

Tagging along on this PR are some changes to the triangle tests that I used to track down some issues. If desired I can bump these to a separate PR. The asserts testing approximate equality now take a relative epsilon, making them a bit more useful for testing values differing in orders of magnitude. I've also recalculated some of the test values to specify them at 17 significant digits (absolute limit of `f64`), and made the equality bounds tighter.

### DIFF
--- a/src/triangle.rs
+++ b/src/triangle.rs
@@ -128,7 +128,11 @@ impl Triangle {
         let bc = self.b.distance(self.c);
         let ac = self.a.distance(self.c);
 
-        Circle::new(self.circumcenter(), 2.0 * self.area() / (ab + bc + ac))
+        let perimeter_recip = 1. / (ab + bc + ac);
+        let incenter = (self.a.to_vec2() * bc + self.b.to_vec2() * ac + self.c.to_vec2() * ab)
+            * perimeter_recip;
+
+        Circle::new(incenter.to_point(), 2.0 * self.area() * perimeter_recip)
     }
 
     /// The circumscribed circle of [`Triangle`].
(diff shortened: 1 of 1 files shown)


src/triangle.rs
```diff
@@ -128,7 +128,11 @@ impl Triangle {
         let bc = self.b.distance(self.c);
         let ac = self.a.distance(self.c);
 
-        Circle::new(self.circumcenter(), 2.0 * self.area() / (ab + bc + ac))
+        let perimeter_recip = 1. / (ab + bc + ac);
+        let incenter = (self.a.to_vec2() * bc + self.b.to_vec2() * ac + self.c.to_vec2() * ab)
+            * perimeter_recip;
```
raphlinus, line 133: It might be cleaner to just divide by (ab + bc + ac) here, as the Div implementation already does the recip optimization.

raphlinus (approved):
Looks good, and thanks. I didn't carefully check the math for calculation of the incenter, but looks like you have.

src/triangle.rs
tomcur: Ah, indeed! I remember seeing that, but had forgotten it in the meantime.

In this specific case, line 135 calculates `2.0 * self.area() * perimeter_recip`. The compiler is smart enough to still only do the recip calculation once when dividing the Vec2 by `(ab + bc + ac)`, but maybe it's clearer this way anyway that there's only one division instruction.

tomcur: Added a code comment.

PR #390, opened 2024-10-08
Reduce number of operations in `Triangle::circumscribed_circle`

tomcur:
This is based on the simplified form for Cartesian coordinates in https://en.wikipedia.org/w/index.php?title=Circumcircle&oldid=1247179617#Cartesian_coordinates_2

Also adds a test to check the circle radius' sign is equal to the sign of the triangle's area.

### DIFF
--- a/src/triangle.rs
+++ b/src/triangle.rs
@@ -72,27 +72,6 @@ impl Triangle {
         (1.0 / 3.0 * (self.a.to_vec2() + self.b.to_vec2() + self.c.to_vec2())).to_point()
     }
 
-    /// The circumcenter of the [`Triangle`].
-    #[inline]
-    fn circumcenter(&self) -> Point {
-        let d = 2.0
-            * (self.a.x * (self.b.y - self.c.y)
-                + self.b.x * (self.c.y - self.a.y)
-                + self.c.x * (self.a.y - self.b.y));
-
-        let ux = ((self.a.x.powi(2) + self.a.y.powi(2)) * (self.b.y - self.c.y)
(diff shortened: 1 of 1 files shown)


src/triangle.rs
```diff
@@ -141,11 +120,17 @@ impl Triangle {
...
+        let b = self.b - self.a;
+        let c = self.c - self.a;
+        let b_len2 = b.hypot2();
+        let c_len2 = c.hypot2();
+        let d_recip = 1. / (2. * b.cross(c));
 
-        Circle::new(self.circumcenter(), (ab * bc * ac) / (4.0 * self.area()))
+        let x = (c.y * b_len2 - b.y * c_len2) * d_recip;
+        let y = (b.x * c_len2 - c.x * b_len2) * d_recip;
+        let r = b_len2.sqrt() * c_len2.sqrt() * (c - b).hypot() * d_recip;
```
raphlinus, line 131: If I'm reading this right, then a further possible optimization is to to use a single sqrt around the product of the distances squared of the sides.

raphlinus (commented):
Looks reasonable to me, I think this can be committed after the rebase.

src/triangle.rs
tomcur: Well spotted! Nice.

raphlinus (approved):
Sorry to let this slip between the cracks.

Philipp-M:
(I think you need to rebase to trigger the formatting CI, possibly because something changed on main here, Github UI is not very helpful here...)

tomcur:
> (I think you need to rebase/merge main to trigger the formatting CI, possibly because something changed on main here, Github UI is not very helpful here...)

(Yep, that triggered it. Thanks!)

PR #397, opened 2024-11-05
Match CI doc testing with docs.rs.

xStrom:
* Remove `-Zunstable-options -Zrustdoc-scrape-examples` from CI. It wasn't even enabled for docs.rs and the examples we have lead to more noise than anything in the docs, so not worth enabling scraping at this point.
* Treat doc warnings as errors. Historically we've not done so due to various `rustdoc` bugs giving false positives but I got it to pass without failure right now, so perhaps better times have arrived.
* Pass `--all-features` at docs.rs to match our CI and reduce the maintenance burden of manually syncing the features list.
* Target only `x86_64-unknown-linux-gnu` on docs.rs as we don't have any platform specific docs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,17 +260,17 @@ jobs:
         with:
           save-if: ${{ github.event_name != 'merge_group' }}
 
-      # We test documentation using nightly to match docs.rs. This prevents potential breakages
+      # We test documentation using nightly to match docs.rs.
       - name: cargo doc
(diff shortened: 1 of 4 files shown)


waywardmonkeys (approved):
You snuck a typos update in as well!

For other repos, if you look into doing the same, also need to make sure they correctly enable `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` like this crate does (if appropriate).

PR #399, opened 2024-11-08
Implement `Sum` for `Vec2`

Philipp-M:


raphlinus (approved):
Definitely makes sense, thanks!

PR #401, opened 2024-11-18
Add ConcreteShape type

PoignardAzur:
Supersedes #331.
Based on @ratmice's code in that PR.

### DIFF
--- a/src/concrete_shape.rs
+++ b/src/concrete_shape.rs
@@ -64,27 +64,40 @@ from_impl!(Triangle);
 macro_rules! match_shape {
     ($x:ident, $i:ident, $e: expr) => {
         match $x {
-            ConcreteShape::PathSeg($i) => $e,
-            ConcreteShape::Arc($i) => $e,
-            ConcreteShape::BezPath($i) => $e,
-            ConcreteShape::Circle($i) => $e,
-            ConcreteShape::CircleSegment($i) => $e,
-            ConcreteShape::CubicBez($i) => $e,
-            ConcreteShape::Ellipse($i) => $e,
-            ConcreteShape::Line($i) => $e,
-            ConcreteShape::QuadBez($i) => $e,
-            ConcreteShape::Rect($i) => $e,
-            ConcreteShape::RoundedRect($i) => $e,
-            ConcreteShape::Triangle($i) => $e,
+            Self::PathSeg($i) => $e,
+            Self::Arc($i) => $e,
+            Self::BezPath($i) => $e,
+            Self::Circle($i) => $e,
+            Self::CircleSegment($i) => $e,
+            Self::CubicBez($i) => $e,
+            Self::Ellipse($i) => $e,
+            Self::Line($i) => $e,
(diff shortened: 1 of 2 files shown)


src/concrete_shape.rs
```diff
@@ -0,0 +1,122 @@
...
+            ConcreteShape::RoundedRect($i) => $e,
+            ConcreteShape::Triangle($i) => $e,
+        }
+    };
+}
+
+impl Shape for ConcreteShape {
+    type PathElementsIter<'iter> = Box<dyn Iterator<Item = crate::PathEl> + 'iter>;
+
+    fn path_elements(&self, tol: f64) -> Box<dyn Iterator<Item = crate::PathEl> + '_> {
```
Philipp-M, line 86: You may want to look at https://github.com/linebender/xilem/pull/591/files#diff-d503d474ddd46b17b3dcb3674bf88c1654f893c21e81bbf2cf2098196b804e58R362

To avoid the boxed iterator.

PoignardAzur: Oh, right, this did seem more inefficient than I remembered.

PoignardAzur:
We'll see in practice whether people have use for it. I already know it's going to be useful in Masonry.

ratmice:
There is nothing here trait wise which requires the implementation to be in kurbo itself. IMO without supporting downstream implementations of the `Shape` trait if this is useful in Masonry, it is better to just put the implementation in Masonry.  Would it's usefulness there even require it being `pub`?

PoignardAzur:
Well we could always stick to the version introduced in https://github.com/linebender/xilem/pull/591

ratmice:
I definitely feel that I would prefer downstream crates adopting ad-hoc solutions over
inclusion of a solution in kurbo that only works with Kurbo included implementations of the trait.

That is to say, that this is working around an issue with the `Shape` trait not being object-safe,
The proposition with object safety is something like `forall t: Trait, ObjectSafe(t) -> exists dyn Trait`.
This patch by being limited to kurbo specific variants for the `Shape` impls, doesn't apply in all the cases that `forall` would.

Thus my feeling is that this patch doesn't fix the problem, it only fixes the common occurrences of the problem.
In doing so, it would seem to kill any motivation towards actually fixing the problem...

I would rather not see this included lightly based on it's convenience,
Fixing it in downstream libraries like masonry isn't convenient or ideal, 
it in fact impedes code sharing across libraries that depend upon kurbo, 
but ensures that there is motivation to fix it properly, and avoids hamstringing kurbo to something silly should we decide to re-adopt object safety for the trait or some other option...

Apologies if i'm being overly difficult in this regard...

PoignardAzur:
Overall, there's been little enough enthusiasm for this feature that I'd rather close the PR for now.

xorgy:
I just found these PRs. I made a sorta similar thing in Tabulon, but it doesn't try to be `impl Shape`, and also runs into limitations implementing things like `Mul`... https://github.com/waywardmonkeys/tabulon/blob/main/tabulon/src/shape.rs

PR #402, opened 2024-11-22
Add Linebender lint set v2.

xStrom:
Actually satisfying a good chunk of these lints is deferred and they are in an `allow` block in `lib.rs`.

Still, I did satisfy some of the lints.

The `rand` method `gen` has to be a raw `r#gen` until they publish a new version. They have that method renamed in [`master`](https://github.com/rust-random/rand).

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,7 +1,13 @@
-# The default clippy value for this is 8 bytes, which is chosen to improve performance on 32-bit.
-# Given that kurbo is being designed for the future and already even mobile phones have 64-bit CPUs,
+# LINEBENDER LINT SET - .clippy.toml - v1
+# See https://linebender.org/wiki/canonical-lints/
+
+# The default Clippy value is capped at 8 bytes, which was chosen to improve performance on 32-bit.
+# Given that we are building for the future and even low-end mobile phones have 64-bit CPUs,
 # it makes sense to optimize for 64-bit and accept the performance hits on 32-bit.
 # 16 bytes is the number of bytes that fits into two 64-bit CPU registers.
 trivial-copy-size-limit = 16
+
+# END LINEBENDER LINT SET
+
 # Don't warn about these identifiers when using clippy::doc_markdown.
 doc-valid-idents = ["Direct2D", "PostScript", ".."]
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -66,23 +66,52 @@
...
+    clippy::partial_pub_fields,
+    clippy::unseparated_literal_suffix,
+    clippy::duplicated_attributes
+)]
 
 #[cfg(not(any(feature = "std", feature = "libm")))]
 compile_error!("kurbo requires either the `std` or `libm` feature");
 
+// Suppress the unused_crate_dependencies lint when both std and libm are specified.
+#[cfg(all(feature = "std", feature = "libm"))]
```
waywardmonkeys, line 112: but this shouldn't happen?

clippy.toml
```diff
@@ -1,7 +1,13 @@
-# The default clippy value for this is 8 bytes, which is chosen to improve performance on 32-bit.
-# Given that kurbo is being designed for the future and already even mobile phones have 64-bit CPUs,
+# LINEBENDER LINT SET - .clippy.toml - v1
```
waywardmonkeys, line 1: Any chance you want to rename this file to `.clippy.toml` at the same time as the rest of this?

waywardmonkeys (approved):

clippy.toml
xStrom: Yes, good catch.

src/lib.rs
xStrom: If by *this* you mean:
* both `std` and `libm` together -- well, it easily can, e.g. by using `--all-features`
* `libm` being unused when the `libm` feature is specified -- well, `std` overrides it so it won't be used

waywardmonkeys: But the case of `--all-features` will be a compile error because of the lines directly above this?

waywardmonkeys: I'm not trying to be difficult, just want to understand and I'm kind of dumb at the moment.

xStrom: No it won't be a compiler error. The compiler error we triger is only when neither `std` nor `libm` is specified. At least one of those needs to be set. You can achieve that error with just `--no-default-features`.

waywardmonkeys: I am so brain dead right now. Going to sleep in shame and infamy.

PR #409, opened 2024-12-23
Fix documentation of cross product; add convenience Vec2 methods

raphlinus:
I managed to get this wrong when I was trying to nail down the sign conventions, and it's been bothering me since I noticed it.

The PR adds a couple of simple convenience methods I find myself using quite a bit as I implement stroke expansion.

Note that a lot of methods can become const when we bump the MSRV, but that should happen separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,16 @@ This release has an [MSRV][] of 1.65.
 - `Stroke` is now `PartialEq`, `StrokeOpts` is now `Clone`, `Copy`, `Debug`, `Eq`, `PartialEq`. ([#379][] by [@waywardmonkeys][])
 - Implement `Sum` for `Vec2`. ([#399][] by [@Philipp-M][])
 - Add triangle shape. ([#350][] by [@juliapaci][])
+- Add `Vec2::turn_90` and `Vec2::rotate_scale` methods ([#409] by [@raphlinus][])
 
 ### Changed
 
 - Reduce number of operations in `Triangle::circumscribed_circle`. ([#390][] by [@tomcur][])
 - Numerically approximate ellipse perimeter. ([#383] by [@tomcur][])
+
+### Fixed
+
+- Fix documentation of cross product. ([#409] by [@raphlinus][])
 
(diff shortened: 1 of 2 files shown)


src/vec2.rs
```diff
@@ -314,6 +320,32 @@ impl Vec2 {
...
+    pub fn turn_90(self) -> Vec2 {
+        Vec2::new(-self.y, self.x)
+    }
+
+    /// Combine two vectors interpreted as rotation and scaling.
+    ///
+    /// Interpret both vectors as a rotation and a scale, and combine
+    /// their effects. This operation is equivalent to multiplication
+    /// when the vectors are interpreted as complex numbers. It is
+    /// commutative.
```
tomcur, line 341: It might be worth it to state explicitly that the angles are added, and the magnitudes are multiplied.

Maybe:

```suggestion
    /// Combine two vectors interpreted as rotation and scaling.
    ///
    /// Interpret both vectors as a rotation and a scale, and combine
    /// their effects by adding the angles and multiplying the magnitudes.
    /// This operation is equivalent to multiplication when the vectors
    /// are interpreted as complex numbers. It is commutative.
```

CHANGELOG.md
```diff
@@ -20,12 +20,17 @@ This release has an [MSRV][] of 1.65.
 - `Stroke` is now `PartialEq`, `StrokeOpts` is now `Clone`, `Copy`, `Debug`, `Eq`, `PartialEq`. ([#379][] by [@waywardmonkeys][])
 - Implement `Sum` for `Vec2`. ([#399][] by [@Philipp-M][])
 - Add triangle shape. ([#350][] by [@juliapaci][])
+- Add `Vec2::turn_90` and `Vec2::rotate_scale` methods ([#409] by [@raphlinus][])
```
tomcur, line 23: The link to `#409` is missing at the bottom of the file.

tomcur (approved):
Math looks correct. There's a missing link in the changelog and I added a docs suggestion inline.

> Should it be two separate PRs?

It could be two PRs, or maybe just update the title to indicate methods are added.

PR #412, opened 2025-01-22
Add `min` and `max` methods to `Size`.

nils-mathieu:
This pull request adds a `min` and `max` method to the `Size` type.

# Rational

I found myself getting often the minimum of two size component-wise and thought I'd be nicer to have a `min` and `max` method added to the type directly.

Those two methods go in pair with the already existing `.clamp` method, which does exactly that, but with a minimum and a maximum bound.

### DIFF
--- a/src/size.rs
+++ b/src/size.rs
@@ -81,6 +81,44 @@ impl Size {
         self.is_zero_area()
     }
 
+    /// Returns the component-wise minimum of `self` and `other`.
+    ///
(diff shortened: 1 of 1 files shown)


jneem (approved):
I can never remember if I have write access here, but lgtm!

waywardmonkeys:
@jneem You do! But I'll merge this for now.

PR #413, opened 2025-01-22
Add an `INIFNITY` constant to `Size`.

nils-mathieu:
This pull request adds the `INFINITY` constant to `Size`.

```rust
const INFINITY: Size = Size::new(f64::INFINITY, f64::INFINITY);
```

I found myself writing a lot of infinite sizes manually to represent "no constraint", and thought it would be cleaner to have it as an associated constant instead.

# Potential questions

Is there any other constants we might want for `Size`? `INFINITY` is the only one I personally needed, but maybe there are others.

### DIFF
--- a/src/size.rs
+++ b/src/size.rs
@@ -27,6 +27,9 @@ impl Size {
     /// A size with zero width or height.
     pub const ZERO: Size = Size::new(0., 0.);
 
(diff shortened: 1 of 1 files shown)


jneem (approved):
Let's wait on other constants until someone actually wants them.

PR #414, opened 2025-01-31
Update rand to 0.9

waywardmonkeys:


DJMcNab (approved):
Obviously I'm not a massive fan of being forced to add `.cargo`, I agree that there probably isn't a better alternative. What a mess

PR #416, opened 2025-02-14
Update typos to 1.29.7, fix spelling of "collinear"

waywardmonkeys:
This was news to me about how that word is spelled. Interesting.


DJMcNab (approved):
All the signals I can find indicate that this is the correct spelling.
Also, this is not "public" facing, so there's no risk to correcting it.

PR #417, opened 2025-02-15
Add `tangents_to_point` for `QuadBez`

Keavon:
Follows up #288 which added `tangents_to_point()` to `CubicBez`, by adding it to `QuadBez` now as well.

### DIFF
--- a/src/quadbez.rs
+++ b/src/quadbez.rs
@@ -7,7 +7,7 @@ use core::ops::{Mul, Range};
 
 use arrayvec::ArrayVec;
 
-use crate::common::solve_cubic;
+use crate::common::{solve_cubic, solve_quadratic};
 use crate::MAX_EXTREMA;
 use crate::{
     Affine, CubicBez, Line, Nearest, ParamCurve, ParamCurveArclen, ParamCurveArea,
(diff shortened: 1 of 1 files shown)


src/quadbez.rs
```diff
@@ -110,6 +110,27 @@ impl QuadBez {
...
+    /// evaluated at that point goes through the argument point.
+    pub fn tangents_to_point(&self, p: Point) -> ArrayVec<f64, 2> {
+        let a = self.p0.to_vec2() - 2.0 * self.p1.to_vec2() + self.p2.to_vec2();
+        let b = 2.0 * (self.p1.to_vec2() - self.p0.to_vec2());
+        let c = self.p0.to_vec2() - p.to_vec2();
+
+        // coefficients of x(t) \cross x'(t)
+        let c2 = a.cross(b);
+        let c1 = -2.0 * c.cross(a);
+        let c0 = b.cross(c);
```
Keavon, line 127: I'd like a double check on the math here, both in terms of its correctness (I'm not certain of it) and its formatting (is it more faithful to the canonical Bézier equations to reverse the orders of the terms, swap the order of the cross product and negate it, etc.?). This mostly arose from pattern recognition based fiddling and I wasn't able to quite grasp its relationship with the Bézier equations on Wikipedia. So please confirm this is right.

waywardmonkeys: I think this would be a little clearer:

```
      // Coefficients of (c + b t + a t^2) x (b + 2 a t) = 0.
      let c2 = b.cross(a);
      let c1 = 2.0 * c.cross(a);
      let c0 = c.cross(b);
```

PR #422, opened 2025-02-20
ci: update to stable Rust 1.85, typos 1.29.9

waywardmonkeys:


kurbo/src/rounded_rect.rs
```diff
@@ -8,6 +8,7 @@ use core::ops::{Add, Sub};
 
 use crate::{arc::ArcAppendIter, Arc, PathEl, Point, Rect, RoundedRectRadii, Shape, Size, Vec2};
 
+#[allow(unused_imports)] // This is unused in later versions of Rust because of additions to core::f32
```
tomcur, line 11: Maybe we should mention this can be removed when the MSRV is 1.86 -- or potentially earlier, I haven't checked, but a more explicit note to help future-us understand when this can be cleaned would be good.

tomcur (approved):

tomcur (approved):

Issue #317, opened 2023-10-12
Speed up stroke expansion


raphlinus:
Stroke expansion is pretty slow. One aspect of this issue is doing some measurements to see how it compares to other solutions, and also where the time is going. There may be some low hanging fruit.

That said, even without measurement there are some ideas. A lot of time goes into error measurement after fitting to see whether the resulting offset is within tolerance. I bet we could do an analytical error metric (likely based on estimates of min/max curvature) that would predict upper and lower bounds on error. If the upper bound is within tolerance, it can be based on just the fit, no need to measure. If the lower bound isn't, then we subdivide without needing to do the fit and measure.

Additionally, fitting based on "balanced Béziers" ([Zulip thread](https://xi.zulipchat.com/#narrow/stream/260979-kurbo/topic/Sweet.20B.C3.A9ziers)) is likely to be faster. The numerical integration of the curve to be fit (the parallel curve of the source Bézier) needs to compute only area, not also moment. The solver is quadratic, not quartic, which should be *much* faster. The only trick is that it might need more subdivisions; fitting the source curve even with infinitesimal offset is not guaranteed, while it is for the quartic-based solver. It's trivial to detect the imbalance of the source curve. Perhaps we apply the analytical error metrics to see whether it can still be fit by a single cubic. If it needs to be subdivided, then perhaps the quadratic fit for the subdivisions is good enough.

This would be a great project for someone who's handy at math.

raphlinus:
One other thing to add: an error metric based on curvature (and, now that I think about it, possibly parametrization balance as well) would no doubt also be useful in setting the number of samples needed for numerical integration. That's currently [hardcoded at 16](https://github.com/linebender/kurbo/blob/c7cb529bd0ebd95f2fffd368813bc8fb3fd76344/src/fit.rs#L81), not based on any kind of rigorous analysis, it just seemed like a reasonable value.

raphlinus:
[...]

PR #427, opened 2025-03-19
First cut at least-squares offset curves

raphlinus:
Replace the quartic solver with the new least-squares one.

The robustness story is not complete. For now, it relies on the existing "regularize" hack, but it would be better to be robust by construction.

The stroke options are not respected. It tries to do a better job with subdivision but does not attempt an absolutely minimal number of segments as before. That could be re-added.

The old offset methods are still exported. They should probably be deprecated.

Progress toward #317


tomcur (commented):
Some initial observations inline based on a first dive into this.

LaurenzV:
See my Zulip message for two suggestions for performance improvements: https://xi.zulipchat.com/#narrow/channel/260979-kurbo/topic/New.20stroker.20for.20sparse.20strips/near/525882812

kurbo/src/offset.rs
```diff
@@ -161,6 +169,587 @@ impl ParamCurveFit for CubicOffset {
...
+    fn cusp_sign(&self, t: f64) -> f64 {
+        let ds2 = self.q.eval(t).to_vec2().hypot2();
+        ((self.c2 * t + self.c1) * t + self.c0) / (ds2 * ds2.sqrt()) + 1.0
+    }
+
+    /// Compute cusp value of endpoint.
+    ///
+    /// The y parameter should be c0 for the start point, and c0 + c1 + c2
+    /// for the end point; it's just evaluating the polynomial at t=0 and
+    /// t=1.
```
tomcur, line 320:
```suggestion
    /// Compute cusp value of endpoint.
    ///
    /// This is a special case of [`Self::cusp_sign`]. For the start point, `tan` should be
    /// the start point tangent and `y` should be `c0`. For the end point, `tan` should be
    /// the end point tangent and `y` should be `c0 + c1 + c2`.
    ///
    /// This is just evaluating the polynomial at t=0 and t=1.
```

kurbo/src/offset.rs
```diff
@@ -161,6 +169,587 @@ impl ParamCurveFit for CubicOffset {
...
+struct CubicOffset2 {
+    /// The cubic being offset. This has been regularized.
+    c: CubicBez,
+    /// The derivative of `c`.
+    q: QuadBez,
+    /// The offset distance (same as the argument).
+    d: f64,
+    /// c0 + c1 t + c2 t^2 is the cross product of second and first
+    /// derivatives of the underlying cubic, multiplied by offset (for
+    /// computing cusp).
```
tomcur, line 182:
```suggestion
    /// `c0 + c1 t + c2 t^2` is the cross product of second and first
    /// derivatives of the underlying cubic, multiplied by the offset.
    /// This is used for computing cusps on the offset curve.
    ///
    /// Note that given a curve `c(t)`, its signed curvature is
    /// `c''(t) x c'(t) / ||c'(t)||^3`. See also [`Self::cusp_sign`].
```

tomcur (approved):
The additional docstrings are great.

I've added a few suggestions to clarify curvature/cusps.

kurbo/src/offset.rs
DJMcNab, line 1: We should probably also add a deprecation warning to the module-level docs (probably not using the `deprecated` attribute, but I'm not sure either way)

It's also not clear if those module level docs are now out of date; is the algorithm used still the one documented in that comment?

DJMcNab (commented):
I've reviewed here wit han eye towards 0.11.3/0.12.0. This PR has clearly done what it can to be non-breaking, which is good. But it leaves a lot of things in a weird intermediate state as a result.

These can all be addressed post-merge if needed.

PR #428, opened 2025-03-22
Always inline trivial casts, splats, and swizzles.

xorgy:
These functions are not even code, they only express changes in abstract type. Therefore, they should always be inlined (even in debug builds) because, being infallible, they will never be debugged.

rustc does not currently inline/eliminate these in debug builds, so they tend to contribute considerably to the size and slowness of debug builds.

### DIFF
--- a/src/affine.rs
+++ b/src/affine.rs
@@ -44,20 +44,20 @@ impl Affine {
     /// formulation of affine transformation as augmented matrix. The
     /// idea is that `(A * B) * v == A * (B * v)`, where `*` is the
     /// [`Mul`] trait.
-    #[inline]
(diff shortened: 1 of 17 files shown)


xorgy:
I've included some casts/swizzles in here that technically involve constructing an enum, I can remove those if that's a step too far.

waywardmonkeys (approved):
I'm in agreement with this PR and I don't feel that it goes too far.

I think Raph should chime in as well as to whether he agrees or disagrees.

This will need a changelog entry.

Perhaps we should (after this) start on some design notes for Kurbo and include a section on the inlining policy. But that's for later for sure.

PR #429, opened 2025-03-22
Add `Affine::scale_about`.

xorgy:


xorgy:
This is commonly used in applications where a user is zooming in and out of something at a specific point. e.g. here:
https://github.com/waywardmonkeys/tabulon/blob/f275bdcf7cc1246ca26ebe8de063aabe829e82c0/examples/vello_viewer/src/main.rs#L252-L254

waywardmonkeys (commented):
I think this looks good.

I think the comment could indicate how this is used to help people find the method more readily.

Should we also have one for non-uniform scale? (Is the current API consistent in having a non-uniform version of every uniform scale function?)

Also, we'll need a changelog entry.

platlas:
> Should we also have one for non-uniform scale? (Is the current API consistent in having a non-uniform version of every uniform scale function?)

I'm only using uniform one but I can imagine that non-uniform could be used for something like zooming octaves in DAW's piano-roll.

xorgy:
Alright, I've added something to the doc comment about _why_ you might want this.

waywardmonkeys (approved):

PR #430, opened 2025-03-24
Update CHANGELOG

waywardmonkeys:


DJMcNab (approved):
LGTM. I've checked that everything worth including merged since #406 is included.

#407 is skipped because @waywardmonkeys thinks that the relevant code hasn't been released.

PR #433, opened 2025-03-26
Improve doc comment on `Shape::contains`

### LINKED ISSUE #432: Containment check with evenodd fill
Is it possible to check if a shape contains a point using even-odd fill? If not, how hard would it be to add it?

waywardmonkeys:
Fixes #432.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -137,9 +137,16 @@ pub trait Shape {
     /// [winding number]: https://mathworld.wolfram.com/ContourWindingNumber.html
     fn winding(&self, pt: Point) -> i32;
 
-    /// Returns `true` if the [`Point`] is inside this shape.
+    /// Returns `true` if the [`Point`] is inside this shape using the non-zero
+    /// winding rule.
     ///
-    /// This is only meaningful for closed shapes.
+    /// This is only meaningful for closed shapes. Some shapes may have specialized
+    /// implementations of this function or of [`winding`] determination.
+    ///
+    /// To determine containment using the even-odd winding rule, check the
+    /// [`winding`] directly.
+    ///
+    /// [`winding`]: Self::winding
     fn contains(&self, pt: Point) -> bool {
         self.winding(pt) != 0
     }


src/shape.rs
```diff
@@ -137,9 +137,16 @@ pub trait Shape {
     /// [winding number]: https://mathworld.wolfram.com/ContourWindingNumber.html
     fn winding(&self, pt: Point) -> i32;
 
-    /// Returns `true` if the [`Point`] is inside this shape.
+    /// Returns `true` if the [`Point`] is inside this shape using the non-zero
+    /// winding rule.
     ///
-    /// This is only meaningful for closed shapes.
+    /// This is only meaningful for closed shapes. Some shapes may have specialized
+    /// implementations of this function or of [`winding`] determination.
```
tomcur, line 144: All shapes must implement `Shape::winding`, so perhaps this can just say ``Some shapes may have specialized implementations of this function rather than calculating [`winding`].``

tomcur (approved):
One minor comment inline.

src/shape.rs
waywardmonkeys: I don’t think any actually override `contains` instead overriding `winding`.

waywardmonkeys: I should edit it to say that by default it is the non-zero winding rule since an override might not be the same.

mattfbacon:
I would add this to `contains`:

```rust
/// This is non-zero containment. To check even-odd containment, use `windings() % 2 != 0`.
```

And to `windings`:

```rust
/// This can be used to check for containment.
/// To check non-zero containment, use `windings() != 0`.
/// To check even-odd containment, use `windings() % 2 != 0`.
```

kurbo/src/shape.rs
```diff
@@ -139,7 +139,15 @@ pub trait Shape {
...
     /// Returns `true` if the [`Point`] is inside this shape.
     ///
-    /// This is only meaningful for closed shapes.
+    /// This is only meaningful for closed shapes. Some shapes may have specialized
+    /// implementations of this function or of [`winding`] determination.
+    ///
+    /// The default implementation uses the non-zero winding rule.
+    ///
+    /// To determine containment using the even-odd winding rule, check the
+    /// [`winding`] directly.
```
DJMcNab, line 148: We should probably just guarantee it uses the non-zero winding rule. I don't think hedging here is that meaningful (that would however make this a changelog worthy change, and maybe *technically* breaking).

DJMcNab (approved):
This is an improvement. I think it's correct to observe that we could lay this out even more clearly, but giving a hint here is good enough.

I do think that changing the semantics here would be worthwhile, but I wouldn't block this PR on that.

PR #434, opened 2025-04-06
Make `kurbo` a crate within a workspace

waywardmonkeys:
This lays the groundwork for having more crates within this workspace.


Cargo.toml
```diff
@@ -82,44 +79,5 @@ clippy.redundant_feature_names = "warn"
...
-
-# This is used for research but not really needed; maybe refactor.
-[dev-dependencies]
-rand = "0.9.0"
-
-[target.wasm32-unknown-unknown.dev-dependencies]
-# We have a transitive dependency on getrandom and it does not automatically
-# support wasm32-unknown-unknown. We need to enable the wasm_js feature.
-getrandom = { version = "0.3.1", features = ["wasm_js"] }
+[workspace.dependencies]
```
DJMcNab, line 82: I know our "standard" is to put `workspace.dependencies` after `workspace.lints`, but I wonder if doing the other way around would be clearer; I don't recall having any discussion about this.

Not a blocking concern

Cargo.toml
```diff
@@ -1,25 +1,22 @@
...
+resolver = "2"
+members = ["kurbo"]
+
+[workspace.package]
+# Kurbo version, also used by other packages which want to mimic Kurbo's version.
+#
+# NOTE: When bumping this, remember to also bump the aforementioned other packages'
+#       version in the dependencies section at the bottom of this file.
 version = "0.11.1"
-authors = ["Raph Levien <raph.levien@gmail.com>"]
```
DJMcNab, line 4: I see that this is not added back; that's fine and is probably the right thing to do, but it's worth calling it out explicitly.

DJMcNab (approved):
One other question, which is what's the plan for the root README. That doesn't need to be dealt with in this PR.

tomcur (approved):

Cargo.toml
waywardmonkeys: Putting it back in and we can do it explicitly.

Cargo.toml
waywardmonkeys: Agreed, moving it in this PR.

waywardmonkeys:
> One other question, which is what's the plan for the root README. That doesn't need to be dealt with in this PR.

Like all of our other repos, I don't have one yet. :(

DJMcNab (approved):

PR #435, opened 2025-04-08
Sketch out a constraint solver crate

waywardmonkeys:
This proposes the creation of a `kurbo_constraints` crate which is useful for CAD-style constraints (also useful in vector editors, graphics editors, etc.).

These would include:

* Coincident points
* Parallel lines
* Perpendicular lines
* Maintaining an angle
* Point on curve
* Tangents

There's a bunch of open questions ...

And things that will need to be added to this interface:
* Mark points as fixed / variable so the solver doesn't try to move everything
* Bring over the SVG debug visualization
* Name points for debug visualization and error reporting
* Provide error reporting
* Think about how to specify lines, curves, etc. (I have a couple of approaches locally.)

Right now, I've been writing with the Newton-Raphson method in the constraint implementations, but this may not be sufficient and we should consider multiple solver strategies. As such, it may make sense to rename `Solver` to `Problem` or something and allow for different solvers to tackle a given problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ env:
   RUST_MIN_VER: "1.65"
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
-  RUST_MIN_VER_PKGS: "-p kurbo"
+  RUST_MIN_VER_PKGS: "-p kurbo -p kurbo_constraints"
   # List of features that depend on the standard library and will be excluded from no_std checks.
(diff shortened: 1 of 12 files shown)


xorgy (approved):
I think this is a good start, and I don't see any glaring issues with it.

tomcur:
> I think this is superseded by [fiksi](https://github.com/endoli/fiksi)

That's right!

waywardmonkeys:
Yes… this was the proof of concept to convince me it was reasonable to do and get Fiksi going. :)

PR #450, opened 2025-06-12
Conversions between CoreGraphics and Kurbo types

waywardmonkeys:
This is an optional feature, not enabled by default.

### DIFF
--- a/kurbo/Cargo.toml
+++ b/kurbo/Cargo.toml
@@ -11,9 +11,8 @@ rust-version.workspace = true
 
 [package.metadata.docs.rs]
 all-features = true
-# There are no platform specific docs.
 default-target = "x86_64-unknown-linux-gnu"
-targets = []
+targets = ["aarch64-apple-darwin"]
 
 [features]
 default = ["std"]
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
This is most usable when it is able to use `From` impls, so having it here in the `kurbo` crate is easiest to use.

This is still missing support for various types.

It might be that the features should change so that they're `objc2-core-foundation` for the core types (affine, point, rect, size, vec) and `objc2-core-graphics` for the things like line caps, joins, paths, etc. There should also be support for `UIBezierPath` and `NSBezierPath` (which are 2 other crates).

kurbo/Cargo.toml
```diff
@@ -42,6 +43,18 @@ optional = true
...
+version = "0.3.1"
+optional = true
+default-features = false
+features = ["CFCGTypes"]
+
+[dependencies.objc2-core-graphics]
+version = "0.3.1"
+optional = true
+default-features = false
+features = ["CGPath"]
```
xStrom, line 56:
```suggestion
[target.'cfg(target_vendor = "apple")'.dependencies.objc2-core-foundation]
version = "0.3.1"
optional = true
default-features = false
features = ["CFCGTypes"]

[target.'cfg(target_vendor = "apple")'.dependencies.objc2-core-graphics]
version = "0.3.1"
optional = true
default-features = false
features = ["CGPath"]
```

kurbo/src/lib.rs
```diff
@@ -142,6 +142,9 @@ mod translate_scale;
 mod triangle;
 mod vec2;
 
+#[cfg(feature = "objc2-core-graphics")]
+mod interop_objc2_core_graphics;
```
xStrom, line 146:
```suggestion
#[cfg(all(target_vendor = "apple", feature = "objc2-core-graphics"))]
mod interop_objc2_core_graphics;
```

xStrom (commented):
To solve this at the CI level we would have to use `cargo-hack` for everything, including testing. However, I think we should solve this at a different level because being able to use `--all-features` locally for development is very nice.

What I think we should do is make this feature a `NOP` on other platforms. Then `--all-features` will continue working and no changes will be needed to CI either.

Issue #443, opened 2025-04-30
More moments


simoncozens:
`moment_integrals` currently returns y dx, x y dx, and y^2 dx integrals for the curve; to find centre of mass, variance, covariance, slant, etc. it would be useful to also have a full set of x/y/x^2/y^2/xy integrals. (See e.g. [fontTools.pens.statisticsPen](https://github.com/fonttools/fonttools/blob/1c2704dfc79d753fe822dc7699b067495c0edede/Lib/fontTools/pens/statisticsPen.py#L80-L88).)

raphlinus:
Something like this is very feasible, but just adding more moments doesn't feel like the right thing, as doing extra calculation could potentially slow down curve fitting. I think an expedient thing would be adding another method, but I'm wondering if the right thing is to have some kind of generic parameter to indicate which moments will be calculated. I'm open to ideas.

Part of my thinking is that you can also imagine an O(n^5) curve fitting algorithm which just uses area. So the ability to specify which moments are desired at fine grain seems like it may have value.

rsheeter:
+1 to allow the user to specify what they want, that feels right for a general purpose library.

simoncozens:
I got stuck on this because of your mention of a "generic" parameter. If you genuinely mean `moment_integrals<T>(&self, range: Range<f64>) -> T` then I've got no idea how to specify `T` in a way that allows all the different combinatorial possibilities of moments that a user might or might not want. If on the other hand you mean `moment_integrals(&self, range: Range<f64>, which_moments: MomentFlags) -> Moments` where all the fields of `Moments` are `Option<f64>` then sure, I can do that.

behdad:
I think the moments are implemented using Green's theorem and with numeric integration?

Would be nice if there was a trait that could be implemented non-generically for Bezier's up to third degree, using the analytical solution. The analytical solution can be generated using `fontTools.misc.symfont` with a Rust output option.

raphlinus:
I think the idea of making the methods generic over which set of moments is probably unnecessary complication, let's back off of that. I have some ideas of how to express that in the Rust type system, but not fully worked out, and at this point I don't think it's worth the effort. There *is* followup work to be done on curve fitting, but it should be disentangled from this. While the current approach (`ParamCurveFit`) is heavily based on moments, it's likely we'll move away from that, towards least squares error minimization.

[...]

PR #452, opened 2025-06-16
Add Green’s theorem moments derived from fontTools

simoncozens:
See discussion in #443.

### DIFF
--- a/kurbo/src/green.rs
+++ b/kurbo/src/green.rs
@@ -1,0 +1,590 @@
+use std::ops::Add;
+
+use crate::{PathEl, Point};
+
+/// A collect of moment integrals for a path, used for computing
+/// shape properties like area, centroid, and inertia.
+///
+/// The moments are computed using Green's theorem, which relates the
+/// integrals of a function over a closed curve to the integrals of its
+/// partial derivatives over the region enclosed by the curve.
+#[derive(Debug, Default, Copy, Clone)]
+pub struct GreenMomentIntegrals {
+    /// The first moment of area about the x-axis.
+    pub moment_x: f64,
+    /// The first moment of area about the y-axis.
+    pub moment_y: f64,
+    /// The second moment of area about the x-axis (also known as the
+    /// area moment of inertia about the x-axis).
+    pub moment_xx: f64,
+    /// The second moment of area about the xy-axis (also known as the
+    /// product of inertia).
+    pub moment_xy: f64,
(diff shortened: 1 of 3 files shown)


kurbo/src/green.rs
```diff
@@ -0,0 +1,595 @@
...
+    /// The second moment of area about the xy-axis (also known as the
+    /// product of inertia).
+    pub moment_xy: f64,
+    /// The second moment of area about the y-axis (also known as the
+    /// area moment of inertia about the y-axis).
+    pub moment_yy: f64,
+}
+
+#[cfg(feature = "std")]
+impl std::ops::Add<Self> for GreenMomentIntegrals {
```
jneem, line 33: Could this be `core::ops::Add` without the conditional compilation?

kurbo/src/green.rs
```diff
@@ -0,0 +1,595 @@
...
+use crate::common::FloatFuncs;
+
+/// A collect of moment integrals for a path, used for computing
+/// shape properties like area, centroid, and inertia.
+///
+/// The moments are computed using Green's theorem, which relates the
+/// integrals of a function over a closed curve to the integrals of its
+/// partial derivatives over the region enclosed by the curve.
+#[derive(Debug, Default, Copy, Clone)]
+pub struct GreenMomentIntegrals {
```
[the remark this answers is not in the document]
simoncozens: The reason I was more explicit with the naming here is that there is already a *different* moments calculation in `ParamCurveFit` and it would make sense to distinguish between the two.

jneem:
This was discussed and approved in office hours, but @raphlinus noticed a typo he wanted to fix...

kurbo/src/moments.rs
```diff
@@ -0,0 +1,594 @@
...
+            moment_xx: self.moment_xx + rhs.moment_xx,
+            moment_xy: self.moment_xy + rhs.moment_xy,
+            moment_yy: self.moment_yy + rhs.moment_yy,
+        }
+    }
+}
+
+impl Moments {
+    // The code below is derived from fontTools's fontTools.pens.momentsPen, which is in turn generated
+    // by running sympy on the 's theorem integral expressions.
```
raphlinus, line 48: This should say "Green's theorem integral expressions."

raphlinus (approved):
Looks good modulo the comment typos. This is useful functionality and I think the API additions are not very intrusive. Thanks!

jneem (approved):

PR #454, opened 2025-06-25
Properly handle negative dash offset by normalization

sagudev:
We normalize dash offset to minimal positive offset using a period.

This should also reduce number of executions of `while dash_remaining < 0.0` for big offsets.


kurbo/src/stroke.rs
```diff
@@ -650,6 +650,18 @@ fn dash_impl<T: Iterator<Item = PathEl>>(
...
     dashes: &[f64],
 ) -> DashIterator<'_, T> {
+    fn rem_euclid(lhs: f64, rhs: f64) -> f64 {
+        let r = lhs % rhs;
+        if r < 0.0 {
+            r + rhs.abs()
+        } else {
+            r
+        }
+    }
```
PoignardAzur, line 660: Any reason not to use [`f64::rem_euclid`](https://doc.rust-lang.org/std/primitive.f64.html#method.rem_euclid)? It's available at our MSRV.

sagudev: I used it before, but no_std checks failed because it's not available in core: https://github.com/rust-lang/rust/issues/137578#issuecomment-3003851171

Issue #437, opened 2025-04-09
(reasonably) efficient `ParamCurveNearest` on `BezPath`


xorgy:
`ParamCurveNearest` is not implemented for `BezPath`, and running it piecewise on `PathSeg` wastes a lot of effort computing `t` on segments between the start and the end of the path (and `t` not within a segment is never used on a closed path).

Also, you often don't care about `t` anyway, which may be an indication that a new trait for `dist_sq` alone is the more important.

PR #456, opened 2025-06-25
Add `point_on_shape`

### LINKED ISSUE #455: Add `point_on_shape`
As discussed in #453, servo needs `point_on_shape(&self, pt: Point) -> bool` on `Shape trait`, that would return true iff the point is one edge of path (it's on bounding curves).

sagudev:
For most of shapes we simply create path segments with provided tolerance and check if point lies on any of the segments (within the same tolerance). This means we need real implementation at least for `PathSeg` and consequently for `Line`, `Curve`, `QuadCurve`, using `nearest`. In the future we could create more specialized implementations.

Fix #455

### DIFF
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -820,6 +820,10 @@ impl<I: Iterator<Item = PathEl>> Segments<I> {
         self.map(|seg| seg.winding(p)).sum()
     }
 
+    pub(crate) fn point_on_shape(mut self, p: Point, tolerance: f64) -> bool {
+        self.any(|seg| seg.point_on_shape(p, tolerance))
+    }
+
     // Same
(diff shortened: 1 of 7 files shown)


jneem:
I don't quite understand why we need a new API for this. Couldn't we just implement `ParamCurveNearest` for `BezPath` and `Circle`? Then you could just test `path.nearest(p).distance_sq.sqrt() < tolerance`. (#437 seems related)

sagudev:
> I don't quite understand why we need a new API for this. Couldn't we just implement `ParamCurveNearest` for `BezPath` and `Circle`? Then you could just test `path.nearest(p).distance_sq.sqrt() < tolerance`. (#437 seems related)

Thanks for the related issue, with it I could already implemented this in servo.

PR #459, opened 2025-06-27
Move `rem_euclid` impl to `FloatFuncs`

sagudev:
kurbo already have mechanism to provide missing no_std impls of float methods using `FloatFuncs`, but I didn't know that in #454.

### DIFF
--- a/kurbo/src/common.rs
+++ b/kurbo/src/common.rs
@@ -31,8 +31,9 @@ macro_rules! define_float_funcs {
         /// For documentation see the respective functions in the std library.
         #[cfg(not(feature = "std"))]
         pub trait FloatFuncs : Sized + sealed::FloatFuncsSealed {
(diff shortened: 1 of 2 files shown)


kurbo/src/common.rs
```diff
@@ -31,8 +31,9 @@ macro_rules! define_float_funcs {
         /// For documentation see the respective functions in the std library.
         #[cfg(not(feature = "std"))]
         pub trait FloatFuncs : Sized + sealed::FloatFuncsSealed {
-            /// Special implementation for signum, because libm doesn't have it.
+            /// Special implementations, because libm doesn't have them.
```
waywardmonkeys, line 34: Just duplicate this comment for each. I'd prefer if we linked to the docs for these from the std library directly. Also, worth keeping the link to the core floats issue along with the rem_euclid comment.

jneem (approved):

PR #462, opened 2025-07-04
Add `BezPath.current_position()`

sagudev:
Sometimes it is useful to obtain current position (sometimes also called last point/current point) of path. That's the one that also takes ClosePath into account. Implementation is very similar to [`PathEl.end_point`](https://docs.rs/kurbo/0.11.2/kurbo/enum.PathEl.html#method.end_point), except that it handles close path.

### DIFF
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -393,6 +393,27 @@ impl BezPath {
         cbox.unwrap_or_default()
     }
 
+    /// Returns current position in the path, if path is not empty.
+    ///
+    /// Unlike the `end_point` method, this also handles [`PathEl::ClosePath`],
+    /// by finding the first point of subpath, hence the time complexity is O(n).
+    pub fn current_position(&self) -> Option<Point> {
+        match self.0.last()? {
+            PathEl::MoveTo(p) => Some(*p),
+            PathEl::LineTo(p1) => Some(*p1),
+            PathEl::QuadTo(_, p2) => Some(*p2),
+            PathEl::CurveTo(_, _, p3) => Some(*p3),
+            PathEl::ClosePath => self
+                .elements()
+                .iter()
+                .rev()
+                .skip(1)
+                .take_while(|el| !matches!(el, PathEl::ClosePath))
+                .last()
+                .and_then(|el| el.end_point()),
+        }
+    }
+
     /// Returns a new path with the winding direction of all subpaths reversed.
     pub fn reverse_subpaths(&self) -> BezPath {
         let elements = self.elements();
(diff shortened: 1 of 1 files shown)


kurbo/src/bezpath.rs
```diff
@@ -393,6 +393,17 @@ impl BezPath {
...
     }
 
+    /// Get the last point of the path, if path is not empty.
+    pub fn last_point(&self) -> Option<Point> {
+        match self.0.last()? {
+            PathEl::MoveTo(p) => Some(*p),
+            PathEl::LineTo(p1) => Some(*p1),
+            PathEl::QuadTo(_, p2) => Some(*p2),
+            PathEl::CurveTo(_, _, p3) => Some(*p3),
+            PathEl::ClosePath => segments(self).last_point(),
```
jneem, line 403:
```suggestion
            PathEl::ClosePath => self.0.iter().rev().skip(1).take_while(|el| !matches(|el|, PathEl::ClosePath).last().expect("can't start a segment on a ClosePath").end_point().unwrap()
```

Ok, I'm sure this can be clearer. But the point is that if we have a long `BezPath` containing multiple `ClosePath`s then `segments` will need to go through all of them but in fact we only need to look at the last closed path.

kurbo/src/bezpath.rs
```diff
@@ -393,6 +393,17 @@ impl BezPath {
         cbox.unwrap_or_default()
     }
 
+    /// Get the last point of the path, if path is not empty.
```
jneem, line 396: I think it's worth documenting here that the time complexity could be linear in the path length. I feel like this might be a little surprising otherwise, given that `Vec::last` is `O(1)`.

kurbo/src/bezpath.rs
jneem: I think the Segments iterator will panic in that case anyway.

sagudev:
I did all changes, improved docs and renamed to `current_position` as I think it's the most appropriate (in case you have more moves you have no points, but you have changing position of the pen).

kurbo/src/bezpath.rs
```diff
@@ -393,6 +393,27 @@ impl BezPath {
         cbox.unwrap_or_default()
     }
 
+    /// Returns current position in the path, if path is not empty.
+    ///
+    /// Unlike the `end_point` method, this also handles [`PathEl::ClosePath`],
+    /// by finding the first point of subpath, hence the time complexity is O(n).
```
jneem, line 399:
```suggestion
    /// This is different from calling [`PathEl::end_point`] on the last entry of [`BezPath::elements`]:
    /// this method handles [`PathEl::ClosePath`],
    /// by finding the first point of our last subpath, hence the time complexity is O(n).
```

jneem (approved):
Thanks, this looks fine to me now. Just a nitpick about the doc comment.

I think I actually liked the old name better, as "current" to me suggests state that is being modified over time. But this doesn't need to be blocked on naming.

sagudev:
If there will be no opposition I will land this as-is on Thursday.

PR #466, opened 2025-07-16
Use more `Into<Point>` instead of `Point` for arguments

sagudev:
We already did for most of `Point` arguments. This is especially relevant for users of euclid now that we have #463.

### DIFF
--- a/kurbo/src/affine.rs
+++ b/kurbo/src/affine.rs
@@ -69,8 +69,8 @@ impl Affine {
     ///
     /// See [`Affine::scale()`] for more info.
     #[inline]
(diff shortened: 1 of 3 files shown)


tomcur (commented):
I'm cautiously in favor of doing this for `Point`, but we may want to make sure we only do this for types whose `From` impls are free (within the confines of this crate). Otherwise the API would hide conversions that could potentially needlessly slow down user code, if the same conversions are performed multiple times.

`From<TranslateScale> for Affine` is not quite free, and, e.g., for `Rect` (not included here), `From<(Point, Size)>` and `From<euclid::Rect<f64, UnknownUnit>>` aren't free.

PR #469, opened 2025-07-18
Make `Vec2::splat` public.

xorgy:


waywardmonkeys (commented):
Since there is a new release being attempted ... if this lands right now, it should have a proper changelog entry.

xorgy:
Thanks!

PR #472, opened 2025-07-21
Inline `From` conversions to/from `euclid` types

tomcur:


sagudev:
Hm, I always just assumed that compiler would inline this automatically, but I think this makes sense.

Why are some function marked with only `#[inline]` and others with `#[inline(always)]`?

PR #473, opened 2025-07-24
Fixes aspect ratio calculation.

ErisianArchitect:
Aspect ratio was calculated with `height / width`, but that's incorrect. It should be `width / height`.

https://en.wikipedia.org/wiki/Aspect_ratio_(image)
https://en.wikipedia.org/wiki/Ratio#Notation_and_terminology

### DIFF
--- a/kurbo/src/size.rs
+++ b/kurbo/src/size.rs
@@ -256,7 +256,7 @@ impl Size {
 
     /// Returns the aspect ratio of a rectangle with the given size.
     ///
-    /// If the width is `0`, the output will be `sign(self.height) * infinity`. If The width and
+    /// If the height is `0`, the output will be `sign(self.width) * infinity`. If The width and
     /// height are `0`, then the output will be `NaN`.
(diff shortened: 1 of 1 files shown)


ErisianArchitect:
Ahh, right. I forgot to update the comments.

ErisianArchitect:
Has there been anything going on with this?

raphlinus:
Apologies for the slow response. We discussed this in office hours yesterday, and because of the possibility of breakage, I think it would be best to add new methods and deprecate the old ones. We *could* do it as a breaking change as we have the 0.12 milestone coming up, but if there are users of the method, it would be unfriendly.

I don't have a great idea for the new name, but "aspect_ratio_width" is what comes to mind.

raphlinus:
I'm not sure what you mean by "the internal API." It makes sense to switch as much stuff as possible to the new functions, as the old ones should be deprecated.

DJMcNab:
This has been superseded by #486. Thanks!

PR #475, opened 2025-07-30
Expose the `StrokeCtx`

LaurenzV:


kurbo/src/stroke.rs
```diff
@@ -169,9 +169,9 @@ impl StrokeOpts {
 /// Collection of values representing lengths in a dash pattern.
 pub type Dashes = SmallVec<[f64; 4]>;
 
-/// Internal structure used for creating strokes.
-#[derive(Default)]
-struct StrokeCtx {
+/// A structure that is used for creating strokes.
```
jneem, line 172:
```suggestion
/// A structure that is used for creating strokes.
///
/// See also [`stroke_with`].
```

(Just because there isn't anything to do with `StrokeCtx` otherwise, and since `stroke_with` is a free function it might not be obvious where to find it.)

jneem (approved):

PR #476, opened 2025-08-04
Add new `Axis` type lifted from Masonry.

PoignardAzur:
This type is helpful for widget code, but it would be even more helpful if we could pass it directly to Kurbo type methods.

This PR implements those methods.

### DIFF
--- a/kurbo/src/point.rs
+++ b/kurbo/src/point.rs
@@ -204,7 +204,7 @@ impl Point {
     }
 
     /// Get the member matching the given axis.
-    pub fn get(self, axis: Axis) -> f64 {
+    pub fn get_coord(self, axis: Axis) -> f64 {
         match axis {
             Axis::Horizontal => self.x,
             Axis::Vertical => self.y,
@@ -212,7 +212,7 @@ impl Point {
     }
 
     /// Get a mutable reference to the member matching the given axis.
-    pub fn get_mut(&mut self, axis: Axis) -> &mut f64 {
+    pub fn get_coord_mut(&mut self, axis: Axis) -> &mut f64 {
         match axis {
             Axis::Horizontal => &mut self.x,
             Axis::Vertical => &mut self.y,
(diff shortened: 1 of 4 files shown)


PoignardAzur:
@DJMcNab Thoughts?

PoignardAzur:
> We have similar functionality in Taffy (although we have several different Axis types there for different layout systems).

Right. If we ever do some sort of "browser on top of kurbo" project, I expect we'd have some conversion functions from `LogicalAxis` (inline/block) to `kurbo::Axis`, so we could still use this code.

DJMcNab:
My gut instinct is against this. It doesn't feel like it fits into Kurbo. Taking for example Kurbo's main purpose statement at https://docs.rs/kurbo/latest/kurbo/, I don't really feel that this is meaningfully a data structure for curves or vector paths.

I think we'll probably want to discuss this at office hours.

kurbo/src/point.rs
```diff
@@ -202,6 +202,30 @@ impl Point {
...
+        }
+    }
+
+    /// Set the member matching the given axis to the given value.
+    pub fn set(&mut self, axis: Axis, value: f64) {
+        match axis {
+            Axis::Horizontal => self.x = value,
+            Axis::Vertical => self.y = value,
+        }
+    }
```
DJMcNab, line 228: I don't much like the names of these methods; I think they're too unspecific/general for this functionality.

PoignardAzur: I could call them e.g. `get_axis`, `set_axis`, etc? I do want short names, because in some algorithms we end up using them a lot.

tomcur: In other parts of Kurbo these are called "coordinate," or "coordinate value," as in "x-coordinate" and "y-coordinate". Perhaps something like `get_axis_coordinate` would be clearest, but a lot longer than your suggestions.

PR #478, opened 2025-08-08
Fix `mismatched_lifetime_syntaxes`

tomcur:
This warns since Rust 1.89.


tomcur:
Oops, based this on the wrong remote's main! It has already been fixed.

PR #484, opened 2025-08-14
[cu2qu] Add some failing edge case tests (and a question)

cmyr:
Hi all,

In the course of investigating a difference in font compilation between rust and python I found some behaviour of the cubic to quadratic conversion code (in both implementations) that I think is counter intuitive, and which might be worth special-casing.

Specifically, if a cubic bezier has three identical points in a row (as in, the two control points + either the start or end) I would expect conversion to return a single quadratic bezier that just drops one of the input control points. (There could also be a special case if there are three equal points with unequal control points, but in that case there are at least several reasonable representations, so i'm ignoring it.)

That said, the current behaviour is not _wrong_, it's just somewhat unintuitive/sub-optimal.

I'm opening this as a PR but ultimately it's a discussion question; there's a patch open to change this in the python, but I'd only want to land that if we're also going to change it here.

see https://github.com/fonttools/fonttools/pull/3904

### DIFF
--- a/kurbo/src/cubicbez.rs
+++ b/kurbo/src/cubicbez.rs
@@ -1130,6 +1130,44 @@ mod tests {
     }
 
     #[test]
+    fn cubic_to_quadratic_pathological_front() {
+        let cubic = CubicBez::new(Point::ZERO, Point::ZERO, Point::ZERO, (10., 10.).into());
+        let quads = cubics_to_quadratic_splines(&[cubic], 0.1).unwrap();
+        assert_eq!(
+            quads,
+            [QuadSpline::new(vec![
+                Point::ZERO,
+                Point::ZERO,
+                (10., 10.).into()
+            ])]
+        );
+    }
+
+    #[test]
+    fn cubic_to_quadratic_pathological_back() {
+        let cubic = CubicBez::new((10., 10.).into(), Point::ZERO, Point::ZERO, Point::ZERO);
+        let quads = cubics_to_quadratic_splines(&[cubic], 0.1).unwrap();
+        assert_eq!(
+            quads,
+            [QuadSpline::new(vec![
+                (10., 10.).into(),
+                Point::ZERO,
+                Point::ZERO
+            ])]
+        );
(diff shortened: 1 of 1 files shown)


anthrotype:
are you sure we need to handle the case for three identical points? only the four indentical points is unequivocal. The other "pathological front and "pathological back" test cases, I don't think they need special handling. The algorithm will handle them by splitting the cubic and giving back a qcurve spline with two quad offcurves.

```python
>>> from fontTools.cu2qu import curve_to_quadratic
>>> curve_to_quadratic([(10, 10), (0, 0), (0, 0), (0, 0)], 0.1, all_quadratic=True)
[(10.0, 10.0), (6.25, 6.25), (2.1354166666666665, 2.1354166666666665), (0.3645833333333333, 0.3645833333333333), (0.0, 0.0), (0.0, 0.0)]
```

cmyr:
right, so it is 'correct' but it is suboptimal (if we are optimizing for spline size).

behdad:
I will think about to see if there's a math solution.

anthrotype:
I don't think there is. The only way a cubic bezier curve can be downgraded to a single quadratic Bezier curve is when its off-curve points are exactly at 2/3 of the distance between their respective on-curve point and the intersection point of the tangent lines at the start and end of the curve, which will be the quadratic off-curve point. That's how a quadratic curve that is elevated to cubic looks like. Those pathological back and front cases can't be reduced to a single quadratic.

behdad:
> Those pathological back and front cases can't be reduced to a single quadratic.

We can still come up with a guess and let the algorithm see if that's within tolerance or not.

behdad:
> > Those pathological back and front cases can't be reduced to a single quadratic.
> 
> We can still come up with a guess and let the algorithm see if that's within tolerance or not.

I think we should just return one of the off-curve points. Currently the algorithm doesn't even try a one-curve solution because div-by-zero happens. We can, in that case, try returning the simple solution and it will only be accepted if it's within tolerance.

PR #490, opened 2025-08-25
Fix miter join in dashed strokes

### LINKED ISSUE #482: [sparse_strips]: Incorrect dash rendering
**Project**: Vello Sparse Strips

### Description

I haven’t looked into this problem in depth yet, but there’s incorrect rendering when dash patterns are used with certain stroke widths and dash pattern values.

### Steps to Reproduce

1. Copy and paste the test below into the `vello_sparse_strips` test suite
2. Run the test individually
3. Observe the newly generated snapshot

```rust
#[vello_test(width = 120, height = 120)]
fn stroke_dash_pattern_issue(ctx: &mut impl Renderer) {
    let rect = Rect::new(0.0, 0.0, 80.0, 80.0);

    ctx.set_transform(Affine::translate((20.0, 20.0)));
    ctx.set_paint(REBECCA_PURPLE);
    ctx.set_stroke(Stroke {
        width: 20.0,
        join: Join::Miter,
        miter_limit: 4.0,
        start_cap: Cap::Butt,
        end_cap: Cap::Butt,
        dash_offset: 0.0,
        dash_pattern: SmallVec::from_vec(vec![73.0, 12.0]),
    });
    ctx.stroke_rect(&rect);
}
```

### Expected Behavior
- Dashes should render consistently and align correctly

### Actual Behavior
-  Parts of the path are not rendered as expected

Image

grebmeg:
Fixes #482

Before | After
---|---
<video src=https://github.com/user-attachments/assets/e21afea8-5e2c-47b6-9200-aa215d560373/>|<video src=https://github.com/user-attachments/assets/f9cd429b-94b1-4617-9817-23080a7d0bc7/>

Essentially, the problem was that when a miter point was detected, it was added only on one side, which caused incorrect behavior in some cases. Below is a quick diagram showing what was happening before and after the fix for path `M70,80 L0,80 L0,77`.

image     

I also verified the output against tiny-skia with identical rectangle parameters and stroke styles, and it matches now.

image

### DIFF
--- a/kurbo/src/stroke.rs
+++ b/kurbo/src/stroke.rs
@@ -442,12 +442,14 @@ impl StrokeCtx {
                                 let h = ab.cross(fp_this - fp_last) / cross;
                                 let miter_pt = fp_this - cd * h;
                                 self.forward_path.line_to(miter_pt);
+                                self.backward_path.line_to(p0);
                             } else if cross < 0.0 {
                                 let fp_last = p0 + last_norm;
                                 let fp_this = p0 + norm;
                                 let h = ab.cross(fp_this - fp_last) / cross;
                                 let miter_pt = fp_this - cd * h;
                                 self.backward_path.line_to(miter_pt);
+                                self.forward_path.line_to(p0);
                             }
                         }
                         self.forward_path.line_to(p0 - norm);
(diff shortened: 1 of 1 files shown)


platlas (approved):
I have checked also with different angles in corner.
state before, with rectangle and two triangles
state after, with rectangle and two triangles

Nice improvement.

PR #495, opened 2025-09-03
Mark C-style pub enums as `repr(u8)`

waywardmonkeys:


waywardmonkeys:
This is to help support future FFI work without having to require a breaking change.

This should be supplemented later with bytemuck support like we have in peniko and color.

nicoburns (approved):

PR #497, opened 2025-09-03
Add `Ellipse::{major,minor}_radius` methods

tomcur:
This adds methods to `Ellipse` to get the major and minor radii.

This documents the (private) `Affine::svd` method to guarantee the part of its behavior we're using here for efficiency.

The `Ellipse::radii()` method effectively already allows for this, by taking the `x` component for the major radius and the `y` component for the minor radius; however, we do not currently guarantee that behavior, and may not want to, to keep the option of changing internal representations. Plus, within crate boundaries and the specified inlining, the compiler should be able to eliminate the dead singular value decomposition code.

Quickly looking over the assembly, there are just the two `sqrtsd` instructions as expected.

Like https://github.com/linebender/kurbo/pull/496, this is motivated by https://github.com/linebender/vello/pull/1180, to allow eliding more computations.

<details>
<summary>x86 assembly</summary>

```assembly
		// kurbo/src/ellipse.rs:132
		pub fn major_radius(&self) -> f64 {
	.cfi_startproc
	sub rsp, 104
	.cfi_def_cfa_offset 112
		// kurbo/src/ellipse.rs:133
		self.inner.svd().0.x
	movupd xmm0, xmmword ptr [rdi]
	movupd xmm4, xmmword ptr [rdi + 16]
		// kurbo/src/affine.rs:401
		let a2 = a * a;
	movapd xmm3, xmm0
	mulpd xmm3, xmm0
		// kurbo/src/affine.rs:405
		let ab = a * b;
	movapd xmm2, xmm0
	unpcklpd xmm2, xmm4
	unpckhpd xmm0, xmm4
		// kurbo/src/affine.rs:403
		let c2 = c * c;
	mulpd xmm4, xmm4
		// kurbo/src/affine.rs:405
		let ab = a * b;
	mulpd xmm0, xmm2
		// kurbo/src/affine.rs:407
		let angle = 0.5 * (2.0 * (ab + cd)).atan2(a2 - b2 + c2 - d2);
	movapd xmm1, xmm0
	unpckhpd xmm1, xmm0
	addsd xmm1, xmm0
	movapd xmmword ptr [rsp + 80], xmm1
	movapd xmm0, xmm1
	addsd xmm0, xmm1
	movapd xmmword ptr [rsp + 48], xmm3
		// kurbo/src/affine.rs:408
		let s1 = a2 + b2 + c2 + d2;
	movapd xmm2, xmm3
	unpckhpd xmm2, xmm3
[...]

PR #500, opened 2025-09-04
Release 0.12.0

tomcur:


tomcur (commented):
~~Requires landing https://github.com/linebender/kurbo/pull/495.~~

Edit: it doesn't, we decided to skip it and release in 0.12.1.

It would be nice to have https://github.com/linebender/kurbo/pull/499 (to speed up the methods introduced for Vello sparse strips in https://github.com/linebender/kurbo/pull/497), but that could be done in a non-breaking follow-up.

nicoburns (approved):
LGTM, but someone with release permissions should merge.

PR #501, opened 2025-09-04
Fixup changelog for 0.12.0 release

tomcur:
Line about bumping MSRV was in the wrong place in kurbo#500. (While we're at it, add a missing period.)


tomcur (commented):
Will merge and publish in a bit.

PR #505, opened 2025-09-14
Change `Line::nearest` implementation to be branchless

tomcur:
This is a branchless calculation of a point-onto-line-segment projection. Also marks the method `#[inline]` and adds some testing.

I started looking at this because of the math we introduced in https://github.com/linebender/vello/pull/1214. This might allow us to introduce some tighter bounds there without introducing additional branching.

This should be useful here regardless of whether the above happens.

### DIFF
--- a/kurbo/src/line.rs
+++ b/kurbo/src/line.rs
@@ -166,10 +166,10 @@ impl ParamCurveNearest for Line {
         // Calculate projection parameter `t` of the point onto s(t), with s(t) the line segment
         // such that s(t) = (1-t) * p0 + t * p1.
         //
-        // Note this will be inf when the segment has 0 length; see the clamping below.
+        // Note this will be infinite or nan when the segment has 0 length; see the clamping below.
         let t = d.dot(v) / d.hypot2();
 
-        // Clamp the parameter to be on the line segment. This results in `t==0` if `t==inf` above.
+        // Clamp the parameter to be on the line segment. This clamps `-inf` and `nan` to `0`, and `inf` to `1`.
         #[expect(
             clippy::manual_clamp,
             reason = "`clamp` uses slightly more instructions than chained `max` and `min` on x86 and aarch64"
--- a/kurbo/src/line.rs
+++ b/kurbo/src/line.rs
@@ -170,7 +170,11 @@ impl ParamCurveNearest for Line {
         let t = d.dot(v) / d.hypot2();
 
         // Clamp the parameter to be on the line segment. This results in `t==0` if `t==inf` above.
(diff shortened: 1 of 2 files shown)


kurbo/src/line.rs
```diff
@@ -158,19 +158,27 @@ impl ParamCurveArea for Line {
...
+        // Calculate projection parameter `t` of the point onto s(t), with s(t) the line segment
+        // such that s(t) = (1-t) * p0 + t * p1.
+        //
+        // Note this will be inf when the segment has 0 length; see the clamping below.
+        let t = d.dot(v) / d.hypot2();
+
+        // Clamp the parameter to be on the line segment. This results in `t==0` if `t==inf` above.
+        #[expect(
+            clippy::manual_clamp,
+            reason = "`clamp` uses slightly more instructions than chained `max` and `min` on x86 and aarch64"
```
tomcur, line 175: aarch64: https://godbolt.org/z/Mv4Y19jrM
x86: https://godbolt.org/z/WW19hMd6v

Perhaps we should disable this lint in `kurbo`.

tomcur: Seems that way! (As a separate note, we actually use some of the `NaN` semantics of the chained-min-max in the analytical pixel coverage calculation in the sparse strips work, so this lint can be a bit noisy.)

kurbo/src/line.rs
```diff
@@ -158,19 +158,27 @@ impl ParamCurveArea for Line {
...
-            let t = dotp / d_squared;
-            let dist = (p - self.eval(t)).hypot2();
-            (t, dist)
-        };
+        let v = p - self.p0;
+
+        // Calculate projection parameter `t` of the point onto s(t), with s(t) the line segment
+        // such that s(t) = (1-t) * p0 + t * p1.
+        //
+        // Note this will be inf when the segment has 0 length; see the clamping below.
```
jneem, line 169:
```suggestion
        // Note this will be infinite or nan when the segment has 0 length; see the clamping below.
```

kurbo/src/line.rs
```diff
@@ -158,19 +158,27 @@ impl ParamCurveArea for Line {
...
+        let t = d.dot(v) / d.hypot2();
+
+        // Clamp the parameter to be on the line segment. This results in `t==0` if `t==inf` above.
```
jneem, line 172:
```suggestion
        // Clamp the parameter to be on the line segment. This clamps `-inf` and `nan` to `0`, and `inf` to `1`.
```

jneem (approved):

PR #515, opened 2025-09-28
Add workflow_dispatch trigger to CI workflow

sagudev:


xStrom:
As discussed in [vello#1250](https://github.com/linebender/vello/pull/1250), we can already re-run failed jobs with the current configuration, so this isn't needed to solve that problem.

PR #521, opened 2025-11-04
Mark a bunch of `is_finite` and `is_nan` methods `const`

tomcur:
Stabilized in Rust 1.83 (we now have MSRV 1.85).


cmyr (approved):
nice to keep on top of these, thanks!

PR #527, opened 2025-11-16
Derive Hash for Axis

jrmoulton:


waywardmonkeys (approved):
Curious where you need this. Assume as part of some larger type that is `Hash` rather than using this itself as a key.

jrmoulton:
Correct. I have a a module to encode pointer events into an affine and I do some lookup based on axis/pointer type

tomcur (approved):

PR #530, opened 2025-11-24
Release v0.13.0

tomcur:


tomcur:
We release this as breaking due to the MSRV bump, though we state in the README we don't consider MSRV bumps breaking. During office hours it was agreed that due to the amount of dependents on Kurbo we probably should consider the bump breaking.

Perhaps we should remove (or negate) that statement in the README.

waywardmonkeys (approved):
You'll have to update the date again...

PR #532, opened 2025-12-15
fix: incorrect segment order

### LINKED ISSUE #528: stroke() function with dashed stroke returns first piece last
Graphite wraps Kurbo's [`stroke()`](https://docs.rs/kurbo/latest/kurbo/fn.stroke.html) with our Solidify Stroke node. I am noticing that, when a dashed line (regardless of cap style) is involved, it returns the first piece in the wrong order, at the very end. See this demonstration where we move an object along the edges of each subpath in order, starting from the second piece, going to the last piece, then returning to end with the first piece:

https://github.com/user-attachments/assets/0ded642e-957d-457f-a2d6-3182fdb63e07

I had `stroke()` print out its SVG directly, and I split up each `M` command into its own `<path />` so I can delete each one sequentially to show the ordering issue, directly from Kurbo (not a Graphite node bug):

https://github.com/user-attachments/assets/f4104d00-8105-48a2-945a-1740dbc912db

We can also see that this causes problems assigning colors along a gradient to each separate subpath:

Image

If we add in a more subpaths as part of the input compound path, we see it affects every individual subpath:

Image

ankushKun:
Added a new `OutputStashThenWork` state to the `DashIterator` in stroke.rs to handle correct output order of stashed path segments. For open paths, the stash is output first and then it continues like usual, for closed paths the logic remains same.

edit: fixes #528

### DIFF
--- a/kurbo/src/stroke.rs
+++ b/kurbo/src/stroke.rs
@@ -602,6 +602,8 @@ enum DashState {
     ToStash,
     Working,
     FromStash,
+    /// for open paths: output stash contents then work
+    OutputStashThenWork,
 }
 
 impl<T: Iterator<Item = PathEl>> Iterator for DashIterator<'_, T> {
@@ -646,6 +648,17 @@ impl<T: Iterator<Item = PathEl>> Iterator for DashIterator<'_, T> {
                         } else {
                             self.state = DashState::ToStash;
                         }
+                    }
+                }
+                DashState::OutputStashThenWork => {
+                    // output stashed content then continue working
+                    if let Some(el) = self.stash.get(self.stash_ix) {
+                        self.stash_ix += 1;
+                        return Some(*el);
+                    } else {
+                        self.stash.clear();
+                        self.stash_ix = 0;
+                        self.state = DashState::Working;
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


jneem:
Sorry it's taken so long to respond. I don't think a single-segment approach can solve this: this PR makes the decision about whether to output the first dash segment when it reaches the end of the first path segment. But at that point we don't yet know whether the first path segment is part of a closed path or not. For example, consider this test case:

```
    #[test]
    fn dash_sequence() {
        let shape = crate::Rect::from_points((0.0, 0.0), (4.0, 4.0));
        let dashes = [5., 1.];
        let expansion = [
            PathEl::MoveTo((4.0, 2.0).into()),
            PathEl::LineTo((4.0, 4.0).into()),
            PathEl::LineTo((1.0, 4.0).into()),
            PathEl::MoveTo((0.0, 4.0).into()),
            PathEl::LineTo((0.0, 0.0).into()),
            PathEl::LineTo((4.0, 0.0).into()),
            PathEl::LineTo((4.0, 1.0).into()),
        ];
        let iter = dash(shape.path_elements(0.), 0., &dashes);
        assert_eq!(iter.collect::<Vec<PathEl>>(), expansion);
    }
```

The current PR will break this test, because it will issue the path (0, 0) -- (4, 0) -- (4, 1) first instead of joining it with the final dash.

I wonder if a possible solution is to put the burden on the caller: we allow the caller to specify that they want the dashes reported eagerly, without a stash.

ankushKun:
@RobertBrewitz Give it a go! I havent had the chance to look at this further :)

RobertBrewitz:
Superseded by #550

PR #534, opened 2025-12-23
Micro-optimize `RoundedRect::winding`

tomcur:
In some Understory code we're using quite some `RoundedRect::contains` checking currently, which calls out to `RoundedRect::winding`, so that prompted some looking into its performance.

Doing three explicit `if`-statements to find the point's quadrant is slightly faster than the explicit `match`, and, stupidly, doing those `if`s through bit operations is quite a bit faster than that still, the result being ~50% faster than `main`. Perhaps the bit operations are faster only because the compiler makes better choices about data fetching. (Even though the explicit `if`s and these bit operations do otherwise compile to more or less the same branchless instructions!)

That last result may not be consistent going into the future, but in any case the code proposed here shouldn't become slower than the `match`. Using an early `std::hint::prefetch_read_instruction(&radii as *const _, std::hint::Locality::L1)` gets to roughly the same timings regardless of `select` implementation, but that's not available in stable Rust.

Timings (main, this PR, and the equivalent `if`-statements respectively):

```
rounded_rect_contains/winding-main
                        time:   [40.983 µs 41.110 µs 41.234 µs]
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe

rounded_rect_contains/winding-bit-magic
                        time:   [19.162 µs 19.595 µs 20.122 µs]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

rounded_rect_contains/winding-ifs
                      time:   [37.842 µs 38.065 µs 38.317 µs]
```

(Benchmark is running a winding check on `10,000` PRNG'd points that are contained by `rrect.rect()` but may be outside the corners of the `rrect` itself, thanks @waywardmonkeys. Same rough timings for when the points are very unlikely to be contained by the rounded rect, or are always in the rounded parts.)

Godbolt, showing that `if`-statements and bit operations are somewhat better than a `match` (the last compile down to be identical on aarch64, though the relative timings of these two as in the above benchmark hold on aarch64): https://godbolt.org/z/xhPxhnndG. This doesn't show Rust making better memory fetching decisions.

Assembly for the full method inside the `<details>` below.

<details>
  <summary>`RoundedRect::winding` x86 assembly</summary>

  The `RoundedRect::winding` assembly on x86 with bit ops:
  
  ```asm
          // kurbo/src/rounded_rect.rs:113
          self.radii
  lea rax, [rdi + 32]
  lea rcx, [rdi + 40]
  lea rdx, [rdi + 48]
  lea rsi, [rdi + 56]
          // kurbo/src/rect.rs:174
          Point::new(0.5 * (self.x0 + self.x1), 0.5 * (self.y0 + self.y1))
  movupd xmm3, xmmword ptr [rdi]
  movupd xmm2, xmmword ptr [rdi + 16]
  movapd xmm4, xmm3
  addpd xmm4, xmm2
  movapd xmm5, xmmword ptr [rip + .LCPI128_0]
  mulpd xmm4, xmm5
          // kurbo/src/rounded_rect.rs:299
          pt.x -= center.x;
  unpcklpd xmm0, xmm1
  subpd xmm0, xmm4
  xorpd xmm1, xmm1
          // kurbo/src/rounded_rect.rs:324
          let r_top = select_f64(radii.top_left, radii.top_right, pt.x >= 0.);
  ucomisd xmm0, xmm1
          // kurbo/src/rounded_rect.rs:313
          f64::from_bits((a.to_bits() & !mask) | (b.to_bits() & mask))
  cmovae rax, rcx
  cmovae rsi, rdx
          // kurbo/src/rounded_rect.rs:326
          let radius = select_f64(r_top, r_bot, pt.y >= 0.).max(0.0);
  movapd xmm4, xmm0
  unpckhpd xmm4, xmm0
  ucomisd xmm4, xmm1
          // kurbo/src/rounded_rect.rs:313
          f64::from_bits((a.to_bits() & !mask) | (b.to_bits() & mask))
  cmovb rsi, rax
  movsd xmm4, qword ptr [rsi]
  maxsd xmm4, xmm1
          // kurbo/src/rect.rs:108
          self.x1 - self.x0
  subpd xmm2, xmm3
          // kurbo/src/rounded_rect.rs:331
          let inside_half_width = (self.width() / 2.0 - radius).max(0.0);
  mulpd xmm2, xmm5
  movapd xmm1, xmm4
  unpcklpd xmm1, xmm4
  subpd xmm2, xmm1
  xorpd xmm1, xmm1
  maxpd xmm2, xmm1
  andpd xmm0, xmmword ptr [rip + .LCPI128_1]
          // kurbo/src/rounded_rect.rs:344
          let px = (pt.x.abs() - inside_half_width).max(0.0);
  subpd xmm0, xmm2
  maxpd xmm0, xmm1
          // kurbo/src/rounded_rect.rs:351
          let inside = px * px + py * py <= radius * radius;
  mulpd xmm0, xmm0
  movapd xmm1, xmm0
  unpckhpd xmm1, xmm0
  addsd xmm1, xmm0
  mulsd xmm4, xmm4
  xor eax, eax
  ucomisd xmm4, xmm1
  setae al
          // kurbo/src/rounded_rect.rs:357
          }
  ret
  ```
  
  And with `if` statements instead:
  
  ```asm
          // kurbo/src/rounded_rect.rs:113
          self.radii
  lea rax, [rdi + 48]
  lea rcx, [rdi + 56]
          // kurbo/src/rect.rs:174
          Point::new(0.5 * (self.x0 + self.x1), 0.5 * (self.y0 + self.y1))
  movupd xmm3, xmmword ptr [rdi]
  movupd xmm2, xmmword ptr [rdi + 16]
  movapd xmm4, xmm3
  addpd xmm4, xmm2
  movapd xmm5, xmmword ptr [rip + .LCPI128_0]
  mulpd xmm4, xmm5
          // kurbo/src/rounded_rect.rs:299
          pt.x -= center.x;
  unpcklpd xmm0, xmm1
  subpd xmm0, xmm4
  xorpd xmm1, xmm1
          // kurbo/src/rounded_rect.rs:324
          let r_top = select_f64(radii.top_left, radii.top_right, pt.x >= 0.);
  xor edx, edx
  ucomisd xmm0, xmm1
  setae dl
          // kurbo/src/rounded_rect.rs:314
          if !cond { a } else { b }
  lea rdx, [rdi + 8*rdx + 32]
  cmovae rcx, rax
          // kurbo/src/rounded_rect.rs:326
          let radius = select_f64(r_top, r_bot, pt.y >= 0.).max(0.0);
  movapd xmm4, xmm0
  unpckhpd xmm4, xmm0
  ucomisd xmm4, xmm1
          // kurbo/src/rounded_rect.rs:314
          if !cond { a } else { b }
  cmovb rcx, rdx
  movsd xmm4, qword ptr [rcx]
  maxsd xmm4, xmm1
          // kurbo/src/rect.rs:108
          self.x1 - self.x0
  subpd xmm2, xmm3
          // kurbo/src/rounded_rect.rs:331
          let inside_half_width = (self.width() / 2.0 - radius).max(0.0);
  mulpd xmm2, xmm5
  movapd xmm1, xmm4
  unpcklpd xmm1, xmm4
  subpd xmm2, xmm1
  xorpd xmm1, xmm1
  maxpd xmm2, xmm1
  andpd xmm0, xmmword ptr [rip + .LCPI128_1]
          // kurbo/src/rounded_rect.rs:344
          let px = (pt.x.abs() - inside_half_width).max(0.0);
  subpd xmm0, xmm2
  maxpd xmm0, xmm1
          // kurbo/src/rounded_rect.rs:351
          let inside = px * px + py * py <= radius * radius;
  mulpd xmm0, xmm0
  movapd xmm1, xmm0
  unpckhpd xmm1, xmm0
  addsd xmm1, xmm0
  mulsd xmm4, xmm4
  xor eax, eax
  ucomisd xmm4, xmm1
  setae al
          // kurbo/src/rounded_rect.rs:357
          }
  ret
  ```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ You can find its changes [documented below](#0130-2025-11-27).
 
 This release has an [MSRV][] of 1.85.
 
+## Changed
+
+- Improve performance of `RoundedRect::winding` and `RoundedRect::contains`. ([#534][] by [@tomcur][])
+
 ## [0.13.0] (2025-11-27)
 
 This release has an [MSRV][] of 1.85.
@@ -263,6 +267,7 @@ Note: A changelog was not kept for or before this release
 [#525]: https://github.com/linebender/kurbo/pull/525
 [#526]: https://github.com/linebender/kurbo/pull/526
 [#527]: https://github.com/linebender/kurbo/pull/527
+[#534]: https://github.com/linebender/kurbo/pull/534
 
 [Unreleased]: https://github.com/linebender/kurbo/compare/v0.13.0...HEAD
 [0.13.0]: https://github.com/linebender/kurbo/releases/tag/v0.13.0


kurbo/src/rounded_rect.rs
```diff
@@ -301,13 +301,26 @@ impl Shape for RoundedRect {
...
+            #[inline(always)]
+            fn select(a: f64, b: f64, cond: bool) -> f64 {
+                let mask = (cond as u64).wrapping_neg(); // 0 or !0
+                f64::from_bits((a.to_bits() & !mask) | (b.to_bits() & mask))
+            }
+
+            let radii = self.radii();
+            let radius_top = select(radii.top_left, radii.top_right, pt.x >= 0.);
+            let radius_bottom = select(radii.bottom_left, radii.bottom_right, pt.x >= 0.);
+            select(radius_top, radius_bottom, pt.y >= 0.).max(0.0)
```
jneem: Why the `.max(0.0)`? The original seems to allow a negative radius, although I don't really understand why...

tomcur: Ah, indeed, apparently the original allowed for that... I've just removed the `max` so the behavior remains the same.

PR #537, opened 2026-01-13
Fix winding number solver near endpoints

### LINKED ISSUE #531: Numerical instability in winding number / contains.
Hi there.
I've stumbled upon a very specific case where the `contains` check returns a wrong result if the test point sits right between the endpoint of two Bézier curves:
```rust
let bez = BezPath::from_vec(vec![
    PathEl::MoveTo((200.0, 410.0).into()), 
    PathEl::CurveTo((139.12278747558594, 410.0).into(), (90.0, 360.8772277832031).into(), (90.0, 300.0).into()), 
    PathEl::CurveTo((90.0, 239.12278747558594).into(), (139.12278747558594, 190.0).into(), (200.0, 190.0).into()),
    PathEl::CurveTo((150.19137573242188, 210.0).into(), (110.0, 250.19137573242188).into(), (110.0, 300.0).into()), 
    PathEl::CurveTo((110.0, 349.8086242675781).into(), (150.19137573242188, 390.0).into(), (200.0, 390.0).into()),
    PathEl::ClosePath]);

assert!(bez.contains((100.0,300.1).into())); // Works fine.
assert!(bez.contains((100.0,299.9).into())); // Works fine.
assert!(bez.contains((100.0,300.0).into())); // Fails.
```
I have tried simplifying more, but even if I remove the digits after the decimal points, the test produces the correct results instead of reproducing the issue.
Tried this in both debug and release.
I have no idea if there is an established way to numerically stablise this check. Kurbo can't be the first program to run into this issue.
From what I read, issue #180 is not directly related to this. But that's for the crate experts to decide. :)
All the best,
Simon

Bonus SVG code to quickly see that the point should indeed be contained:
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="800" height="800" viewBox="0 0 500 500" fill="none">
<path d="M200,410 C139.12278747558594,410 90,360.8772277832031 90,300 C90,239.12278747558594 139.12278747558594,190 200,190 C150.19137573242188,210 110,250.19137573242188 110,300 C110,349.8086242675781 150.19137573242188,390 200,390 Z" fill="red"/>
   <rect x="100" y="300" width="3" height="3" fill="#000"/>
</svg>
```

jneem:
In the winding number calculation, it when the ray's crossing was close to an endpoint it was possible for the solver to mistakenly put the solution outside of `[0.0, 1.0]`. This fixes the solver so that it always finds a solution (which it must, because the segment is continuous and we check that its y values span the y we're looking for).

Fixes #531.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ This release has an [MSRV][] of 1.85.
 - Improve performance of `RoundedRect::winding` and `RoundedRect::contains`. ([#534][] by [@tomcur][])
 - `Axis`, `Insets`, `Line::reversed`, `Rect::min_x/max_x/min_y/max_y`, `RoundedRectRadii::abs/clamp/as_single_radius`, `RoundedRect::width/height/radii/rect/origin/center`, `Size::aspect_ratio`, `Stroke::new/with_join/with_miter_limit/with_start_cap/with_end_cap/with_caps` are all now `const`. ([#536][] by [@xStrom], [#539][] by [@tomcur][])
 
+## Fixed
+
+- Improved numerical robustness of winding numbers for paths. ([#537][] by [@jneem][])
+
 ## [0.13.0] (2025-11-27)
 
 This release has an [MSRV][] of 1.85.
@@ -271,6 +275,7 @@ Note: A changelog was not kept for or before this release
 [#527]: https://github.com/linebender/kurbo/pull/527
 [#534]: https://github.com/linebender/kurbo/pull/534
 [#536]: https://github.com/linebender/kurbo/pull/536
+[#537]: https://github.com/linebender/kurbo/pull/537
 [#539]: https://github.com/linebender/kurbo/pull/539
 
 [Unreleased]: https://github.com/linebender/kurbo/compare/v0.13.0...HEAD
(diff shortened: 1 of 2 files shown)


tomcur (approved):
This looks good, thanks! A changelog entry mentioning improved robustness would be great.

I'm idly wondering whether there are still adversarial inputs possible at the level of paths (not segments), e.g. where two near-identical and mirrored segments fail slightly differently such that an `x` is falsely considered inside or outside... for any reasonable horizontal distance to the path edges perhaps not.

jneem:
As you already alluded to, if the point's x coordinate is very close to the segments then you can definitely get inconsistent answers from adjacent segments, but I think that's ok: the point is close to the boundary anyway and so errors are allowed.

As long as there's a reasonable horizontal distance, I think we're safe against inconsistencies -- I added a little comment to `winding_inner` giving a hint as to why. But for example, if you have a decreasing-y segment followed by an increasing-y segment and the point's y coordinate is exactly at the vertex or slightly above then (with this PR) we are guaranteed to say that both segments cross the point's y coordinate. If there's a reasonable horizontal distance involved, we'll get the x coordinates correct enough and either get a winding number of 1 - 1 or 0 + 0. On the other hand, if the point's y coordinate is strictly below the vertex then we'll way that neither segment crosses the point's y coordinate.

The other case (increasing-y followed by decreasing-y) handles the exact equality case differently, but it should still be ok :crossed_fingers:

PR #547, opened 2026-02-20
Add `scaling`, `with_scaling`, `skewing`, `with_skewing` to `Affine`

laurenz-canva:
`scale` (and `skew`) would be better IMO, but that name's already taken. :(

### DIFF
--- a/kurbo/src/affine.rs
+++ b/kurbo/src/affine.rs
@@ -459,7 +459,7 @@ impl Affine {
         }
     }
 
-    /// Replaces the translation portion of this affine map
+    /// Replaces the translation portion of this affine map.
     ///
     /// The translation can be seen as being applied after the linear part of the map.
     #[must_use]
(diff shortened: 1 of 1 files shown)


nicoburns:
I'd be inclined to go with `get_scale` and `get_skew` for the getters. That's not considered entirely idiomatic in Rust, but we've violated the idiom by using those names for something else.

jneem:
Could you point me to the vello PR? Just so I can understand the context.

My (weakly-held) opinion is that it makes sense to think of a general affine map as having a linear part and a translation part, but it doesn't make sense to think of it as having a skew part or a scaling part (or a rotation part, for that matter), because those parts get all mixed up with one another.

laurenz-canva:
I can't find the exact location anymore unfortunately 😦 I just remember it was in this PR https://github.com/linebender/vello/pull/1456, but I'm not able to find it anymore.

Anyway, I think I agree now that this is probably not the best API, so I will close it for now. Perhaps we can add an API to change scaling + skewing all at once, but it's not urgent for me. Thanks for all the comments! 😄

PR #552, opened 2026-03-06
SVG paths: parsing NaN and Infinity

nicoburns:
Allows paths like `M0,0 L0,0 L90,0 LNaN,NaN` to be parsed. This is useful for deserializing paths encoded in AnyRender serialized scenes. Otherwise deserializing panics.

(one might argue that we ought to strip NaN's and Infinity's from scene's, but we may also want to specifically test for them, so I think it is best to at least have the capability to preserve them)

### DIFF
--- a/kurbo/src/svg.rs
+++ b/kurbo/src/svg.rs
@@ -324,6 +324,17 @@ impl SvgLexer<'_> {
         self.ix -= 1;
     }
 
+    fn get_str(&mut self, s: &[u8]) -> Result<(), SvgParseError> {
+        for &expected_c in s.into_iter() {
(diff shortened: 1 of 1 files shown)


raphlinus:
This doesn't feel right to me, I would consider non-finite values to be invalid. I think ideally deserialization should fail but not panic.

I confirm @jneem's observation, the SVG 2 "spec" is obviously defective. However, SVG 1.1 has a clear [grammar](https://www.w3.org/TR/SVG11/paths.html#PathDataBNF) in which these would be rejected.

nicoburns:
I've vendored a copy of kurbo's path parsing code into `anyrender_serialize`. So I guess if this is a "no" then we should just close it.

PR #563, opened 2026-03-27
Add `Affine::then_skew` and `Affine::then_reflect`

waywardmonkeys:


tomcur (approved):
Looks good. We should also add the `pre_{skew,reflect}` variants.

I think the tests can be dropped, as they mirror the implementation exactly.

tomcur (approved):

PR #565, opened 2026-03-29
Align path validity with SVG: only the first subpath requires a `MoveTo`

tomcur:
Match SVG semantics:

> A path data segment (if there is one) must begin with a "moveto" command.

And:

> If a "closepath" is followed immediately by a "moveto", then the "moveto" identifies the start point of the next subpath. If a "closepath" is followed immediately by any other command, then the next subpath starts at the same initial point as the current subpath.

### DIFF
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -520,10 +520,8 @@ impl BezPath {
 /// A subpath of a [`BezPath`].
 ///
 /// Returned by [`BezPath::subpaths`]. See
-/// [`BezPath`'s documentation on elements and segments][elements-and-segment]
+/// [`BezPath`'s documentation on elements and segments][BezPath#elements-and-segments]
 /// for more information.
-///
-/// [elements-and-segments]: BezPath#elements-and-segments
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Subpath<'a> {
     start: Point,
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -48,7 +48,7 @@ use crate::common::FloatFuncs;
 /// more `LineTo`, `QuadTo`, and `CurveTo` elements. A subpath can optionally
 /// be closed with `ClosePath`. New subpaths start after a `ClosePath` or
 /// `MoveTo`. Subpaths without an initial `MoveTo` start at the previous
-/// subpath's initial point.
+/// subpath's end point.
 ///
 /// Internally, a `BezPath` is a list of [`PathEl`]s; as such it implements
 /// [`FromIterator<PathEl>`] and [`Extend<PathEl>`]:
@@ -135,9 +135,11 @@ pub enum PathEl {
     ///
     /// The first two points are the Bézier's control points, the last point is its end point.
     CurveTo(Point, Point, Point),
-    /// Close off the path.
+    /// Close off the path by drawing a line from the current location to the current subpath's
+    /// start point.
     ///
-    /// Any element after this starts a new subpath.
+    /// Any element after this starts a new subpath from the closing line's end point (or,
+    /// equivalently, from this subpath's start point).
     ClosePath,
 }
 
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -43,9 +43,12 @@ use crate::common::FloatFuncs;
 /// For tasks like drawing, elements are a natural fit, but when doing
 /// hit-testing or subdividing, we need to have access to the segments.
 ///
-/// Conceptually, a `BezPath` contains zero or more subpaths. Each subpath
-/// *always* begins with a `MoveTo`, then has zero or more `LineTo`, `QuadTo`,
-/// and `CurveTo` elements, and optionally ends with a `ClosePath`.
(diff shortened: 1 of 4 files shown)


dominikh:
This will likely lead to additional follow-up changes to add support for the implied MoveTo after ClosePath. I've noticed the following so far:

- https://github.com/linebender/kurbo/blob/d3398c57ab06ba69fa56ddffbf2242a62d62f6ac/kurbo/src/simplify.rs#L335-L340
- https://github.com/linebender/kurbo/blob/d3398c57ab06ba69fa56ddffbf2242a62d62f6ac/kurbo/src/bezpath.rs#L330-L336
- https://github.com/linebender/kurbo/blob/d3398c57ab06ba69fa56ddffbf2242a62d62f6ac/kurbo/src/bezpath.rs#L701-L703

It looks like we weren't very consistent in supporting this in the past.

kurbo/src/bezpath.rs
```diff
@@ -483,6 +517,47 @@ impl BezPath {
...
+/// A subpath of a [`BezPath`].
+///
+/// Returned by [`BezPath::subpaths`]. See
+/// [`BezPath`'s documentation on elements and segments][BezPath#elements-and-segments]
+/// for more information.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Subpath<'a> {
+    start: Point,
+    elements: &'a [PathEl],
+}
```
tomcur, line 529: Subpaths started after a `Close` without a `MoveTo` need to know the where the previous subpath ends (or equivalently, where that subpath started). This perhaps would be the main argument in favor of *not* aligning with SVG, as always requiring an explicit `MoveTo` would allow taking `&[PathEl]` slices from the `BezPath` and  having them be valid paths without further context.

Note that if we want such semantics, we may still want to update code (and improve docs) as Kurbo currently is not consistent and does different things in different places.

`BezPath::segments` does the SVG-aligned thing:
https://github.com/linebender/kurbo/blob/0c31f878f9e73f9aba94f10c93d1552fafc27ea7/kurbo/src/bezpath.rs#L853-L859

`BezPath::get_seg` returns `None` (even though `BezPath::segments` does return the segment):
https://github.com/linebender/kurbo/blob/0c31f878f9e73f9aba94f10c93d1552fafc27ea7/kurbo/src/bezpath.rs#L330-L336

`BezPath::moments` also does the SVG-aligned thing:
https://github.com/linebender/kurbo/blob/0c31f878f9e73f9aba94f10c93d1552fafc27ea7/kurbo/src/moments.rs#L554-L559

`BezPath::current_position` requires all subpaths to start with `MoveTo` (otherwise it treats the first element as if it were a `MoveTo`:
https://github.com/linebender/kurbo/blob/0c31f878f9e73f9aba94f10c93d1552fafc27ea7/kurbo/src/bezpath.rs#L402-L422

`Shape for &'a [PathEl]` similarly treats the first drawing command in case of a missing `MoveTo`s as a `MoveTo`:
https://github.com/linebender/kurbo/blob/0c31f878f9e73f9aba94f10c93d1552fafc27ea7/kurbo/src/bezpath.rs#L1444-L1448

tomcur:
I believe things are now consistent (or at least mostly consistent). There actually is an argument for keeping the requirement that subpaths start with `MoveTo`, as then each `&[PathEl]` subpath slice is itself a valid path, but that would still require changes: see https://github.com/linebender/kurbo/pull/565#discussion_r3123292739.

I can see two possible behaviors in case of missing `MoveTo`s that requires no further context of previous subpaths:
- treat the first drawing command as if it were a `MoveTo` (i.e., just using its end point); or
- implicitly start from `(0, 0)`.

E.g., Skia does the latter https://api.skia.org/classSkPath.html#a27c6505cd16db954d6f0268e2de5d87f. On current `main` in some places we do the former or we already do the SVG-aligned thing of using the previous subpath's end point (see https://github.com/linebender/kurbo/pull/565#discussion_r3123292739).

dominikh:
I think that being able to refer to subpaths without additional context is more valuable than the flexibility of omitting some MoveTos. Plus, paths are quite a bit easier to use if they don't require as much state tracking by the consumer. 
SVG's model isn't necessarily an example of good design in this regard, anyway, considering it tries to simultaneously optimize the text format for writing documents by hand and making paths as succinct and unreadable as possible to save some bytes.

> treat the first drawing command as if it were a MoveTo (i.e., just using its end point)

IMO that'll lead to annoying to debug issues in user code.

In addition to slicing paths, we should also consider the semantics of concatenating paths. But I am not sure that _any_ model that permits omitting MoveTo makes intuitive sense in that context. However, starting at the previous subpath's _start point_ seems to rank rather low on expectedness, IMHO.

tomcur:
> I think that being able to refer to subpaths without additional context is more valuable than the flexibility of omitting some MoveTos.

In a vacuum I completely agree with this statement.

> SVG's model isn't necessarily an example of good design in this regard

While I agree, I think following SVG's model should be the "default" unless arguments against are strong enough (which they may be in this case).

> But I am not sure that any model that permits omitting MoveTo makes intuitive sense in that context. However, starting at the previous subpath's start point seems to rank rather low on expectedness, IMHO.

Another way to think about SVG's model is that you always continue from the *current* point: a `Close` is a line from the current point to the subpath's start, that start point becoming the new current point, and the next draw command will continue from that start point.

PR #567, opened 2026-04-07
Bugfix: Incorrect ordering for `Affine::pre_rotate_about`

### LINKED ISSUE #566: Incorrect ordering for `Affine::pre_rotate_about`
For `Affine::pre_rotate_about`, the documented transformation order is:

- rotate-about -> self

But the actual transformation order is:

- self -> rotate-about

This is likely just a typo. The doc comment says "Equivalent to `self * Affine::rotate_about(th, center)`", but the implementation is written with the opposite order.

dannymcgee:
Fixes #566


tomcur (approved):
Good catch, thank you!

PR #569, opened 2026-04-15
Add nuclear/Frobenius/spectral norms to `Affine`, add interpretation for `determinant`

tomcur:
These are coming up a few times in the sparse strips Vellos. E.g., https://github.com/linebender/vello/blob/8afc1c6b1be8b4dff6bc93f95a4113e2db69d7cd/sparse_strips/vello_common/src/filter/gaussian_blur.rs#L27-L28 is the nuclear norm. https://github.com/linebender/vello/pull/1180 introduces the spectral norm (through `Ellipse` in that PR).

### DIFF
--- a/kurbo/src/affine.rs
+++ b/kurbo/src/affine.rs
@@ -308,8 +308,72 @@ impl Affine {
     }
 
     /// Compute the determinant of this transform.
+    ///
+    /// # Geometric interpretation
+    ///
+    /// Consider a region transformed by this affine. The transformed region's area is the area of
+    /// the original region scaled by the absolute value of the determinant. A negative determinant
+    /// indicates orientation reversal.
+    #[inline]
     pub const fn determinant(self) -> f64 {
         self.0[0] * self.0[3] - self.0[1] * self.0[2]
+    }
+
+    /// Compute the square of the nuclear norm of this transform.
+    ///
+    /// This is the square of the [Schatten p-norm][schatten] with `p=1`, also known as the "trace norm."
+    ///
+    /// Returns the squared norm for efficiency; take the square root as necessary.
+    ///
+    /// # Geometric interpretation
+    ///
+    /// Consider a unit circle transformed by this affine. The nuclear norm is the sum of the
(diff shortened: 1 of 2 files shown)


kurbo/src/affine.rs
```diff
@@ -308,10 +308,74 @@ impl Affine {
...
+    /// (semi-major axis) of the Ellipse.
+    ///
+    /// [schatten]: <https://en.wikipedia.org/w/index.php?title=Matrix_norm&oldid=1348997593#Schatten_norms>
+    #[inline]
+    pub fn spectral_norm(self) -> f64 {
+        // Note a different calculation, returning the `_squared` form like our nuclear and
+        // Frobenius norms, could be `0.5 (frob^2 + sqrt(frob^4 - 4 det^2))`. In terms of operations
+        // it's a wash: one fewer sqrt if the user actually wants the squared form, but it uses more
+        // muls. More importantly, that form has worse numeric conditioning.
+        self.svd().0.x
```
raphlinus, line 420: Unfortunately when I check the asm (at least on mac) I still see the `atan2` call, even though its output is discarded. Pretty sure it's not being treated as a pure function. Not sure what's the best thing to do here.

raphlinus (approved):
These are correct, and I appreciated the links to Schatten norms and the geometric interpretations. I found a performance problem, but not sure what to do about it without making the code more verbose. Perhaps it's best to address that at the Rust level.

tomcur (commented):
> Unfortunately when I check the asm (at least on mac) I still see the atan2 call, even though its output is discarded. Pretty sure it's not being treated as a pure function. Not sure what's the best thing to do here.

Ah, that indeed is unfortunate; thank you for checking the asm. I mistakenly thought we had checked that, but apparently missed it. I agree it's likely that the compiler must assume side effects are possible.

I'll land this, as it's a pre-existing issue (there are other call sites where we assume inlining to get rid of the `atan2` call). I have some ideas that should not increase verbosity too much. I believe we can split the SVD into two private methods: one for the singular values and one for the rotation. For callers that need both, inlining should let the compiler do the optimization/interleaving that it wants to do, as the regular float ops are transparent to it.

PR #570, opened 2026-04-25
Add anisotropic expansion

jrmoulton:
Adds a new kurbo::expand API for expanding filled paths by a directional Vec2.

This PR was heavily assisted by Codex.

expand takes a `&BezPath` and offsets it by the given vector. The notion of “outward” is chosen from the overall winding of the path. The offset is applied per segment, perpendicular to the tangent. The magnitude is determined from the segment’s normal direction and the anisotropic x/y expand values (so different segment angles get different effective offset).  The offset segments are then connected together using the chosen join style.

For same-sign expansion, the implementation uses an affine normalization step: it rescales into a space where the requested Vec2 expand becomes a unit-radius scalar offset, performs the expansion there, then maps the result back. In that normalized space, the implementation preserves curves when possible by offsetting lines directly,
fitting cubic approximations to offset quadratics, and approximating cubics by quadratic pieces before offsetting them. When that path is not applicable, it falls back to a flattened contour offset. Purely linear contours and mixed-sign cases use the coordinate anisotropic path directly.

https://github.com/user-attachments/assets/3ab0141c-7503-4a37-8d38-709d20ea90f4

### DIFF
--- a/kurbo/src/stroke.rs
+++ b/kurbo/src/stroke.rs
@@ -12,7 +12,7 @@ use crate::common::FloatFuncs;
 
 use crate::{
     common::solve_quadratic, Affine, Arc, BezPath, CubicBez, Line, ParamCurve, ParamCurveArclen,
-    PathEl, PathSeg, Point, QuadBez, Vec2,
+    PathEl, PathSeg, Point, QuadBez, Shape, Vec2,
 };
 
 /// Defines the connection between two segments of a stroke.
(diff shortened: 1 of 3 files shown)


jneem:
I haven't looked at the implementation, but I'm confused by the description: specifically, the meaning of "contour-aware" and "winding-sensitive". On the one hand,

> positive values grow the filled region and negative values shrink it

makes me think you're adjusting the expansion direction so that whichever way a curve winds, a positive expansion will "enlarge" it. But then

> font outlines work naturally when outer contours and counters are oppositely wound

seems to contradict this (at least, if I was correct in re-interpreting "counters" to be a typo for "inner contours").

I guess I don't understand why a simple rule like "positive expansions are to the right when walking the contour" isn't easier...

jrmoulton:
Whoops I was inconsistent from different iterations. I'll clean up the description. 

> I guess I don't understand why a simple rule like "positive expansions are to the right when walking the contour" isn't easier...

If I understand correctly, that rule still depends on the input’s orientation: if two outlines are wound oppositely, “to the right” for the same positive value points to opposite sides and can invert emboldening. In this implementation we pick a single ‘outward’ sense from the overall signed area of the `BezPath` and apply it consistently across the whole outline, rather than relying on individual subpath directions. In my experiments, this does happen in practice across glyph outlines.

jneem:
> If I understand correctly, that rule still depends on the input’s orientation

Right, but it's a simple rule that's easy to describe, which makes it flexible for the user of the function and also easy to document. If you depend on the signed area, what happens if someone wants to operate on a non-simple path, like a figure-eight? The area could be zero, or have a difficult-to-predict sign because of rounding issues.

jrmoulton:
Yeah that is probably a better fit for the general case here in kurbo. and it should still be pretty easy for glyphs to build on

kurbo/src/stroke.rs
```diff
@@ -326,18 +425,788 @@ fn stroke_undashed(
...
+/// Build the affine-normalized scalar-offset representation for a same-sign anisotropic expand.
+fn normalized_expand_space(expand: Vec2, tolerance: f64) -> Option<NormalizedExpandSpace> {
+    // Same-sign axis-aligned expansion is equivalent to unit-radius offsetting
+    // after scaling into ellipse-normalized coordinates.
+    let common_sign = if expand.x != 0.0 {
+        expand.x.signum()
+    } else {
+        expand.y.signum()
+    };
+    if common_sign == 0.0 {
```
raphlinus, line 648: This will in fact never happen, as none of the values returned by signum test true for (partial) equality with zero.

kurbo/src/stroke.rs
```diff
@@ -290,23 +399,13 @@ fn stroke_undashed(
     ctx: &mut StrokeCtx,
 ) {
     ctx.reset();
-    ctx.join_thresh = 2.0 * tolerance / style.width;
+    ctx.join_thresh = 2.0 * tolerance / style.width.max(f64::EPSILON);
 
     for el in path {
         let p0 = ctx.last_pt;
         match el {
-            PathEl::MoveTo(p) => {
```
raphlinus, line 298: From what I can tell, this change is not related to expansion; I can't find anything else that calls `stroke_undashed_line_el`. I think it should be reverted.

kurbo/src/stroke.rs
```diff
@@ -258,6 +258,115 @@ pub fn stroke(
     ctx.output
 }
 
+/// Expand a filled path.
+///
+/// This applies a contour-aware offset to each subpath, which is useful for
```
raphlinus, line 263: I'm rewriting this based on an assumption we won't support shrinking. The existing docs seemed to be getting at this but are vague and wordy.

```suggestion
This expands a filled path by an amount given by the `expand` parameter. The two components must both be nonnegative. All subpaths in the input path must be closed.

The filled path is expanded no matter whether its signed area is positive or negative, assuming nonzero winding interpretation. In an even-odd interpretation where an inner contour winds in the same direction as an outer one, the adjustment applied to the inner contour will be in the wrong direction. Similarly, a path with both positive and negative winding numbers, such as a figure eight, will give results inconsistent with expansion.
```

kurbo/src/stroke.rs
```diff
@@ -326,18 +425,788 @@ fn stroke_undashed(
...
+    tolerance: f64,
+    contour_sign: f64,
+    style: &Stroke,
+    radius: f64,
+    from_unit: Affine,
+    points: &mut Vec<Point>,
+) {
+    match points.as_slice() {
+        [p0, p1] => {
+            let stroked = stroke(
```
raphlinus, line 567: We shouldn't have to deal with this case, we can assume that all subpaths are closed.

kurbo/src/stroke.rs
```diff
@@ -326,18 +425,788 @@ fn stroke_undashed(
...
+        start: quad.p0 - start_norm,
+        tan_start: tan0,
+        tan_end: tan1,
+        seg: OffsetSeg::Cubic(cubic),
+    })
+}
+
+/// Prefer a nonzero tangent, falling back to a chord direction for degenerate endpoints.
+fn nonzero_tangent(tangent: Vec2, fallback: Vec2) -> Option<Vec2> {
+    if tangent.hypot2() > f64::EPSILON {
```
raphlinus, line 902: Easier to just say `!= 0.0` here, as hypot2 is always nonnegative.

kurbo/src/stroke.rs
```diff
@@ -326,18 +425,788 @@ fn stroke_undashed(
...
+        Some(tangent)
+    } else if fallback.hypot2() > f64::EPSILON {
+        Some(fallback)
+    } else {
+        None
+    }
+}
+
+fn offset_quad_cubic(quad: QuadBez, distance: f64) -> Option<CubicBez> {
+    let tan0 = nonzero_tangent(quad.p1 - quad.p0, quad.p2 - quad.p0)?;
```
raphlinus, line 912: Fyi this will probably end up doing the right thing in cases where the endpoint derivatives are exactly zero. The normalize calls will produce finite results, then `det` will be zero, so you'll get the straight line. But I suspect you can still get bad results when the derivatives are very small.

I think I just need to find some time to write the robust version.

raphlinus (changes requested):
I really wanted to merge this, but find it falls short on several fronts. Fundamentally, it doesn't produce correct output (see [#kurbo > filled path expansion](https://xi.zulipchat.com/#narrow/channel/260979-kurbo/topic/filled.20path.20expansion/with/592207629) for screenshot). I believe the fundamental reason is that it's trying to do inner join optimisation, but gets the geometry wrong when computing the intersection point between the two offset segments.

There is lots wrong with the code. It's verbose and repetitive, especially considering that most of the primitives already exist in kurbo. It's also not doing tolerance-based subdivision or handle cusps, so doesn't meet its implied contract. Note that this typically *won't* be an important issue for the common case of glyphs, but of course is responsible for a great deal of the complexity in kurbo stroking.

The fallback to flattening is very unappealing, especially given that there were artifacts present in the output.

I can imagine landing a version of this, subject to the following changes:

* Correctness is "good enough" for glyph cases
  + I'm not asking for tolerance-based subdivision or robust cusp handling
  + but it does need to make the crossbar of the Roboto "e" a consistent width
  + likely getting the intersection point correct for join optimisation could get there
* Move it into its own module, so it doesn't potentially affect existing stroke logic
* Get rid of needless complexity for handling negative/mixed offsets and open subpaths

Then of course I'd want to redo the implementation, but that's a separate question.

raphlinus:
> > If I understand correctly, that rule still depends on the input’s orientation
> 
> Right, but it's a simple rule that's easy to describe, which makes it flexible for the user of the function and also easy to document. If you depend on the signed area, what happens if someone wants to operate on a non-simple path, like a figure-eight? The area could be zero, or have a difficult-to-predict sign because of rounding issues.

I rewrote the description, hopefully it's clearer now. I think it's reasonable to not handle these kinds of cases. They absolutely won't occur for the common case of dealing with glyphs.

jrmoulton:
Superseded by #575

PR #573, opened 2026-04-29
Unify workspace dependencies.

xStrom:
Now that this workspace has two packages it's worth keeping track of dependency versions at the workspace level, which matches how other Linebender repos do it.

This revealed that we were pulling in two different versions of `criterion`. Looks like `kurbo` can bench fine with the unified newer version too.


tomcur (approved):
Nice, thanks!

PR #575, opened 2026-05-03
Add anisotropic expansion methods for paths

raphlinus:
Adds methods for anisotropic expansion of paths, which is expected to be useful for stem darkening of glyphs. It is intended to work well when the amount of expansion is small or moderate relative to the curvature of the path segments.

The computational geometry is very sophisticated, and uses unpublished algorithms that have been discussed in the Zulip, in particular the "two point linear" algorithm is a refinement of what's described in [#kurbo > Fast stroke expansion state of play](https://xi.zulipchat.com/#narrow/channel/260979-kurbo/topic/Fast.20stroke.20expansion.20state.20of.20play/with/561307472). The algorithms have been chosen for speed and simplicity, as well as excellent accuracy in the small offset case. They're not ideal for very fine tolerances, but then these methods don't do tolerance-based subdivision, nor do they have proper cusp handling.

There's a fair amount of "math golfing" to make computations efficient, which has the effect of making the code more concise, but also likely denser. I also found it difficult to reason about negative expansion values. It is *almost* the case that expansion by an anisotropic expansion factor is equivalent to doing an inverse affine, expansion by a unit offset, then the forward affine, but when you think about it more carefully, that doesn't quite get the signs right; nonuniform scaling by two negative transform components still has a positive determinant. In practice I made it work by sprinkling in `.abs()` calls until the output was right. I suspect there is a clear mathematical formulation that automatically takes care of the signs, but I was not smart enough to find it.

I created a new `Diagonal2` type to represent the expansion, especially because I found it useful to model it as a linear transformation. But given that the math around signs is not as clean as I would have hoped, I am having second thoughts. Of our existing "tuple of two floats" types, I think `Size` might be the best fit, and I am open to making that change. Alternatively, it's possible that `Diagonal2` deserves its place, and its role could be expanded. For example, it easily deserves a `From` impl into `Affine`. I'm also thinking it could get `From<f64>`, giving uniform expansion. I found the `Mul` impls against `Point` and `Vec2` to be very convenient and intuitive, and those are not appropriate for `Size`. In any case, I think the current state is pretty good and none of the alternatives I'm thinking are obviously better.

I haven't done any validation, but suspect that expansions along one axis (the other zero) already work with polyline input. Curves would be conceptually straightforward to add but a fair amount of code. The underlying geometry operation is splitting into monotonic sections, for which we already have some support (`ParamCurveExtrema`).

### DIFF
--- a/kurbo/src/expand.rs
+++ b/kurbo/src/expand.rs
@@ -1,0 +1,393 @@
+// Copyright 2026 the Kurbo Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::ops::Mul;
+
+use crate::{
+    BezPath, CubicBez, Join, ParamCurve, ParamCurveDeriv, PathEl, PathSeg, Point, QuadBez, Vec2,
+};
+
+#[cfg(not(feature = "std"))]
+use crate::common::FloatFuncs;
+
+/// A diagonal matrix, suitable for representing anisotropic scaling.
+#[derive(Clone, Copy, Debug)]
+pub struct Diagonal2 {
+    xx: f64,
+    yy: f64,
+}
+
+struct ExpandCtx {
+    expand: Diagonal2,
+    join: Join,
+    miter_limit: f64,
+    tolerance: f64,
+    result: BezPath,
+    first_n: Option<Vec2>,
+    first_tan: Vec2,
+    last_pt: Point,
+    last_n: Option<Vec2>,
+    last_tan: Vec2,
+}
+
+struct CubicCtx {
+    q: QuadBez,
+    utan0: Vec2,
+    utan1: Vec2,
+}
+
+struct TwoPointSample {
+    a_n: f64,
+    b_n: f64,
+    c_n: f64,
+}
+
+impl Diagonal2 {
+    /// Create a diagonal matrix.
+    pub fn new(xx: f64, yy: f64) -> Self {
+        Self { xx, yy }
+    }
+
+    /// Matrix inverse.
+    ///
+    /// Will of course produce infinities if a component is zero.
+    pub fn inv(self) -> Self {
+        Diagonal2::new(1.0 / self.xx, 1.0 / self.yy)
+    }
+
+    /// Scale a normal vector.
+    ///
+    /// This is mathematically equivalent to `self * (self * n).normalize()`
+    /// but handles zeros.
+    ///
+    /// Note that `n` need not be unit length.
+    pub fn scale_normal(self, n: Vec2) -> Vec2 {
+        let z = self * n;
+        let z_hypot2 = z.hypot2();
+        if z_hypot2 == 0.0 {
+            Vec2::ZERO
+        } else {
+            let inv_scale = 1.0 / z_hypot2.sqrt();
+            self * z * inv_scale
+        }
+    }
+
+    /// Determine whether transform is uniform scale.
+    pub fn as_uniform(self) -> Option<f64> {
+        if self.xx == self.yy {
+            Some(self.xx)
+        } else {
+            None
+        }
+    }
+}
+
+impl Mul<Vec2> for Diagonal2 {
+    type Output = Vec2;
+
+    fn mul(self, rhs: Vec2) -> Vec2 {
+        Vec2::new(self.xx * rhs.x, self.yy * rhs.y)
+    }
+}
+
+impl Mul<Point> for Diagonal2 {
+    type Output = Point;
+
+    fn mul(self, rhs: Point) -> Point {
+        Point::new(self.xx * rhs.x, self.yy * rhs.y)
+    }
+}
+
+// Note: a bunch more ops could be implemented, but for now we'll stick to what we need.
+
+/// Expand a path.
+///
+/// Expands a filled path by the expansion, which allows separate x and y factors. The
+/// path (and the result) is interpreted according to the nonzero winding rule.
+///
+/// Each subpath must be closed.
+///
+/// The tolerance is mostly for joins and robustness; it is not used to guide
+/// subdivision. Rather, each Bézier segment in the input generally results in one
+/// cubic Bézier in the output. Thus, it is not expected to work well when the
+/// expansion factor is large compared with the radius of curvature on the input.
+pub fn expand(
+    path: impl IntoIterator<Item = PathEl>,
+    expand: Diagonal2,
+    join: Join,
+    miter_limit: f64,
+    tolerance: f64,
+) -> BezPath {
+    // TODO: evaluate area, adjust sign of expand
+    let result = BezPath::new();
+    let mut ctx = ExpandCtx {
+        expand,
+        join,
+        miter_limit,
+        tolerance,
+        result,
+        first_n: None,
+        first_tan: Vec2::default(),
+        last_pt: Point::default(),
+        last_n: None,
+        last_tan: Vec2::default(),
+    };
+    let mut first_pt = Point::default();
+    for el in path {
+        match el {
+            PathEl::MoveTo(point) => {
+                if ctx.last_n.is_some() {
+                    ctx.do_close_path(first_pt);
+                }
+                first_pt = point;
+                ctx.last_pt = point;
+            }
+            PathEl::LineTo(p1) => ctx.do_line(p1),
+            PathEl::QuadTo(p1, p2) => ctx.do_quad(p1, p2),
+            PathEl::CurveTo(p1, p2, p3) => ctx.do_cubic(p1, p2, p3),
+            PathEl::ClosePath => ctx.do_close_path(first_pt),
+        }
+    }
+    ctx.result
+}
+
+impl ExpandCtx {
+    /// Helper function to determine if a distance is within tolerance.
+    fn in_tolerance(&self, v: Vec2) -> bool {
+        v.hypot2() < self.tolerance * self.tolerance
+    }
+
+    /// Process a line segment. Includes initial join.
+    fn do_line(&mut self, p1: Point) {
+        let tan = p1 - self.last_pt;
+        let n = self.expand.scale_normal(tan.turn_90());
+        self.do_join(n, tan, true);
+        let out_p1 = p1 + n;
+        self.result.line_to(out_p1);
+        self.last_n = Some(n);
+        self.last_tan = tan;
+        self.last_pt = p1;
+    }
+
+    /// Process a quadratic Bézier segment. Includes initial join.
+    fn do_quad(&mut self, p1: Point, p2: Point) {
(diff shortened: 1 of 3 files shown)


kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+/// cubic Bézier in the output. Thus, it is not expected to work well when the
+/// expansion factor is large compared with the radius of curvature on the input.
+pub fn expand_path(
+    path: impl Shape,
+    mut expand: Diagonal2,
+    join: Join,
+    miter_limit: f64,
+    tolerance: f64,
+) -> BezPath {
+    if path.area() >= 0.0 {
```
waywardmonkeys, line 134: This uses the path's area rather than the contour's area:

```rust
path.extend(Rect::new(0.0, 0.0, 100.0, 100.0).path_elements(1e-3));
path.extend(Rect::new(30.0, 70.0, 70.0, 30.0).path_elements(1e-3));

let expanded = expand_path(path, Diagonal2::new(10.0, 10.0), Join::Miter, 4.0, 1e-3);
assert!((expanded.area() - 14_000.0).abs() < 1e-9);
```

Which expectation is correct?

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+                    ctx.do_close_path(first_pt);
+                }
+                first_pt = point;
+                ctx.last_pt = point;
+            }
+            PathEl::LineTo(p1) => ctx.do_line(p1),
+            PathEl::QuadTo(p1, p2) => ctx.do_quad(p1, p2),
+            PathEl::CurveTo(p1, p2, p3) => ctx.do_cubic(p1, p2, p3),
+            PathEl::ClosePath => ctx.do_close_path(first_pt),
+        }
```
waywardmonkeys, line 178: What should the behavior be for a path that isn't closed?  Should it implicitly close it?

```rust
path.move_to((0.0, 0.0));
path.line_to((100.0, 0.0));
path.line_to((100.0, 100.0));
path.line_to((0.0, 100.0));

let expanded = expand_path(path, Diagonal2::new(10.0, 10.0), Join::Miter, 4.0, 1e-3);
assert_eq!(expanded.elements().last(), Some(&kurbo::PathEl::ClosePath));
```

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+        self.last_tan = q1;
+        self.last_pt = p2;
+    }
+
+    /// Process a cubic segment. Includes initial join.
+    fn do_cubic(&mut self, p1: Point, p2: Point, p3: Point) {
+        let einv = self.expand.inv();
+        let c = CubicBez::new(einv * self.last_pt, einv * p1, einv * p2, einv * p3);
+        let (tan0, tan1) = PathSeg::Cubic(c).tangents();
+        let utan0 = tan0.normalize();
```
waywardmonkeys, line 247: A degenerate cubic can emit a NaN here, I think.

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+    //let angle_err = delta_tan.cross(utan);
+    //let err_est = 0.16 * angle_err.abs();
+    Some((a, b))
+}
+
+fn two_point_linear(cx: &CubicCtx) -> Option<(f64, f64)> {
+    const T0: f64 = 0.35;
+    let s = [T0, 1.0 - T0].map(|t| {
+        let q = cx.q.eval(t).to_vec2();
+        let qscale = 1.0 / q.length();
```
waywardmonkeys: A degenerate cubic can produce a NaN here as well, I think.

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+        let b1 = 3.0 * mt * t * mt;
+        let b2 = 3.0 * mt * t * t;
+        let b3 = t * t * t;
+        let a_n = b1 * utan0_n;
+        let b_n = b2 * utan1_n;
+        let c_n = (b0 + b1) * utan0xn + (b2 + b3) * utan1xn - 1.0;
+        TwoPointSample { a_n, b_n, c_n }
+    });
+    let det = s[0].a_n * s[1].b_n - s[1].a_n * s[0].b_n;
+    if det.abs() < 1e-12 {
```
waywardmonkeys: If `det` is NaN, then this will still be false ... and in the case of `"M0,0 C0,0 0,0 0,0 Z", I think it can be NaN (according to Codex).

waywardmonkeys: It has suggestions about where to normalize / check for degenerate cubics, but I'm not going to teach my grandmother to suck eggs when I'm just repeating what it says. :)

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+        }
+    }
+
+    fn do_close_path(&mut self, first_pt: Point) {
+        // maybe do this test inside do_line for all lines
+        if first_pt.distance_squared(self.last_pt) > self.tolerance * self.tolerance {
+            self.do_line(first_pt);
+        }
+        self.do_join(self.first_n.unwrap(), self.first_tan, true);
+        self.result.close_path();
```
waywardmonkeys: See what happens if you test this with `"M0,0 Z"`.

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+                                    n.x,
+                                    n.y,
+                                    nt.x,
+                                    nt.y,
+                                    self.last_pt.x,
+                                    self.last_pt.y,
+                                ]);
+                                let arc: Arc =
+                                    Arc::new(Point::ORIGIN, (1.0, 1.0), -angle, angle, 0.0);
+                                let tolerance = self.tolerance * einv.xx.abs().min(einv.yy.abs());
```
waywardmonkeys: Codex says this, but I'm not informed enough to verify it myself, so just passing it along:

> ... scales round-join arc tolerance in the wrong direction. The unit arc is later affine-scaled by roughly the expansion size, so tolerance should be divided by a scale bound, not multiplied by min(abs(expand)). Large expansions under-subdivide; tiny expansions over-subdivide

kurbo/src/expand.rs
```diff
@@ -0,0 +1,551 @@
...
+
+#[cfg(not(feature = "std"))]
+use crate::common::FloatFuncs;
+
+/// A diagonal matrix, suitable for representing anisotropic scaling.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Diagonal2 {
+    xx: f64,
```
waywardmonkeys: Should these be pub? The type is kind of strange being serializable, can be created, but no accessors.

waywardmonkeys (commented):
Some from Codex, some from me!

kurbo/src/expand.rs
raphlinus: My personal feeling is that the docstring on the method disclaims the behavior in this case. And I don't know what "the contour's area" refers to.

kurbo/src/expand.rs
raphlinus: Yeah, I think that's more robust, I'll do it.

kurbo/src/expand.rs
raphlinus: Codex is right.

kurbo/src/expand.rs
raphlinus: It now handles at least this case correctly. I think it's worth doing more validation against adversarial inputs, though I tried pretty hard to reason through those (this det thing was in fact on my radar and I had a fix in my editor window which did an early-return of `None`, but it broke autovectorization so I didn't commit it).

raphlinus (commented):
Thanks, these are indeed quite helpful! I've addressed most of them but I'll take another pass tomorrow.

waywardmonkeys:
I ran a fuzz of 250,000 generated paths with lines / quads / cubics, duplicate points, optional close, and all three types of joins. No panics or non-finite output.

But did find a new thing ...

`ExpandCtx::do_close_path` clears `last_n` but leaves `first_n`/`first_tan` stale. After a real subpath, a later degenerate subpath reuses the previous contour’s first normal and emits a ghost empty contour.

Minimal shape:

```rust
path.extend(Rect::new(0.0, 0.0, 10.0, 10.0).path_elements(1e-3));
let expected = expand_path(&path, Diagonal2::new(1.0, 1.0), Join::Miter, 4.0, 1e-3);

path.move_to((20.0, 20.0));
path.close_path();
let actual = expand_path(&path, Diagonal2::new(1.0, 1.0), Join::Miter, 4.0, 1e-3);

assert_eq!(actual, expected);
```

Current output appends:

```
MoveTo((20.0, 19.0)), ClosePath
```

Codex also complained that this diagonal fails while meeting the documented requirement that the expansion factors should be positive: `Diagonal2::new(f64::MIN_POSITIVE, 10.0)` ... I think you don't care though. :)

kurbo/src/expand.rs
waywardmonkeys: The second rect has a negative area / opposite winding. The wording isn't clear to me in the doc string about what should happen here.

waywardmonkeys:
This is a fun one:

`expand_path` now treats open subpaths as closed, but still chooses the sign from `path.area()` before adding the implicit close. For open paths, `Shape::area()` is not the same as the area of the closed contour, so an open path and the same path with explicit `Z` / `ClosePath` can expand in opposite directions.

Minimal repro:

```rust
let mut open = BezPath::new();
open.move_to((-3.0, -3.0));
open.line_to((-3.0, -2.0));
open.line_to((-2.0, -3.0));

let mut closed = open.clone();
closed.close_path();

let a = expand_path(&open, Diagonal2::new(0.25, 0.25), Join::Miter, 4.0, 1e-3);
let b = expand_path(&closed, Diagonal2::new(0.25, 0.25), Join::Miter, 4.0, 1e-3);

assert_eq!(a, b);
```

Current facts for that path:

* `open.area() = 1`
* `closed.area() = -0.5`

So line 134 selects the opposite sign for the open form even though lines 180-181 later close it.

raphlinus:
The unclosed path area one is interesting. It feels like fixing it may be out of the scope of this PR, as it's unlikely to affect glyph rendering, but it would be good to get resolved.

waywardmonkeys:
I'm willing to approve this with 3 caveats:

* We should do a follow up to fix that unclosed path area issue.
* I haven't actually run this and seen it in action.
* My approval is based partly on having run through it extensively with Codex, including the work mentioned above which included running hundreds of thousands of test cases.

waywardmonkeys (approved):

PR #578, opened 2026-05-06
fix: Dash iterator zero-length LineTo noise

### LINKED ISSUE #577: Dash iterator zero-length LineTo noise

Example:

```rust
path.move_to((0., 0.));
path.move_to((3., 0.));
path.move_to((3., 3.));
let dashes = [3., 1.];

```

RobertBrewitz:
Resolves and adds tests to quality assures the fix for https://github.com/linebender/kurbo/issues/577

- dash_iter now treats an exactly exhausted dash the same as an overshot one at initialization
- dash_iter now advances dash phase immediately when a dash ends exactly on a path vertex, instead of carrying an active zero remainder into the next segment.
- New dashed-stroke regression tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ You can find its changes [documented below](#0131-2026-05-13).
 
 This release has an [MSRV][] of 1.85.
 
+## Fixed
+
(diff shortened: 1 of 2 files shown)


RobertBrewitz:
Closing in favor of https://github.com/linebender/kurbo/pull/597

PR #579, opened 2026-05-07
Add subpath closing iterator and change Shape::area() behavior

raphlinus:
This patch changes the expectation for the `area` method of the `Shape` trait, so that open subpaths are implicitly closed.

Updates the implementations for QuadBez and CubicBez.

Note: this is technically a breaking change, but it only affects external implementations of the `Shape` trait, and it seems exceedingly unlikely this will cause actual problems.

Work in progress. The implementation for `Arc` no longer meets the spec.

### DIFF
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -793,6 +793,69 @@ impl Mul<&BezPath> for TranslateScale {
     }
 }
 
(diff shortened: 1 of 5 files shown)


raphlinus:
Closing in favor of #580, which is less invasive and does not make any potentially semver-breaking changes.

PR #580, opened 2026-05-07
Close subpaths when computing area for expansion

raphlinus:
Previously the area() method on the Shape trait was used to compute the area for the purposes of determining the orientation for expansion. However, this method is not valid unless the input has closed subpaths, which can lead to inconsistent results.

This patch adds an iterator to close subpaths, and computes area based on that. The iterator is made publicly available, as it may be handy.

### DIFF
--- a/kurbo/src/bezpath.rs
+++ b/kurbo/src/bezpath.rs
@@ -793,6 +793,69 @@ impl Mul<&BezPath> for TranslateScale {
     }
 }
 
+/// Close all open subpaths in a path, expressed as iterator transform.
+pub fn close_subpaths<I>(elements: I) -> CloseSubpaths<I::IntoIter>
+where
+    I: IntoIterator<Item = PathEl>,
+{
+    CloseSubpaths {
+        elements: elements.into_iter(),
+        state: CloseSubpathState::Start,
+    }
+}
+
+/// An iterator that closes all open subpaths.
+///
+/// This struct is created by the [`segments`] function.
+#[derive(Clone)]
+pub struct CloseSubpaths<I: Iterator<Item = PathEl>> {
+    elements: I,
+    state: CloseSubpathState,
+}
+
+#[derive(Clone)]
+enum CloseSubpathState {
+    Start,
+    InSubpath,
+    ClosedLast,
+    PendingMoveTo(Point),
+}
+
+impl<I: Iterator<Item = PathEl>> Iterator for CloseSubpaths<I> {
+    type Item = PathEl;
+
+    fn next(&mut self) -> Option<PathEl> {
+        match self.state {
+            CloseSubpathState::Start => {
+                let el = self.elements.next();
+                if !matches!(el, Some(PathEl::ClosePath)) {
+                    self.state = CloseSubpathState::InSubpath;
+                }
+                el
+            }
+            CloseSubpathState::InSubpath => {
+                let el = self.elements.next();
+                match el {
+                    None => {
+                        self.state = CloseSubpathState::ClosedLast;
+                        Some(PathEl::ClosePath)
+                    }
+                    Some(PathEl::MoveTo(point)) => {
+                        self.state = CloseSubpathState::PendingMoveTo(point);
+                        Some(PathEl::ClosePath)
+                    }
+                    _ => el,
+                }
+            }
+            CloseSubpathState::ClosedLast => None,
+            CloseSubpathState::PendingMoveTo(point) => {
+                self.state = CloseSubpathState::Start;
+                Some(PathEl::MoveTo(point))
+            }
+        }
+    }
+}
+
 /// Transform an iterator over path elements into one over path
 /// segments.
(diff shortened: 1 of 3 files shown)


kurbo/src/bezpath.rs
```diff
@@ -793,6 +793,69 @@ impl Mul<&BezPath> for TranslateScale {
...
+{
+    CloseSubpaths {
+        elements: elements.into_iter(),
+        state: CloseSubpathState::Start,
+    }
+}
+
+/// An iterator that closes all open subpaths.
+///
+/// This struct is created by the [`segments`] function.
```
waywardmonkeys: Should say `close_subpaths` now, not `segments`

kurbo/src/bezpath.rs
```diff
@@ -793,6 +793,69 @@ impl Mul<&BezPath> for TranslateScale {
     }
 }
 
+/// Close all open subpaths in a path, expressed as iterator transform.
+pub fn close_subpaths<I>(elements: I) -> CloseSubpaths<I::IntoIter>
```
waywardmonkeys: Should this be `pub`? or just `pub(crate)` ?

kurbo/src/bezpath.rs
```diff
@@ -793,6 +793,69 @@ impl Mul<&BezPath> for TranslateScale {
...
+        elements: elements.into_iter(),
+        state: CloseSubpathState::Start,
+    }
+}
+
+/// An iterator that closes all open subpaths.
+///
+/// This struct is created by the [`segments`] function.
+#[derive(Clone)]
+pub struct CloseSubpaths<I: Iterator<Item = PathEl>> {
```
waywardmonkeys: As above, if `close_subpaths` becomes `pub(crate)`, so should this ....

kurbo/src/bezpath.rs
```diff
@@ -793,6 +793,69 @@ impl Mul<&BezPath> for TranslateScale {
...
+                match el {
+                    None => {
+                        self.state = CloseSubpathState::ClosedLast;
+                        Some(PathEl::ClosePath)
+                    }
+                    Some(PathEl::MoveTo(point)) => {
+                        self.state = CloseSubpathState::PendingMoveTo(point);
+                        Some(PathEl::ClosePath)
+                    }
+                    _ => el,
```
waywardmonkeys, line 851: Shouldn't this handle `Some(PathEl::ClosePath)` to avoid duplicate close paths for already-closed subpaths?

(But this makes the name `ClosedLast` slightly confusing and maybe `BetweenSubpaths` would be better in this part of this case?

Test to add:

```
  #[test]
  fn close_subpaths_does_not_duplicate_existing_closepath() {
      let path = BezPath::from_vec(vec![
          PathEl::MoveTo((0.0, 0.0).into()),
          PathEl::LineTo((10.0, 0.0).into()),
          PathEl::LineTo((10.0, 10.0).into()),
          PathEl::ClosePath,
          PathEl::MoveTo((20.0, 0.0).into()),
          PathEl::LineTo((30.0, 0.0).into()),
      ]);

      let closed = close_subpaths(path.iter()).collect::<Vec<_>>();

      assert_eq!(
          closed,
          vec![
              PathEl::MoveTo((0.0, 0.0).into()),
              PathEl::LineTo((10.0, 0.0).into()),
              PathEl::LineTo((10.0, 10.0).into()),
              PathEl::ClosePath,
              PathEl::MoveTo((20.0, 0.0).into()),
              PathEl::LineTo((30.0, 0.0).into()),
              PathEl::ClosePath,
          ]
      );
  }
```

waywardmonkeys (commented):
Maybe we should get an explicit test for the thing being fixed here, maybe not (with `expand_path`)?

Like this in `kurbo/src/expand.rs`:

```
  #[test]
  fn expand_open_subpath_uses_implicit_close_for_area_sign() {
      let mut open = BezPath::new();
      open.move_to((100.0, 100.0));
      open.line_to((110.0, 100.0));
      open.line_to((100.0, 110.0));

      let mut closed = open.clone();
      closed.close_path();

      let expand = Diagonal2::new(1.0, 1.0);
      let open_expanded = expand_path(open, expand, Join::Miter, 4.0, 1e-3);
      let closed_expanded = expand_path(closed, expand, Join::Miter, 4.0, 1e-3);

      assert_eq!(open_expanded, closed_expanded);
  }
```

kurbo/src/bezpath.rs
raphlinus: I'll do `pub(crate)` for now and if someone wants it public it can easily be opened.

waywardmonkeys (approved):
Looks good. Remaining issue is minor but if you poll the close_subpaths again after it returns `Non`, it'll return a `Some(ClosePath)`, but given that it isn't pub and how it is used, that shouldn't be an issue for now.

waywardmonkeys (approved):

PR #582, opened 2026-05-08
Add reuseable context to new expand_* functions

taj-p:
While reviewing https://github.com/linebender/vello/pull/1628#discussion_r3210834436, I noticed that the new expand methods allocate a new `BezPath` per call.

The proposed variant (exposing a `ExpandCtx`) parallels the existing `StrokeCtx`.

https://github.com/linebender/kurbo/blob/6836244be0b280ff68728f100fca0a5fcf62638a/kurbo/src/stroke.rs#L200-L220

Further, the proposed `ExpandCtx` provides the ability to better integrate with a `BezPath` free list. For example, by providing methods into `ExpandCtx` that allow a caller to "take" ownership of the underlying `BezPath` we can more easily move that allocation into a glyph cache (without cloning). 

Happy to change any and all naming - not attached to anything here.

cc @jrmoulton

### DIFF
--- a/kurbo/src/expand.rs
+++ b/kurbo/src/expand.rs
@@ -20,17 +20,80 @@ pub struct Diagonal2 {
     pub yy: f64,
 }
 
-struct ExpandCtx {
+/// Reusable context for [`expand_path_with`] and [`expand_path_signed_with`].
+#[derive(Debug)]
+pub struct ExpandCtx {
     expand: Diagonal2,
     join: Join,
     miter_limit: f64,
     tolerance: f64,
-    result: BezPath,
+    output: BezPath,
     first_n: Option<Vec2>,
     first_tan: Vec2,
     last_pt: Point,
     last_n: Option<Vec2>,
     last_tan: Vec2,
+}
+
+impl Default for ExpandCtx {
+    fn default() -> Self {
+        Self {
+            expand: Diagonal2::new(1.0, 1.0),
+            join: Join::Bevel,
+            miter_limit: 0.0,
+            tolerance: 0.0,
+            output: BezPath::new(),
+            first_n: None,
+            first_tan: Vec2::default(),
+            last_pt: Point::default(),
+            last_n: None,
+            last_tan: Vec2::default(),
+        }
+    }
+}
+
+impl ExpandCtx {
+    /// Get the output path.
+    #[inline(always)]
+    pub fn output(&self) -> &BezPath {
+        &self.output
+    }
+
+    /// Create a new `ExpandCtx` with the given output path.
+    #[inline(always)]
+    pub fn with_output(output: BezPath) -> Self {
+        Self {
+            output,
+            ..Self::default()
+        }
+    }
+
+    /// Replace the output path with the given path, returning the previous one.
+    #[inline(always)]
+    pub fn replace_output(&mut self, output: BezPath) -> BezPath {
+        core::mem::replace(&mut self.output, output)
+    }
+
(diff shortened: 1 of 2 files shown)


kurbo/src/expand.rs
```diff
@@ -20,19 +20,82 @@ pub struct Diagonal2 {
...
+    #[inline(always)]
+    pub fn replace_output(&mut self, output: BezPath) -> BezPath {
+        core::mem::replace(&mut self.output, output)
+    }
+
+    /// Take the output path by value, leaving an empty `BezPath` in its place.
+    #[inline(always)]
+    pub fn take_output(&mut self) -> BezPath {
+        core::mem::take(&mut self.output)
+    }
```
taj-p: I dropped this. Consumers can simply `replace_output(BezPath::default())` for the same behaviour. But happy to add back if requested 🙏

kurbo/src/expand.rs
```diff
@@ -126,15 +183,30 @@ impl core::ops::Mul<Point> for Diagonal2 {
...
     tolerance: f64,
 ) -> BezPath {
+    let mut ctx = ExpandCtx::default();
+    expand_path_with(path, expand, join, miter_limit, tolerance, &mut ctx);
+    ctx.output
+}
+
+/// The same as [`expand_path`], but using a caller-provided [`ExpandCtx`] to enable
+/// allocation re-use.
+pub fn expand_path_with(
```
raphlinus, line 198: It feels appropriate to have this (and the signed version) as a `&mut self` method on `ExpandCtx`, as opposed to taking a context argument.

kurbo/src/expand.rs
```diff
@@ -20,19 +20,76 @@ pub struct Diagonal2 {
...
+        self.last_n = None;
+        self.last_tan = Vec2::default();
+
+        self.expand = expand;
+        self.join = join;
+        self.miter_limit = miter_limit;
+        self.tolerance = tolerance;
+    }
+}
+
```
raphlinus, line 92: Thinking about the ergonomics in the simple case, I'm wondering if it might be nice to have an into `BezPath` (ie `From<ExpandCtx> for BezPath`). Then the simple case looks like:

```rust
    let mut ctx = ExpandCtx::default();
    ctx.expand(...);
    let bezpath: BezPath = ctx.into();
```

Not quite as short as the free function, but I'm thinking this could be a reminder that there's an allocation and a nudge to maybe get rid of it.

raphlinus (commented):
I think this is a good idea and am ok with the PR landing as-is. One question I have is whether we need the API surface duplication and just have everything on ExpandCtx, get rid of the free functions.

waywardmonkeys (commented):
I know it is obvious to us, but we should be a bit more clear in the API docs. They should explicitly say _with functions return output through ExpandCtx::output(), that the output is cleared/reused on the next call, and that callers must copy/consume it before reusing the context.

kurbo/src/expand.rs
nicoburns: I don't really like using `Into` for this. I feel like if we're going to do this it ought to just be a regular method (`fn into_bez_path(self)` or similar). A bonus of this is that you won't run into type inference issues.

kurbo/src/expand.rs
nicoburns: I'd be in favour of adding this method. Seems like a common case, and there's little cost to having it.
(importance level: low)

kurbo/src/expand.rs
raphlinus: Fair enough, the type inference concern is legit.

PR #583, opened 2026-05-12
Fix failed `atan2` elimination by splitting SVD into two methods

tomcur:
One method calculates the singular values, the other the rotation.

This is motivated by the observations in https://github.com/linebender/kurbo/pull/569#discussion_r3186050826: we're often interested only in the scaling, but the `atan2` call for the rotation is not optimized out by the compiler when it's dead code (probably because the compiler does not know the call is side-effect free). With this change, the call no longer shows up in places where we only use the scales.

This is somewhat more verbose, but the methods are not public.

The overlap between the two calculations is quite small, and when we do need both (e.g., in `Ellipse::radii_and_rotation`), the inlining of both methods should still allow the compiler to optimize (though I haven't directly verified).

### DIFF
--- a/kurbo/src/affine.rs
+++ b/kurbo/src/affine.rs
@@ -417,7 +417,7 @@ impl Affine {
         // Frobenius norms, could be `0.5 (frob^2 + sqrt(frob^4 - 4 det^2))`. In terms of operations
         // it's a wash: one fewer sqrt if the user actually wants the squared form, but it uses more
         // muls. More importantly, that form has worse numeric conditioning.
-        self.svd().0.x
+        self.svd_singular_values().x
     }
 
     /// Compute the inverse transform.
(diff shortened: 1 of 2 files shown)


kurbo/src/affine.rs
```diff
@@ -787,40 +786,36 @@ mod tests {
...
 
-        let s = mat(1., 0., 0., -1.).svd().0;
+        let s = mat(1., 0., 0., -1.).svd_singular_values();
         assert_near(s.to_point(), Point::new(1., 1.));
 
-        let s = mat(1., 1., 1., 1.).svd().0;
-        assert_near(s.to_point(), Point::new(2., 0.));
-
-        let s = mat(1., 1., 1., 1.).svd().0;
+        let s = mat(1., 1., 1., 1.).svd_singular_values();
```
tomcur, line 795: This test was duplicated. Removed one, updated the other.

raphlinus (approved):
Thanks, looks good. Glad we didn't make `svd` a public API.

PR #584, opened 2026-05-12
fix: drop zero-length LineTo when dash transition lands on a vertex

### LINKED ISSUE #577: Dash iterator zero-length LineTo noise

SAY-5:
When a dash transition lands exactly on a path vertex, the dash iterator emitted a zero-length `LineTo` at the corner before transitioning to the next dash; this skips that empty segment. Closes #577.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This release has an [MSRV][] of 1.85.
 
 - Improved numerical robustness of winding numbers for paths. ([#537][] by [@jneem][])
 - `Affine::pre_rotate_about` calculation. ([#567][] by [@dannymcgee][])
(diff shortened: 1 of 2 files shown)


RobertBrewitz (changes requested):
Hey, there are some more things surrounding this fix that needs to be considered, see https://github.com/linebender/kurbo/pull/578 for inspiration if you want to take another stab at it and improve further.

This PR fixes the symptom but leaves behind stale state in the dash iterator.

SAY-5:
Thanks for the pointer. Since #578 already covers issue #577 more completely and is yours, it makes more sense to land that one. Happy to defer to #578.

PR #585, opened 2026-05-12
Fix odd-length dash patterns swapping dashes and gaps when `dash_length` increases

Keavon:
Given any odd-length dash pattern like `[10]` or `[10, 3, 20]` (shown in the demo below), increasing the `dash_length` property would incorrectly cause the stroked dash areas and the non-stroked gap areas to swap places, in violation of the SVG spec.

Before:

https://github.com/user-attachments/assets/e5150ede-918f-4a6b-a52b-023c041c8235

After:

https://github.com/user-attachments/assets/ba384948-f192-49cc-b242-117299704403

This fixes Kurbo to comply with the [SVG spec](https://www.w3.org/TR/SVG2/painting.html#StrokeDasharrayProperty):

image

AI disclosure: I used Claude Code to pinpoint the bug and find the appropriate solution, which I then cleaned up to be a tidy one-liner. I've tested it in Graphite and confirmed (as shown in the videos above) that it works as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,8 +34,9 @@ This release has an [MSRV][] of 1.85.
 
 ## Fixed
 
-- Improved numerical robustness of winding numbers for paths. ([#537][] by [@jneem][])
+- Improve numerical robustness of winding numbers for paths. ([#537][] by [@jneem][])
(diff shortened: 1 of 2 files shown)


tomcur (approved):
Thanks! I believe this correctly matches the SVG spec (and is the nicer behavior in general).

PR #588, opened 2026-05-19
fix: avoid u64 overflow in solve_itp for large nmax values

greysquirr3l:
## Problem

When `epsilon` is very small (e.g. `1e-24`), `nmax` in `solve_itp_fallible` can exceed 63. The current expression:

```rust
let mut scaled_epsilon = epsilon * (1u64 << nmax) as f64;
```

overflows `u64` when `nmax >= 64`:
- **Debug builds**: panic with "attempt to shift left with overflow"
- **Release builds**: silent wrong result — x86 masks shift amount to 6 bits, so `<< 64` wraps to `<< 0`, producing `scaled_epsilon = epsilon * 1` instead of `epsilon * 2^nmax`

### How `nmax` gets large

```rust
let n1_2 = (((b - a) / epsilon).log2().ceil() - 1.0).max(0.0) as usize;
let nmax = n0 + n1_2;
```

For `b - a = 1.0` and `epsilon = 1e-24`, `log2(1e24) ≈ 79.7`, giving `nmax ≈ 79`.

## Fix

Replace the integer shift with `f64::exp2()`:

```rust
let mut scaled_epsilon = epsilon * (nmax as f64).exp2();
```

This is mathematically identical (`2^nmax`) but uses floating-point arithmetic. `f64` handles `nmax` up to ~1023 before saturating to `+inf` — well beyond any epsilon that makes physical sense.

## Test

Added `test_solve_itp_large_nmax_does_not_overflow` with `epsilon = 1e-24`, which forces `nmax ≈ 79`. This panics on the original code in debug mode and returns the wrong root in release mode.

### DIFF
--- a/kurbo/src/common.rs
+++ b/kurbo/src/common.rs
@@ -710,7 +710,8 @@ pub(crate) fn solve_itp_fallible<E>(
 ) -> Result<(f64, f64), E> {
     let n1_2 = (((b - a) / epsilon).log2().ceil() - 1.0).max(0.0) as usize;
     let nmax = n0 + n1_2;
-    let mut scaled_epsilon = epsilon * (1u64 << nmax) as f64;
+    let powi_nmax: i32 = nmax.min(i32::MAX as usize).try_into().unwrap_or(i32::MAX);
+    let mut scaled_epsilon = epsilon * 2.0_f64.powi(powi_nmax);
     while b - a > 2.0 * epsilon {
         let x1_2 = 0.5 * (a + b);
         let r = scaled_epsilon - 0.5 * (b - a);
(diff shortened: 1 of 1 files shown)


greysquirr3l:
Ope, investigating CI issue.

mlwilkerson:
Oh, I think this may be the fix for issue #602 I submitted recently.

raphlinus:
I'm going to dig deeply into this set of problems soon. My immediate question is whether nmax > 63 is ever valid. A back-of-the-envelope analysis suggests that with bisection you're never going to get more than the number of mantissa bits (52 or 53 bits), so I'm thinking of just clamping nmax to that. That may well mean not being able to satisfy the precision request implied by epsilon.

Clearly the existing code is wrong and needs a fix. I just need to find out where these too-small epsilons are coming from and what the best behavior for the solver should be in those cases.

greysquirr3l:
Answered your question about where the too-small epsilons come from (for the fold-repro case) over on #602: https://github.com/linebender/kurbo/issues/602#issuecomment-5315417478

PR #589, opened 2026-06-05
Add affine transform classification

waywardmonkeys:
Add `TransformClass` for describing how restrictive an `Affine` delta is, with public `Affine` helpers for classifying a transform relative to identity or another transform.

Classify matching linear parts directly so translation-only differences avoid inversion, including singular transforms, while still using relative deltas to recognize rotation, reflection, and uniform-scale changes.


waywardmonkeys:
An LLM (Codex 5.5, xhigh) was used in the production of these changes.

PR #590, opened 2026-06-09
Add Transform3d for homogeneous layer transforms

waywardmonkeys:
Add `Transform3d` as a column-major 4 by 4 homogeneous transform in `kurbo`, with CSS/`DOMMatrix` coefficient order and multiplication semantics matching `Affine`.

The initial API covers construction from coefficients and columns, `Affine` and `TranslateScale` conversion, exact 2D and 3D-affine classification, CSS-style perspective, homogeneous point transformation, and projection with `w > 0` handling.

This intentionally leaves full projective `inverse` out of the first slice, avoiding an affine-only inverse on a type that can represent perspective.

### DIFF
--- a/kurbo/src/transform3d.rs
+++ b/kurbo/src/transform3d.rs
@@ -1,0 +1,716 @@
+// Copyright 2026 the Kurbo Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! 3D homogeneous transforms.
+
+use core::ops::{Mul, MulAssign};
+
+use crate::{Affine, Point, TranslateScale};
+
+#[cfg(not(feature = "std"))]
+use crate::common::FloatFuncs;
+
+/// A 3D homogeneous transform stored as a column-major 4 by 4 matrix.
+///
+/// The coefficients use the same order as CSS `matrix3d()` and
+/// [`DOMMatrix`](https://drafts.csswg.org/geometry/#dommatrix):
+///
+/// ```text
+/// | x' |   | m11 m21 m31 m41 |   | x |
+/// | y' | = | m12 m22 m32 m42 | * | y |
+/// | z' |   | m13 m23 m33 m43 |   | z |
+/// | w' |   | m14 m24 m34 m44 |   | w |
+/// ```
+///
+/// The translation terms are `m41`, `m42`, and `m43`. Multiplication follows
+/// the same convention as [`Affine`]: `(A * B) * p == A * (B * p)`, so `B` is
+/// applied first and `A` is applied second.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Transform3d([[f64; 4]; 4]);
+
+impl Transform3d {
+    /// The identity transform.
+    pub const IDENTITY: Self = Self::scale(1.0);
+
+    /// Construct a transform from 16 coefficients in column-major order.
+    ///
+    /// This is the same order used by CSS `matrix3d()` and `DOMMatrix` arrays:
+    /// `[m11, m12, m13, m14, m21, ... m44]`.
+    #[inline]
+    #[must_use]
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
This PR was done with assistance from Codex (5.5, xhigh).

kurbo/src/transform3d.rs
```diff
@@ -0,0 +1,716 @@
...
+            [coeffs[0], coeffs[1], coeffs[2], coeffs[3]],
+            [coeffs[4], coeffs[5], coeffs[6], coeffs[7]],
+            [coeffs[8], coeffs[9], coeffs[10], coeffs[11]],
+            [coeffs[12], coeffs[13], coeffs[14], coeffs[15]],
+        ])
+    }
+
+    /// Construct a transform from its four columns.
+    ///
+    /// Each column is `[x, y, z, w]`.
```
DJMcNab: Could you expand on this comment? The column is the contribution of their respective element to the final result, right? So when you say `x` in the doc comment, it's the contribution of the element corresponding to the column number to the resulting x?

kurbo/src/transform3d.rs
```diff
@@ -0,0 +1,716 @@
...
+    /// # Panics
+    ///
+    /// Panics if `i >= 4`.
+    #[inline]
+    #[must_use]
+    pub const fn col(self, i: usize) -> [f64; 4] {
+        self.0[i]
+    }
+
+    /// Construct the 3D transform with the same effect as a 2D [`Affine`].
```
DJMcNab: I'd probably add the "in the xy plane" here, but I think we could get away without it.

kurbo/src/transform3d.rs
```diff
@@ -0,0 +1,716 @@
...
+    }
+
+    /// Return column `i`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `i >= 4`.
+    #[inline]
+    #[must_use]
+    pub const fn col(self, i: usize) -> [f64; 4] {
```
DJMcNab, line 108: Is there a use for a `row` function?

kurbo/src/transform3d.rs
```diff
@@ -0,0 +1,716 @@
...
+            && c[1][2] == 0.0
+            && c[1][3] == 0.0
+            && c[2][0] == 0.0
+            && c[2][1] == 0.0
+            && c[2][2] == 1.0
+            && c[2][3] == 0.0
+            && c[3][0] == 0.0
+            && c[3][1] == 0.0
+            && c[3][2] == 0.0
+            && c[3][3] == 1.0
```
DJMcNab, line 423: Could we use `self == Self::IDENTITY`?

DJMcNab (commented):
As an implementation of this feature, this is unobjectionable. I do have a few nits, but none are especially blocking. I don't feel like I'm especially qualified to decide whether we want this (this is the kind of thing I miss having a strong office hours for). I definitely don't want to make that decision at the end of a workday.

I do agree that as a project, Linebender does need a way to talk about 3d transforms in 2d rendering. I don't want this to sound like I'm trying to blackhole this feature, but part of me wonders if it would actually belong in Peniko (as it's more of a 2d graphics than 2d shapes thing). For the sake of keeping moving forward, I'm definitely not against landing this here.

@tomcur do you have any thoughts on this conceptually?

kurbo/src/transform3d.rs
waywardmonkeys: There is, for matrix inspection but also to get the bottom / perspective row. Added in next push.

kurbo/src/transform3d.rs
waywardmonkeys: Adjusted.

kurbo/src/transform3d.rs
waywardmonkeys: Adjusted in next push.

kurbo/src/transform3d.rs
waywardmonkeys: `const fn` in the workspace MSRV, I think.

DJMcNab: It might be possible to do `self.0 == IDENTITY.0`, maybe?

PR #591, opened 2026-06-28
`serde` and `schemars` support for `Axis`

waywardmonkeys:


DJMcNab (approved):
I'm intrigued that this is being used; I was under the impression that these were on the way out.

Out of interest, are you actually using the schemars support, or just serde?

waywardmonkeys:
I don't use the `schemars` support but since everything else with `serde` also had it, it seemed wrong to omit it.

I think having `serde` on various types is okay. The `schemers` is rockier (and is what we've mainly discussed removing, I think).

I ran into this as I had to have my own `Axis` type in some code rather than use this because it was serializing a tiled panel configuration for save/restore.

PR #592, opened 2026-06-28
Axis-aware coordinate helpers on `Point`, `Vec2`, `Size`, and `Rect`

waywardmonkeys:


kurbo/src/point.rs
```diff
@@ -229,6 +229,25 @@ impl Point {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use kurbo::{Axis, Point};
+    ///
+    /// let point = Point::new(2.0, 3.0).with_coord(Axis::Vertical, 5.0);
+    /// assert_eq!(point, Point::new(2.0, 5.0));
+    /// ```
+    #[inline]
+    pub const fn with_coord(self, axis: Axis, value: f64) -> Point {
```
DJMcNab, line 245: What do you think about a name like `snap_to_coord`. That name also doesn't quite have the right ring to it, but I find that `with_coord` is a name which really leans on the docs to explain it.

kurbo/src/rect.rs
```diff
@@ -641,6 +641,93 @@ impl Rect {
...
+    ///
+    /// ```
+    /// use kurbo::{Axis, Rect};
+    ///
+    /// let rect = Rect::new(8.0, 2.0, 4.0, 7.0);
+    /// assert_eq!(rect.min_coord(Axis::Horizontal), 4.0);
+    /// assert_eq!(rect.min_coord(Axis::Vertical), 2.0);
+    /// ```
+    #[inline]
+    pub const fn min_coord(&self, axis: Axis) -> f64 {
```
DJMcNab, line 683: If I read this name in a vacuum, I'd expect it to be `(self.min_x(), self.min_y())`. Especially since that method doesn't exist, the idea of adding it with these semantics at least gives me pause.

kurbo/src/rect.rs
```diff
@@ -667,6 +754,32 @@ impl Rect {
...
+    ///
+    /// let rect = Rect::new(1.0, 2.0, 4.0, 7.0).with_coords(
+    ///     Axis::Horizontal,
+    ///     10.0,
+    ///     3.0,
+    /// );
+    /// assert_eq!(rect, Rect::new(10.0, 2.0, 3.0, 7.0));
+    /// ```
+    #[inline]
+    pub const fn with_coords(self, axis: Axis, v0: f64, v1: f64) -> Rect {
```
DJMcNab, line 777: Does a name like `with_extents` make sense to you? That feels a little bit too jargony. Wow, what a fine line naming things has...

DJMcNab (approved):
I do want to make sure that we've thought about these names, because none of them particularly spark joy. Even if the response to each question is just "I can't think of anything better", that's reasonable.

This really just continues the existing naming scheme from the previously added methods of this flavour. Looking back on them, those names are not something I'm *hugely* happy with. At least this is consistent

kurbo/src/rect.rs
waywardmonkeys: Yeah. This is not the first version of this PR that I've written. Perhaps the third, all with different names!

kurbo/src/rect.rs
waywardmonkeys: Since this is an axis-specific method, one option might be `min_along` / `max_along` ...

waywardmonkeys:
> I do want to make sure that we've thought about these names, because none of them particularly spark joy. Even if the response to each question is just "I can't think of anything better", that's reasonable.
> 
> This really just continues the existing naming scheme from the previously added methods of this flavour. Looking back on them, those names are not something I'm _hugely_ happy with. At least this is consistent

I agree on both counts. The names don't really spark joy, but neither have any others that I've come up with.

I'm happy for this to take a short period of time to see if we can find better names.

kurbo/src/rect.rs
DJMcNab: I like that!

nicoburns:
Also agree that these are helpful additions but the names don't spark joy. In Taffy I have used `start` and `end` for `min` and `max`, but I guess that might be confusing in kurbo :/

PR #597, opened 2026-07-23
Fix zero-length `stroke` and `dash` handling

### LINKED ISSUE #577: Dash iterator zero-length LineTo noise

waywardmonkeys:
Preserve intentional zero-length geometry through `dash()` so `stroke()` can apply its cap style instead of dropping it as a degenerate segment.

Emit cap-only geometry for open and closed zero-length subpaths, including degenerate line, quadratic, and cubic commands. `Butt` caps remain empty, while `Round`, `Square`, and mixed caps produce the corresponding full or half shape.

Keep dash starts in the half-open range `[0, path_length)`, avoiding spurious output at exact segment and open-path boundaries. Preserve explicit zero-length dashes as `MoveTo(p)`/`LineTo(p)` pairs, and treat zero-sum or non-finite dash inputs as undashed without relying on `NaN` normalization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ This release has an [MSRV][] of 1.85.
 ## Added
 - `serde` and `schemars` support for `Axis`. ([#591][] by [@waywardmonkeys][])
 
+## Fixed
+
+- `stroke` now emits cap geometry for zero-length subpaths when round or square caps are selected. ([#597][] by [@waywardmonkeys][])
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
I didn't mean to request a Copilot review. :(

waywardmonkeys:
Ends up Copilot was right about another edge case. Fixed it in a different way.

